PR #32, opened 2022-08-07
De-scope the Data Integrity specification to focus on Verifiable Credentials.

msporny:
This PR de-scopes the Data Integrity specification to primarily focus on Verifiable Credentials.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/32.html" title="Last updated on Aug 15, 2022, 3:17 PM UTC (a8110f0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/32/3d32fae...a8110f0.html" title="Last updated on Aug 15, 2022, 3:17 PM UTC (a8110f0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
 <head>
-    <title>Data Integrity 1.0</title>
+    <title>Verifiable Credential Data Integrity 1.0</title>
     <meta http-equiv='Content-Type' content='text/html;charset=utf-8'/>
     <!--
       === NOTA BENE ===
@@ -17,7 +17,10 @@
           specStatus:           "ED",
 
           // the specification's short name, as in http://www.w3.org/TR/short-name/
-          shortName:            "data-integrity",
+          shortName:            "vc-data-integrity",
+
+          // subtitle
+          subtitle: "Securing the Integrity of Verifiable Credential Data",
 
           // if you wish the publication date to be other than today, set this
           // publishDate:  "2009-08-06",
@@ -28,7 +31,7 @@
           // previousMaturity:  "WD",
 
           // if there a publicly available Editor's Draft, this is the link
-          edDraftURI:           "https://w3c.github.io/data-integrity/",
+          edDraftURI:           "https://w3c.github.io/vc-data-integrity/",
 
           // if this is a LCWD, uncomment and set the end of its review period
           // lcEnd: "2009-08-05",
@@ -71,7 +74,7 @@
           // name (with the @w3c.org) of the public mailing to which comments are due
           wgPublicList: "public-credentials",
 
-          github: "w3c/data-integrity",
+          github: "w3c/vc-data-integrity",
 
           // URI of the patent status for this WG, for Rec-track documents
           // !!!! IMPORTANT !!!!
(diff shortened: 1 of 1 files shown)


dlongley (approved):

mprorock (approved):

peacekeeper:
To me it seems a bit strange to give something that's generically applicable (not just to VCs) such a restrictive name. This feels like renaming generic JWT to VC-JWT, even though JWT can be used for other non-VC applications.

But I won't object if others support the change.

msporny:
> To me it seems a bit strange to give something that's generically applicable (not just to VCs) such a restrictive name. This feels like renaming generic JWT to VC-JWT, even though JWT can be used for other non-VC applications.

I agree, I'd rather keep the specification name aligned with what it's capable of doing. The danger here is that people end up thinking that Data Integrity is only useful when used with Verifiable Credentials, when it is designed to be a generalized data integrity mechanism. While I'm not in favor of this change, I won't object to it as long as actions are not taken to establish DI as a VC-only thing. If we see that occurring, then we will want to undo this PR and re-align long-term expectations.

msporny:
> > To me it seems a bit strange to give something that's generically applicable (not just to VCs) such a restrictive name. This feels like renaming generic JWT to VC-JWT, even though JWT can be used for other non-VC applications.

I agree, I'd rather keep the specification name aligned with what it's capable of doing. The danger here is that people end up thinking that Data Integrity is only useful when used with Verifiable Credentials, when it is designed to be a generalized data integrity mechanism. While I'm not in favor of this change, I won't object to it as long as actions are not taken to establish DI as a VC-only thing. If we see that occurring, then we will want to undo this PR and re-align long-term expectations.

mprorock:
I think If we define it well for VCs we can always look at generalizing later.  If we try to generalize early (even though we can obviously apply some of the techniques here broadly) we risk not getting the core matter of Data Integrity as related to VCs tightly and well defined.  We need extremely well defined and thoroughly picked apart and inspected data integrity for VCs and we should focus first and foremost on ensuring that is in place.  Then if we want to look at other cases of how to use this with broader signing of JSON-LD and similar data cool, but let's get this well proven and testable in the VC only use case first.

index.html
```diff
@@ -138,8 +141,9 @@
     <section id='abstract'>
       <p>
 This specification describes mechanisms for ensuring the authenticity and
-integrity of structured digital documents using cryptography, such as digital
-signatures and other digital mathematical proofs.
+integrity of Verifiable Credentials and other types of structured digital
+documents using cryptography, such as digital signatures and other digital
+mathematical proofs.
```
```suggestion
integrity of Verifiable Credentials using cryptography and especially through the use of digital signatures and related mathematical proofs.
```

mprorock (approved):
minor suggestion.

index.html
msporny: -1, the removal of "and other types of structured digital documents using cryptography" is exactly the sort of goal post shifting that is of concern with this PR. The following modification to your proposal could work:

```suggestion
integrity of Verifiable Credentials and similar types of constrained digital
documents using cryptography, especially through the use of digital 
signatures and related mathematical proofs.
```

mprorock: Good with this

OR13 (approved):

vongohren (approved):
Based on discussions it looks like a good plan

msporny:
Substantive, multiple reviews, no strong objections (but cautious approval), merging.

PR #36, opened 2022-08-14
Add Security Considerations section on Versioning Cryptography Suites.

msporny:
This PR adds a security considerations section on versioning cryptographic suites and outlines some design rationale behind the date-identified suites that we currently have.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/36.html" title="Last updated on Aug 19, 2022, 7:46 PM UTC (b1319da)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/36/3b27d11...b1319da.html" title="Last updated on Aug 19, 2022, 7:46 PM UTC (b1319da)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -127,6 +127,15 @@
               ],
               status: "Internet-Draft",
               publisher: "IETF"
+            },
+            "VC-EXTENSION-REGISTRY": {
+              title: "Verifiable Credentials Extension Registry",
+              href: "https://w3c-ccg.github.io/vc-extension-registry/",
+              authors: [
+                "Manu Sporny"
+              ],
+              status: "CG-DRAFT",
+              publisher: "Credentials Community Group"
             }
           },
           postProcess: [restrictRefs]
@@ -1767,6 +1776,72 @@ software.
 
 <div class="issue">TODO: We need to add a complete list of security
 considerations.</div>
+
+      <section>
+        <h3>Versioning Cryptography Suites</h3>
+
+        <p>
+Cryptography protects information in a way that requires knowledge of a secret
+to access the information (computationally easy), or a concerted
+brute-force attack to access the same information without the secret
+(computationally difficult). All modern cryptography requires the
+computationally difficult approach to remain difficult throughout time, which
+does not always hold due to breakthroughs in science and mathematics. That is
+to say that <em>Cryptography has a shelf life</em>.
+        </p>
+        <p>
+This specification plans for the obsolecense of all cryptographic approaches by
+asserting that whatever cryptography is in use today is highly likely to be
+broken over time. Given this assumption, it makes sense to identify
+cryptographic suites in a way where we can differentiate different versions
+throughout time. When versioning cryptographic suites, there are several
+approaches that can be taken which include: parameters, numbers, and dates.
+        </p>
+        <p>
+Parametric versioning specifies the particular cryptographic parameters that are
+employed in a cryptographic suite. For example, one could use an identifier such
+as `RSAASSA-PKCS1-v1_5-SHA1`. The benefit to this scheme is that a well-trained
+cryptographer will be able to determine all of the parameters in play by the
+identifier. The drawback to this scheme is that most of the population that uses
+these sorts of identifiers are not well trained and thus will not understand
+that the previously mentioned identifier is a cryptographic suite that is no
+longer safe to use.
+        </p>
+        <p>
+Numbered versioning might specify a major and minor version number such as
+`1.0` or `2.1`. Numbered versioning conveys a specific order and suggests that
+higher version numbers are more capable than lower version numbers. The benefit
+of this approach is that it removes complex parameters that less expert
+developers might not understand with a simpler model that conveys that an
+upgrade might be appropriate. The drawback of this approach is that its not
+clear if an upgrade is necessary, as software version number increases often
+don't require an upgrade for the software to continue functioning. This can
+lead to developers thinking their usage of a particular version is safe, when
+it is not.
+        </p>
+        <p>
+Date-based versioning specifies a particular release date for a specific
+cryptographic suite. The benefit of a date, such as a year, is that it is
+immediately clear to a developer if the date is relatively old or new. Seeing
+an old date might prompt the developer to go searching for a newer
+cryptographic suite, where as a parametric or number-based versioning scheme
+might not. The downside of a date-based version is that some cryptographic
+suites might not expire for 5-10 years, prompting the developer to go
+searching for a newer cryptographic suite only to not find one that is newer.
+While this might be an inconvenience, it is one that results in safer
+ecosystem behaviour.
+        </p>
+        <p>
+It is highly encouraged that cryptographic suite identifiers are versioned
+using a year designation. For example, the cryptographic suite identifier
+`ecdsa-2022` implies that the suite is probably ok to use in the year 2025,
+but might not be a safe choice in the year 2042. A date-based versioning
+mechanism, however, is not enough by itself. All cryptographic suites that
+follow this specification are intended to be registered
+[[?VC-EXTENSION-REGISTRY]] in a way that clearly signal which cryptosuites
+are deprecated, standardized, or experimental.
+        </p>
+      </section>
 
       <section>
         <h3>Verification Method Binding</h3>
--- a/index.html
+++ b/index.html
@@ -1872,7 +1872,10 @@ COSE Algorithms Registry</a>, the
 JOSE Algorithms Registry</a>, and the latest
 <a href="https://csrc.nist.gov/publications/detail/fips/186/4/final">
 FIPS Digital Signature Standard
-</a> guidance.  Depending on your jurisdiction you may also need to consider additional guidance from <a href="https://www.etsi.org/deliver/etsi_ts/119300_119399/119312/01.02.01_60/ts_119312v010201p.pdf">ETSI</a> or other regulatory and standards bodies.
+</a> guidance.  Depending on your jurisdiction you may also need to consider
+additional guidance from
+<a href="https://www.etsi.org/deliver/etsi_ts/119300_119399/119312/01.02.01_60/ts_119312v010201p.pdf">ETSI</a>
+or other regional regulatory and standards bodies.
         </p>
       </section>
 
--- a/index.html
+++ b/index.html
@@ -1859,6 +1859,21 @@ Use of experimental suites are expected to throw errors in implementations
 unless a `useExperimentalCryptosuites` option is used used specifying exactly
 the experimental cryptosuite to use.
         </p>
+
+        <p class="issue" title="Resolve VC Extension Registry Governance">
+The following text is debated due to the governance rules of the VC
+Extension Registry not being finalized:<br><br>
+Developers are urged to always refer to the [[?VC-EXTENSION-REGISTRY]] for
+the latest standardized cryptography suites as well as cross-checking their
+usage against, at least, the
+<a href="https://www.iana.org/assignments/cose/cose.xhtml#algorithms">
+COSE Algorithms Registry</a>, the
+<a href="https://www.iana.org/assignments/jose/jose.xhtml#web-signature-encryption-algorithms">
+JOSE Algorithms Registry</a>, and the latest
+<a href="https://csrc.nist.gov/publications/detail/fips/186/4/final">
+FIPS Digital Signature Standard
+</a> guidance.
+        </p>
       </section>
 
       <section>
--- a/index.html
+++ b/index.html
@@ -1842,15 +1842,22 @@ searching for a newer cryptographic suite only to not find one that is newer.
 While this might be an inconvenience, it is one that results in safer
 ecosystem behavior.
         </p>
-        <p>
+        <p class="issue" data-number="38">
+The following text is currently under debate:<br><br>
 It is highly encouraged that cryptographic suite identifiers are versioned
 using a year designation. For example, the cryptographic suite identifier
-`ecdsa-2022` implies that the suite is probably ok to use in the year 2025,
-but might not be a safe choice in the year 2042. A date-based versioning
-mechanism, however, is not enough by itself. All cryptographic suites that
-follow this specification are intended to be registered
+`ecdsa-2022` implies that the suite is probably an acceptable of ECDSA in the
+year 2025, but might not be a safe choice in the year 2042. A date-based
+versioning mechanism, however, is not enough by itself. All cryptographic
+suites that follow this specification are intended to be registered
 [[?VC-EXTENSION-REGISTRY]] in a way that clearly signal which cryptosuites
-are deprecated, standardized, or experimental.
+are deprecated, standardized, or experimental. Cryptosuite registration will
+follow CFRG, IETF, NIST, FIPS, and safecurves guidance. Use of deprecated suites
+are expected to throw errors in implementations unless a `useUnsafeCryptosuites`
+option is used specifying exactly the unsafe cryptosuite to use.
+Use of experimental suites are expected to throw errors in implementations
+unless a `useExperimentalCryptosuites` option is used used specifying exactly
+the experimental cryptosuite to use.
         </p>
       </section>
 
--- a/index.html
+++ b/index.html
@@ -1807,10 +1807,14 @@ Parametric versioning specifies the particular cryptographic parameters that are
 employed in a cryptographic suite. For example, one could use an identifier such
 as `RSASSA-PKCS1-v1_5-SHA1`. The benefit to this scheme is that a well-trained
 cryptographer will be able to determine all of the parameters in play by the
-identifier. The drawback to this scheme is that most of the population that uses
-these sorts of identifiers are not well trained and thus will not understand
+identifier. The drawback to this scheme is that most of the population that
+uses these sorts of identifiers are not well trained and thus will not understand
 that the previously mentioned identifier is a cryptographic suite that is no
-longer safe to use.
+longer safe to use. Additionally, this lack of knowledge might lead software
+developers to generalize the parsing of cryptographic suite identifiers
+such that any combination of cryptographic primitives becomes acceptable,
+resulting in reduced security. Ideally, cryptographic suites are implemented
+in software as specific, acceptable profiles of cryptographic parameters instead.
         </p>
         <p>
 Numbered versioning might specify a major and minor version number such as
--- a/index.html
+++ b/index.html
@@ -1856,7 +1856,7 @@ follow CFRG, IETF, NIST, FIPS, and safecurves guidance. Use of deprecated suites
 are expected to throw errors in implementations unless a `useUnsafeCryptosuites`
 option is used specifying exactly the unsafe cryptosuite to use.
 Use of experimental suites are expected to throw errors in implementations
-unless a `useExperimentalCryptosuites` option is used used specifying exactly
+unless a `useExperimentalCryptosuites` option is used specifying exactly
 the experimental cryptosuite to use.
         </p>
 
@@ -1872,7 +1872,7 @@ COSE Algorithms Registry</a>, the
 JOSE Algorithms Registry</a>, and the latest
 <a href="https://csrc.nist.gov/publications/detail/fips/186/4/final">
 FIPS Digital Signature Standard
-</a> guidance.
+</a> guidance.  Depending on your jurisdiction you may also need to consider additional guidance from <a href="https://www.etsi.org/deliver/etsi_ts/119300_119399/119312/01.02.01_60/ts_119312v010201p.pdf">ETSI</a> or other regulatory and standards bodies.
         </p>
       </section>
 
--- a/index.html
+++ b/index.html
@@ -1781,26 +1781,31 @@ considerations.</div>
         <h3>Versioning Cryptography Suites</h3>
 
         <p>
-Cryptography protects information in a way that requires knowledge of a secret
-to access the information (computationally easy), or a concerted
-brute-force attack to access the same information without the secret
-(computationally difficult). All modern cryptography requires the
+Cryptography secures information through the use of secrets. Knowledge of the
+necessary secret makes it computationally easy to access certain information. The
+same information can be accessed if a computationally-difficult, brute-force effort
+successfully guesses the secret. All modern cryptography requires the
 computationally difficult approach to remain difficult throughout time, which
 does not always hold due to breakthroughs in science and mathematics. That is
 to say that <em>Cryptography has a shelf life</em>.
         </p>
         <p>
-This specification plans for the obsolecense of all cryptographic approaches by
+This specification plans for the obsolescence of all cryptographic approaches by
 asserting that whatever cryptography is in use today is highly likely to be
-broken over time. Given this assumption, it makes sense to identify
-cryptographic suites in a way where we can differentiate different versions
-throughout time. When versioning cryptographic suites, there are several
-approaches that can be taken which include: parameters, numbers, and dates.
+broken over time. Software systems have to be able to change the cryptography
+in use over time in order to continue to secure information. Such changes might
+involve increasing required secret sizes or modifications to the cryptographic
+primitives used. However, some combinations of cryptographic parameters
+might actually reduce security. Given these assumptions, systems need to be able to
+distinguish different combinations of safe cryptographic parameters, also known
+as cryptographic suites, from one another. When identifying or versioning
+cryptographic suites, there are several approaches that can be taken which
+include: parameters, numbers, and dates.
         </p>
         <p>
 Parametric versioning specifies the particular cryptographic parameters that are
 employed in a cryptographic suite. For example, one could use an identifier such
-as `RSAASSA-PKCS1-v1_5-SHA1`. The benefit to this scheme is that a well-trained
+as `RSASSA-PKCS1-v1_5-SHA1`. The benefit to this scheme is that a well-trained
 cryptographer will be able to determine all of the parameters in play by the
 identifier. The drawback to this scheme is that most of the population that uses
 these sorts of identifiers are not well trained and thus will not understand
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1768,6 +1777,72 @@ <h2>Security Considerations</h2>
...
+        </p>
+        <p>
+Parametric versioning specifies the particular cryptographic parameters that are
+employed in a cryptographic suite. For example, one could use an identifier such
+as `RSAASSA-PKCS1-v1_5-SHA1`. The benefit to this scheme is that a well-trained
+cryptographer will be able to determine all of the parameters in play by the
+identifier. The drawback to this scheme is that most of the population that uses
+these sorts of identifiers are not well trained and thus will not understand
+that the previously mentioned identifier is a cryptographic suite that is no
+longer safe to use.
```
mprorock: We can also just recommend that developers check current NIST / FIPS levels, safecurves, and IANA registries.

index.html
```diff
@@ -1768,6 +1777,72 @@ <h2>Security Considerations</h2>
...
+Numbered versioning might specify a major and minor version number such as
+`1.0` or `2.1`. Numbered versioning conveys a specific order and suggests that
+higher version numbers are more capable than lower version numbers. The benefit
+of this approach is that it removes complex parameters that less expert
+developers might not understand with a simpler model that conveys that an
+upgrade might be appropriate. The drawback of this approach is that its not
+clear if an upgrade is necessary, as software version number increases often
+don't require an upgrade for the software to continue functioning. This can
+lead to developers thinking their usage of a particular version is safe, when
+it is not.
```
mprorock: Since we are talking about cryptography, it is likely ok that we could establish and document a policy for versioning - e.g. breaking changes with major versions, minor versions when a security patch is pushed, and / or point to [semvar](https://semver.org/) or similar.  with cryptography we don't want to encourage experimentation, unless it is clear that there is something new and experimental - e.g. 1.0.1-alpha and 1.0.1-pre-release

index.html
```diff
@@ -1768,6 +1777,72 @@ <h2>Security Considerations</h2>
...
+        </p>
+        <p>
+It is highly encouraged that cryptographic suite identifiers are versioned
+using a year designation. For example, the cryptographic suite identifier
+`ecdsa-2022` implies that the suite is probably ok to use in the year 2025,
+but might not be a safe choice in the year 2042. A date-based versioning
+mechanism, however, is not enough by itself. All cryptographic suites that
+follow this specification are intended to be registered
+[[?VC-EXTENSION-REGISTRY]] in a way that clearly signal which cryptosuites
+are deprecated, standardized, or experimental.
```
mprorock: this whole approach seems to be implying that we want people to use "newer" as though that is better, and i am highly skeptical of this. I also do not want to be using cryptography or signatures selected by non-experts, or encourage non experts to make cryptographic selections.

mprorock (changes requested):
I would strongly prefer versioning / naming that identifies the algorithm in use and the core parameter set.  second two that, I would consider versioning using semvar or similar.  

I also would like to see pointers to NIST and eIDAS to encourage the developer to review based on their operating environment

index.html
mprorock: A further note. Seems like we really should be wanting to encourage semvar or similar at the library and implementation layer.

OR13 (changes requested):
> [[?VC-EXTENSION-REGISTRY]] in a way that clearly signal which cryptosuites
are deprecated, standardized, or experimental.

We should discuss the registry and its operational authorities and requirements, before adding language like this... who controls these status changes?

How do I deprecate a suite?

What counts as "standardized"?

Without more clarity on how the W3C will run the vc extensions registry I am concerned we are opening a large can of worms.

index.html
msporny: Yes, I'll add a suggestion to check current NIST/FIPS levels, safecurves, and IANA registries. 

We should also be realistic about doing so -- many developers don't do that and use whatever is available in their libraries after a quick search on stack overflow (which might be an answer from years ago, instead of a current one). All that to say, the standard advice given to date is to do the things above and that has been demonstrated to not be as effective as we need it to be -- a simple read through of monthly CVEs and Equifax-like breaches demonstrate this to be true. We have decades of data on how depending on developers to "read the docs" is not as effective as we need it to be.

index.html
msporny: > it is likely ok that we could establish and document a policy for versioning

You seem to be arguing for two different things -- parametric versioning and semver. We will need to come to ground on ONE way forward for Data Integrity, given that having two ways will just confuse things more.

Semver is more workable than parametric versioning, given that parametric versioning presumes that you get the implementation right the first time around. The proposed pattern is: `CRYPTOSUITE-VERSION`, so `eddsa-2022` or `eddsa-v1`... the parametric version of that would be `urdna2015-eddsa-ed25519` (but based on emails over the past several weeks, we already have a weak signal in CFRG that `eddsa-ed25519` is going to have two variations already -- deterministic and random -- coupled with two canonicalization schemes -- which means that if we had gone with parametric to begin with, we'd have a backwards compatability problem on our hands already -- and that's not even counting all the parametric churn that's happened to RSA and ECDSA over the years).

At this point, I'd like to understand how serious you are about pushing for parametric versioning... because that's probably the worst option that's being contemplated (for at least the reasons outlined above -- there are others that we can get into if someone is going to argue for it).

> Seems like we really should be wanting to encourage semvar or similar at the library and implementation layer.

Yes, we should have implementation guidance in the Data Integrity spec that promotes whatever mechanism we end up settling on. There should be a simple way to warn the developer: "Hey, you're using an outdated mechanism -- turn on `unsafe` mode if you want to keep using it".

index.html
msporny: > I also do not want to be using cryptography or signatures selected by non-experts, or encourage non experts to make cryptographic selections.

This happens today when people pick up and use kitchen sink cryptography libraries. Options are selected where the ramifications of those selections are not known to non-expert developers.

> this whole approach seems to be implying that we want people to use "newer" as though that is better, and i am highly skeptical of this. 

You might be thinking of the date in isolation. It is ONE tool in the tool chest, the other tools being vetted standards registries, sites like safecurves, and naming classes of cryptography suite. The suggestion isn't to name the cryptosuite `2022` and be done with it. The suggestion is to name the suite something like `eddsa-2022` and `eddsa-2029`, where the latter has achieved some form of consensus as being "better" than the former.

msporny:
> > [[?VC-EXTENSION-REGISTRY]] in a way that clearly signal which cryptosuites
> > are deprecated, standardized, or experimental.
> 
> We should discuss the registry and its operational authorities and requirements, before adding language like this... 

I'll note that this PR just refers to a registry that was created by the VCWG and already exists (and is in terrible shape):

https://w3c-ccg.github.io/vc-extension-registry/

> who controls these status changes?
> How do I deprecate a suite?
> What counts as "standardized"?
> Without more clarity on how the W3C will run the vc extensions registry I am concerned we are opening a large can of worms.

I agree that all of these are questions that the VCWG needs to answer. Would an issue marker noting that these questions need to be resolved track your concerns in a way that would unblock this PR, or do you require additional changes?

index.html
mprorock: my personal preference is towards parametric for naming - with libraries that implement encouraged towards semvar for versioning 

parametric naming is separate from versioning in my mind - versioning would be handled by the current WG level (likely with semvar or similar - e.g. `urdna2015-eddsa-ed25519-r` standardized along side the VC 2.0 TR

msporny: > my personal preference is towards parametric for naming

If you prefer parametric naming, how do you plan to address data format changes, like when we switched from `jws` to `proofValue` or if the encoded value of `proofValue` changes for a particular cryptographic suite? One of the drawbacks for parametric naming is when things outside of the parameters change that affect the cryptography (such as the data format for a particular parameter, or the omission of a hash function). If you prefer parametric naming, then do you oppose non-parametric names like "RS1" and "P256" and "RSA1_5"?

index.html
OR13: > The suggestion is to name the suite something like eddsa-2022 and eddsa-2029, where the latter has achieved some form of consensus as being "better" than the former.

I would like to review the registry policy for handling this... If we can't agree to a registry process and confidence in the  authority that maintains it, there is no reason to believe `eddsa-2029` will be anything other than a very lucrative form of planned obsolescence.

Is the W3C the right place to maintain such a registry?

Will browsers implement all the registered suites, or only a single mandatory to implement one?

Why stop at `eddsa` why not `non-repudiable-signature-2029`?

OR13:
> Would an issue marker noting that these questions need to be resolved track your concerns in a way that would unblock this PR, or do you require additional changes?

An issue marker would work.

msporny:
> > Would an issue marker noting that these questions need to be resolved track your concerns in a way that would unblock this PR, or do you require additional changes?
> 
> An issue marker would work.

I've raised an issue here: https://github.com/w3c/vc-data-model/issues/909

That issue will be referenced in an issue marker in the current PR.

index.html
```diff
@@ -1768,6 +1777,72 @@ <h2>Security Considerations</h2>
...
 considerations.</div>
 
+      <section>
+        <h3>Versioning Cryptography Suites</h3>
+
+        <p>
+Cryptography protects information in a way that requires knowledge of a secret
+to access the information (computationally easy), or a concerted
+brute-force attack to access the same information without the secret
+(computationally difficult). All modern cryptography requires the
```
```suggestion
Cryptography secures information through the use of secrets. Knowledge of the
necessary secret makes it computationally easy to access certain information. The
same information can be accessed if a computationally-difficult, brute-force effort
successfully guesses the secret. All modern cryptography requires the
```

index.html
```diff
@@ -1768,6 +1777,72 @@ <h2>Security Considerations</h2>
...
+does not always hold due to breakthroughs in science and mathematics. That is
+to say that <em>Cryptography has a shelf life</em>.
+        </p>
+        <p>
+This specification plans for the obsolecense of all cryptographic approaches by
+asserting that whatever cryptography is in use today is highly likely to be
+broken over time. Given this assumption, it makes sense to identify
+cryptographic suites in a way where we can differentiate different versions
+throughout time. When versioning cryptographic suites, there are several
+approaches that can be taken which include: parameters, numbers, and dates.
```
```suggestion
This specification plans for the obsolescence of all cryptographic approaches by
asserting that whatever cryptography is in use today is highly likely to be
broken over time. Software systems have to be able to change the cryptography
in use over time in order to continue to secure information. Such changes might
involve increasing required secret sizes or modifications to the cryptographic
primitives used. However, some combinations of cryptographic parameters
might actually reduce security. Given these assumptions, systems need to be able to
distinguish different combinations of safe cryptographic parameters, also known
as cryptographic suites, from one another. When identifying or versioning
cryptographic suites, there are several approaches that can be taken which
include: parameters, numbers, and dates.
```

index.html
```suggestion
identifier. The drawback to this scheme is that most of the population that
uses these sorts of identifiers are not well trained and thus will not understand
that the previously mentioned identifier is a cryptographic suite that is no
longer safe to use. Additionally, this lack of knowledge might lead software
developers to generalize the parsing of cryptographic suite identifiers
such that any combination of cryptographic primitives becomes acceptable,
resulting in reduced security. Ideally, cryptographic suites are implemented
in software as specific, acceptable profiles of cryptographic parameters instead.
```

index.html
```suggestion
lead to developers thinking their usage of a particular version is safe, when
it is not. Ideally, additional signals would be given to developers that use
cryptographic suites in their software that periodic reviews of those
suites for continued security are required.
```

index.html
```diff
@@ -1768,6 +1777,72 @@ <h2>Security Considerations</h2>
...
+Date-based versioning specifies a particular release date for a specific
+cryptographic suite. The benefit of a date, such as a year, is that it is
+immediately clear to a developer if the date is relatively old or new. Seeing
+an old date might prompt the developer to go searching for a newer
+cryptographic suite, where as a parametric or number-based versioning scheme
+might not. The downside of a date-based version is that some cryptographic
+suites might not expire for 5-10 years, prompting the developer to go
+searching for a newer cryptographic suite only to not find one that is newer.
+While this might be an inconvenience, it is one that results in safer
+ecosystem behaviour.
```
```suggestion
ecosystem behavior.
```

dlongley (changes requested):

index.html
```diff
@@ -1768,6 +1777,72 @@ <h2>Security Considerations</h2>
...
+asserting that whatever cryptography is in use today is highly likely to be
+broken over time. Given this assumption, it makes sense to identify
+cryptographic suites in a way where we can differentiate different versions
+throughout time. When versioning cryptographic suites, there are several
+approaches that can be taken which include: parameters, numbers, and dates.
+        </p>
+        <p>
+Parametric versioning specifies the particular cryptographic parameters that are
+employed in a cryptographic suite. For example, one could use an identifier such
+as `RSAASSA-PKCS1-v1_5-SHA1`. The benefit to this scheme is that a well-trained
```
dlongley: Typo:

```suggestion
as `RSASSA-PKCS1-v1_5-SHA1`. The benefit to this scheme is that a well-trained
```

selfissued (changes requested):
We should be relying on algorithm assessment by experts and not trying to do this ourselves.  For instance, entries in the IANA JOSE Algorithms Registry https://www.iana.org/assignments/jose/jose.xhtml#web-signature-encryption-algorithms include not just a specification reference but an assessment of the status of the algorithm (Required, Recommended, Optional, Prohibited), etc. that can be relied on by us, rather than trying to reinvent this wheel.

Like @mprorock , I'm also highly skeptical that adding a year to an algorithm name adds actionable information.  For one thing, you can determine the year anyway by following the registry pointer to the defining specification and looking at the spec date, if you care.  Furthermore, you'll still have to consult the registry entry to determine whether the algorithm has been deprecated.  A date isn't going to give you that.  When something isn't useful, we should leave it out.

mprorock:
> For one thing, you can determine the year anyway by following the registry pointer to the defining specification and looking at the spec date, if you care. Furthermore, you'll still have to consult the registry entry to determine whether the algorithm has been deprecated. A date isn't going to give you that. When something isn't useful, we should leave it out.

In this case it is even worse, as it may give the impression something is fine when it is not

OR13:
@msporny I'm fine approving this once the issue is merged into the spec text.

index.html
msporny: > I would like to review the registry policy for handling this... If we can't agree to a registry process and confidence in the authority that maintains it, there is no reason to believe `eddsa-2029` will be anything other than a very lucrative form of planned obsolescence.

This is being tracked in https://github.com/w3c/vc-data-model/issues/909

OR13:
@selfissued @mprorock your suggestion to only use the other registries won't work for all the data integrity cases...

Since the "suites" that are being contemplated here are not limited to those registries...

For JsonWebSignature2020, it would work to point directing to IANA... but that would not be exhaustive since IANA does not define URDNA2015 which JsonWebSignature2020 (and other suites) rely on... and I suspect the RCH WG at W3C will eventually define.

What happens if we change the RDF Canonicalization in the future?

JsonWebSignature2030 uses IANA registered JWS, but with a new URDNA2032 canonicalization algorithm.

I don't see a way for folks to keep inventing new Data Integrity Suites without a registry that lets them talk about how they changed their canonicalization algorithm.

If you aren't changing canonicalization, IANA in sufficient... assuming you are leveraging JSON or CBOR.

msporny:
> We should be relying on algorithm assessment by experts and not trying to do this ourselves.

None of this text asserts that we will be doing this ourselves. We will be looking to authorities like CFRG and NIST to define safe parameters for any given year.

> For instance, entries in the IANA JOSE Algorithms Registry https://www.iana.org/assignments/jose/jose.xhtml#web-signature-encryption-algorithms include not just a specification reference but an assessment of the status of the algorithm (Required, Recommended, Optional, Prohibited), etc. that can be relied on by us, rather than trying to reinvent this wheel.

Sure, and that's one input. Another input is the COSE Algorithms registry. Another input is the safecurves site (or updated versions of such a site). Yet another input is FIPS. None of these sources, however, discuss data transformation/canonicalization algorithms and thus we cannot purely rely on pre-existing identifiers because those do not encode additional things that this specification does. We will need new identifiers (that, at a minimum, will use existing approved crypto).

> Like @mprorock , I'm also highly skeptical that adding a year to an algorithm name adds actionable information. For one thing, you can determine the year anyway by following the registry pointer to the defining specification and looking at the spec date, if you care. 

Are you asserting that most programmers go and read specifications or registries to determine if the cryptography that they're using is safe to use? Or do you think that they're more likely to copy-paste something off of stack overflow or a software library example? This specification is designing for the latter class of developers, not the former, because the former are not the ones that are responsible for most security vulnerabilities. Letting most developers choose cryptographic algorithmic parameters, or depending on them to educate themselves about the parameters is not as effective as ensuring that they cannot pick a wrong set of parameters in the first place.

@OR13 wrote:
> What happens if we change the RDF Canonicalization in the future?

Yes that, AND what happens if we change the format or add parameters in the future? This isn't theoretical, in the past several years we've changed the signature value from using `signatureValue` to `jws` to `proofValue` (as base64url) to `proofValue` as a multibase-encoded string, all without changing the base signature algorithm that's at play. Using parametric names doesn't work in this case, and in fact, it's been demonstrated (both at IETF and W3C) that parametric names are not uniformly enforced (case in point, the discrepancies between the JOSE, COSE, and Signing HTTP Messages cryptographic algorithm registries).

> Furthermore, you'll still have to consult the registry entry to determine whether the algorithm has been deprecated. A date isn't going to give you that. When something isn't useful, we should leave it out.

A date will hint at whether or not what you're using might be out of date... a parameter does not do that in a universal way. A future PR to this specification will attempt to assert that libraries MUST throw an error if a deprecated suite is used. We may go as far to say that any suite older than X years should be automatically deprecated (as @christophera has suggested multiple times)... but, again, that's for a future PR. 

The point here is that we are not going to depend on the date alone to accomplish the design goals of the Data Integrity spec -- we'll have multiple layers of protection to minimize the possibility of misuse of cryptographic suites.

index.html
mprorock: > do you oppose non-parametric names like "RS1" and "P256" and "RSA1_5"

definitely not - provided it is indicative of the scheme / format / or another key functional component to identify what is in use - but I typically like that kind of thing as a "mod" of sorts appended or pre-pended to the alg + params

This is one of those tough things as there is not a clear "right answer" and what we pick can and will have lasting effects.

mprorock:
> @selfissued @mprorock your suggestion to only use the other registries won't work for all the data integrity cases...
> 
> Since the "suites" that are being contemplated here are not limited to those registries...

extremely helpful context - thanks

index.html
```diff
@@ -1768,6 +1777,105 @@ <h2>Security Considerations</h2>
...
+Date-based versioning specifies a particular release date for a specific
+cryptographic suite. The benefit of a date, such as a year, is that it is
+immediately clear to a developer if the date is relatively old or new. Seeing
+an old date might prompt the developer to go searching for a newer
+cryptographic suite, where as a parametric or number-based versioning scheme
+might not. The downside of a date-based version is that some cryptographic
+suites might not expire for 5-10 years, prompting the developer to go
+searching for a newer cryptographic suite only to not find one that is newer.
+While this might be an inconvenience, it is one that results in safer
+ecosystem behavior.
```
mprorock, line 1843: what about the case where they go find a newer one, that is similar, but not exactly the same and assume that newer means better when it may not in fact be better or might contain a flaw?

index.html
```diff
@@ -1768,6 +1777,105 @@ <h2>Security Considerations</h2>
...
+versioning mechanism, however, is not enough by itself. All cryptographic
+suites that follow this specification are intended to be registered
+[[?VC-EXTENSION-REGISTRY]] in a way that clearly signal which cryptosuites
+are deprecated, standardized, or experimental. Cryptosuite registration will
+follow CFRG, IETF, NIST, FIPS, and safecurves guidance. Use of deprecated suites
+are expected to throw errors in implementations unless a `useUnsafeCryptosuites`
+option is used specifying exactly the unsafe cryptosuite to use.
+Use of experimental suites are expected to throw errors in implementations
+unless a `useExperimentalCryptosuites` option is used used specifying exactly
+the experimental cryptosuite to use.
```
mprorock: this is excellent!

index.html
```diff
@@ -1768,6 +1777,105 @@ <h2>Security Considerations</h2>
...
+Developers are urged to always refer to the [[?VC-EXTENSION-REGISTRY]] for
+the latest standardized cryptography suites as well as cross-checking their
+usage against, at least, the
+<a href="https://www.iana.org/assignments/cose/cose.xhtml#algorithms">
+COSE Algorithms Registry</a>, the
+<a href="https://www.iana.org/assignments/jose/jose.xhtml#web-signature-encryption-algorithms">
+JOSE Algorithms Registry</a>, and the latest
+<a href="https://csrc.nist.gov/publications/detail/fips/186/4/final">
+FIPS Digital Signature Standard
+</a> guidance.
```
```suggestion
</a> guidance.  Depending on your jurisdiction you may also need to consider additional guidance from <a href="https://www.etsi.org/deliver/etsi_ts/119300_119399/119312/01.02.01_60/ts_119312v010201p.pdf">ETSI</a> or other regulatory and standards bodies.
```

mprorock (changes requested):
see question regarding years in naming, as well as an ETSI reference suggestion (though they often point back to NIST or ISO) - otherwise this is looking very close

index.html
dlongley: I don't think any of the options here result in perfect outcomes. But we're looking for improvements over an existing system, not necessarily perfection. And it does seem like the case you're describing here is the least likely to occur and the most easily remedied, doesn't it?

I presume the cryptosuite you're describing is new, unrecommended, and non-standard. This is because the user in this scenario is only choosing based on the latest year. If it wasn't new, some other suite with a more recent year would be chosen avoiding the problem (or putting us into another case). If it was recommended, we could remove that recommendation status (or it wouldn't actually be a problem because it's properly recommended). If it was standard, we could update the standard and / or issue a newer cryptosuite that remedies the problem.

In other words, if the date-based scheme results in:

* Fewer people using crypto that's so old it is now unsafe.
* Fewer people combining crypto parameters in unsafe ways.
* Fewer implementation bugs because of less optionality.
* Fewer debates (in aggregate in the community) over which crypto params are safe / best to combine (do this only every few years when considering creating a new cryptosuite and ideally include in the spec the rationale behind the choices).
* More people becoming more aware that crypto has a shelf-life and requesting / performing more audits.
* More people using a new (and unrecommended / non-standard) cryptosuite that isn't as good as an older one (the case you describe, right?).

That seems to be, overall, a win vs. not doing it.

index.html
dlongley: +1, but typo with repeated word: "used used".

index.html
msporny: > definitely not - provided it is indicative of the scheme / format / or another key functional component to identify what is in use 

RS1 does not indicate that 1) it is an encryption scheme identifier, 2) it uses PKCS1-v1_5, 3) it uses SHA-1. In other words, it's not parametric at all, but you seem to indicate that you're not opposed to it.

P256 does not indicate that 1) it is a curve identifier, 2) is easily confused with PS256, which is a signing algorithm.

PS256 does not indicate that 1) it is a signing algorithm identifier, 2) is easily confused with P256, 3) uses the MGF1 mask generation function, 4) uses SHA-256 for BOTH the pre-hash and the MGF function

RSA1_5 does not indicate that 1) it is a CEK from 2003, 2) requires key sizes larger than 2048 bits.

So, if you are not opposed to those, you are not opposed to non-parametric approaches.

I'm trying to understand your mental model here and it's not clear to me. Why are you suggesting that we should do full parameterization, but then ok with items above that don't even come close to full parameterization?

index.html
```diff
@@ -1768,6 +1777,105 @@ <h2>Security Considerations</h2>
...
+year 2025, but might not be a safe choice in the year 2042. A date-based
+versioning mechanism, however, is not enough by itself. All cryptographic
+suites that follow this specification are intended to be registered
+[[?VC-EXTENSION-REGISTRY]] in a way that clearly signal which cryptosuites
+are deprecated, standardized, or experimental. Cryptosuite registration will
+follow CFRG, IETF, NIST, FIPS, and safecurves guidance. Use of deprecated suites
+are expected to throw errors in implementations unless a `useUnsafeCryptosuites`
+option is used specifying exactly the unsafe cryptosuite to use.
+Use of experimental suites are expected to throw errors in implementations
+unless a `useExperimentalCryptosuites` option is used used specifying exactly
```
```suggestion
unless a `useExperimentalCryptosuites` option is used specifying exactly
```

msporny:
@OR13 @mprorock @selfissued I have marked the type of versioning and extension registry governance items as issues in the spec text (instead of asserting them as conclusions). I have added text to note that other sources of cryptographic guidance should be cross-checked. AFAICT, there are now tracking issues for all of the items that each of you identified where there was an objection, which is where the debate can continue. I believe I have addressed as much as I can in this PR and am requesting a re-review now.

index.html
```diff
@@ -1755,6 +1777,108 @@ <h2>Security Considerations</h2>
...
+immediately clear to a developer if the date is relatively old or new. Seeing
+an old date might prompt the developer to go searching for a newer
+cryptographic suite, where as a parametric or number-based versioning scheme
+might not. The downside of a date-based version is that some cryptographic
+suites might not expire for 5-10 years, prompting the developer to go
+searching for a newer cryptographic suite only to not find one that is newer.
+While this might be an inconvenience, it is one that results in safer
+ecosystem behavior.
+        </p>
+        <p class="issue" data-number="38">
```
OR13, line 1845: This is good enough for me.

OR13 (approved):

index.html
msporny: Typo fixed in latest PR.

dlongley (approved):

dmitrizagidulin (approved):
👍

msporny:

PR #40, opened 2022-08-20
Add Security Considerations for appropriate layering of cryptographic options/parameters.

msporny:
This PR adds a security considerations section to guide the ecosystem on appropriately layering cryptographic options. This is meant to add the rationale for why the Data Integrity specification relies on cryptographic suites (and versioning) to provide protections against using dangerous options, and experimental and deprecated cryptography.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/40.html" title="Last updated on Aug 27, 2022, 7:12 PM UTC (1a508e6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/40/dd15452...1a508e6.html" title="Last updated on Aug 27, 2022, 7:12 PM UTC (1a508e6)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1903,10 +1903,10 @@ All options are often exposed to application developers.
         </p>
         <p>
 Application developers that use cryptographic libraries often do not have the
-same level of knowledge as cryptographic algorithm designers, specification
-authors, or cryptographic library implementers. This lack of knowledge can lead
-to an inappropriate selection of cryptographic parameters and options for a
-particular application.
+requisite cryptographic expertise and knowledge necessary to appropriately
+select cryptographic parameters and options for a given application. This lack
+of expertise can lead to an inappropriate selection of cryptographic parameters
+and options for a particular application.
         </p>
         <p>
 This specification sets the priority of constituencies to protect application
--- a/index.html
+++ b/index.html
@@ -1916,10 +1916,10 @@ priorities, the following recommendations are made:
         </p>
         <ul>
           <li>
-Cryptographic algorithm designers are advised to minimize the number of
-options and parameters to as few as possible to ensure that
-cryptographic library implementers have a more easily auditable security
-attack surface for their software libraries.
+Cryptographic algorithm designers are advised [[?RFC7696]] to minimize the
+number of options and parameters to as few as possible to ensure that
+cryptographic library implementers have a more easily auditable security attack
+surface for their software libraries.
           </li>
           <li>
 Cryptographic specification authors are advised to, if possible, further
--- a/index.html
+++ b/index.html
@@ -152,7 +152,7 @@ ol.algorithm li:before { font-weight: bold; counter-increment: numsection; conte
       <p>
 This specification describes mechanisms for ensuring the authenticity and
 integrity of Verifiable Credentials and similar types of constrained digital
-documents using cryptography, especially through the use of digital 
+documents using cryptography, especially through the use of digital
 signatures and related mathematical proofs.
       </p>
     </section>
@@ -1876,6 +1876,88 @@ FIPS Digital Signature Standard
 additional guidance from
 <a href="https://www.etsi.org/deliver/etsi_ts/119300_119399/119312/01.02.01_60/ts_119312v010201p.pdf">ETSI</a>
 or other regional regulatory and standards bodies.
+        </p>
+      </section>
+
+      <section>
+        <h3>Layer Cryptographic Options Appropriately</h3>
+
+        <p>
+Modern cryptographic algorithms provide a number of tunable parameters and
+options to ensure that the algorithms can meet different use case requirements.
+For example, embedded systems have limited processing and memory environments
+and might not have the resources to generate the strongest digital signatures
+for a given algorithm. Other environments, like financial trading systems,
+might only need to protect data for a day while the trade is occuring while
+other environments might need to protect data for multiple decades. To meet
+these needs, cryptographic algorithm designers often provide multiple ways to
+configure a cryptographic algorithm.
+        </p>
+        <p>
+Cryptographic library implementers often take the specifications created by
+cryptographic algorithm designers and specification authors and implement them
+such that all options are available to the application developers that use their
+libraries. This can be due to not knowing which combination of features a
+particular application developer might need for a given cryptographic algorithm.
+All options are often exposed to application developers.
+        </p>
+        <p>
+Application developers that use cryptographic libraries often do not have the
+same level of knowledge as cryptographic algorithm designers, specification
+authors, or cryptographic library implementers. This lack of knowledge can lead
+to an inappropriate selection of cryptographic parameters and options for a
+particular application.
+        </p>
+        <p>
+This specification sets the priority of constituencies to protect application
+developers over cryptographic library implementers over cryptographic
+specification authors over cryptographic algorithm designers. Given these
+priorities, the following recommendations are made:
+        </p>
+        <ul>
+          <li>
+Cryptographic algorithm designers are advised to minimize the number of
+options and parameters to as few as possible to ensure that
+cryptographic library implementers have a more easily auditable security
+attack surface for their software libraries.
+          </li>
+          <li>
+Cryptographic specification authors are advised to, if possible, further
+minimize the number of options and parameters to as few as possible to ensure
+cryptographic agility while also keeping the auditable security
+attack surface for downstream software libraries to a minimum.
+          </li>
+          <li>
+Cryptographic library implementers are advised to, if possible, provide known
+good combinations of options and parameters to application developers. By
+default, there would ideally be two pre-set configurations for any algorithmic
+class, such as Elliptic Curve Digital Signatures, with no ability (by default)
+to fine tune parameters and options. While library options can be provided to
+experts to fine tune the library, usage of those options among the general
+application developer population is to be discouraged.
+          </li>
+          <li>
+Application developers are advised to choose from a number of pre-set
+cryptography library configurations and to avoid modifying cryptographic
+options and parameters, or using experimental or deprecated cryptography.
+          </li>
+        </ul>
+
+        <p>
+The guidance above is provided to ensure that useful cryptographic options and
+parameters are provided at the lower layers of the architecture while not
+exposing those options and parameters to application developers that are ill
+suited to understand the benefits and drawbacks of each option.
+        </p>
+
+        <p class="issue"
+          title="Use of experimental and deprecated cryptography">
+The VCWG is seeking guidance on adding language to allow the use of experimental
+or deprecated cryptography. By default, those features will be disabled and will
+require the developer to specifically allow use on a per-cryptographic suite
+basis. There will be requirements for all implementing libraries to throw errors
+or warnings when deprecated or experimental options are selected without the
+appropriate override flags.
         </p>
       </section>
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1879,6 +1879,88 @@ <h3>Versioning Cryptography Suites</h3>
         </p>
       </section>
 
+      <section>
+        <h3>Layer Cryptographic Options Appropriately</h3>
```
dlongley: This seems to be more about:

```suggestion
        <h3>Prioritize Application Developers</h3>
```

This section seems to weave both of these ideas together: Cryptography is about protecting data, but good cryptographic primitives and implementations also protect application developers from their (often intentional) lack of knowledge about how the cryptography works. Application developers have other things they need to focus on. Additionally, good crypto tools and application of primitives must prioritize application developer use cases -- or else the tools will ultimately not be used. This may mean that where two goals are competing with one another: crypto implementation difficulty vs. crypto tool usefulness to application developers... that application developers should be prioritized -- but not so much so that security becomes infeasible. We may want to mention that test suites for the implementation of the crypto tools (like data integrity suites) can help in these situations?

Note: Also, it's better to avoid the word "appropriately" because it can be interpreted very subjectively -- "fit for purpose" with a discussion of the intended purpose(s) might avoid this problem.

dlongley (approved):

index.html
TallTed: I think this is not just about Application Developers, and I agree that "appropriately" can be problematic. Hence —
```suggestion
        <h3>Protecting Application Developers</h3>
```

index.html
```diff
@@ -1879,6 +1879,88 @@ <h3>Versioning Cryptography Suites</h3>
...
+
+        <p>
+Modern cryptographic algorithms provide a number of tunable parameters and
+options to ensure that the algorithms can meet different use case requirements.
```
```suggestion
options to ensure that the algorithms can meet the varied requirements of different use cases.
```

index.html
```diff
@@ -1879,6 +1879,88 @@ <h3>Versioning Cryptography Suites</h3>
...
+For example, embedded systems have limited processing and memory environments
+and might not have the resources to generate the strongest digital signatures
+for a given algorithm. Other environments, like financial trading systems,
+might only need to protect data for a day while the trade is occuring while
```
```suggestion
might only need to protect data for a day while the trade is occurring, while
```

index.html
```diff
@@ -1879,6 +1879,88 @@ <h3>Versioning Cryptography Suites</h3>
...
+other environments might need to protect data for multiple decades. To meet
+these needs, cryptographic algorithm designers often provide multiple ways to
+configure a cryptographic algorithm.
+        </p>
+        <p>
+Cryptographic library implementers often take the specifications created by
+cryptographic algorithm designers and specification authors and implement them
+such that all options are available to the application developers that use their
+libraries. This can be due to not knowing which combination of features a
+particular application developer might need for a given cryptographic algorithm.
```
```suggestion
particular application developer might need for a given cryptographic deployment.
```

index.html
```diff
@@ -1879,6 +1879,88 @@ <h3>Versioning Cryptography Suites</h3>
...
+attack surface for downstream software libraries to a minimum.
+          </li>
+          <li>
+Cryptographic library implementers are advised to, if possible, provide known
+good combinations of options and parameters to application developers. By
+default, there would ideally be two pre-set configurations for any algorithmic
+class, such as Elliptic Curve Digital Signatures, with no ability (by default)
+to fine tune parameters and options. While library options can be provided to
+experts to fine tune the library, usage of those options among the general
+application developer population is to be discouraged.
```
```suggestion
Cryptographic library implementers are advised to, if possible, provide known
good combinations of options and parameters to application developers. There
would ideally be two pre-set default configurations for any algorithmic
class, such as Elliptic Curve Digital Signatures, with no ability
to fine tune parameters and options when using these pre-sets. Library options can be provided to
experts to fine tune their use of the library, use of those options by the general
application developer population is to be discouraged.
```

index.html
```diff
@@ -1879,6 +1879,88 @@ <h3>Versioning Cryptography Suites</h3>
...
+          </li>
+          <li>
+Application developers are advised to choose from a number of pre-set
+cryptography library configurations and to avoid modifying cryptographic
+options and parameters, or using experimental or deprecated cryptography.
+          </li>
+        </ul>
+
+        <p>
+The guidance above is provided to ensure that useful cryptographic options and
```
```suggestion
The guidance above is meant to ensure that useful cryptographic options and
```

index.html
```diff
@@ -1879,6 +1879,88 @@ <h3>Versioning Cryptography Suites</h3>
...
+parameters are provided at the lower layers of the architecture while not
+exposing those options and parameters to application developers that are ill
+suited to understand the benefits and drawbacks of each option.
```
```suggestion
exposing those options and parameters to application developers who may not
fully understand the balancing benefits and drawbacks of each option.
```

index.html
```diff
@@ -1879,6 +1879,88 @@ <h3>Versioning Cryptography Suites</h3>
...
+        </p>
+
+        <p class="issue"
+          title="Use of experimental and deprecated cryptography">
+The VCWG is seeking guidance on adding language to allow the use of experimental
+or deprecated cryptography. By default, those features will be disabled and will
+require the developer to specifically allow use on a per-cryptographic suite
```
TallTed: Which `developer` here — library or application or other?

TallTed (changes requested):

index.html
```diff
@@ -1879,6 +1879,88 @@ <h3>Versioning Cryptography Suites</h3>
...
+        <p>
+Cryptographic library implementers often take the specifications created by
+cryptographic algorithm designers and specification authors and implement them
+such that all options are available to the application developers that use their
+libraries. This can be due to not knowing which combination of features a
+particular application developer might need for a given cryptographic algorithm.
+All options are often exposed to application developers.
+        </p>
+        <p>
+Application developers that use cryptographic libraries often do not have the
```
thereisnogabe, line 1905: I get the point you're making, but feel like some nuance is missed. A developers lack of mathematical understanding of [S-box construction](https://en.wikipedia.org/wiki/Rijndael_S-box) likely does not impact their ability to use AES affectively. Perhaps rephrasing "cryptographic algorithm designers" with "Security Practitioners" may work...

index.html
```diff
@@ -1879,6 +1879,88 @@ <h3>Versioning Cryptography Suites</h3>
...
+        </p>
+        <p>
+This specification sets the priority of constituencies to protect application
+developers over cryptographic library implementers over cryptographic
+specification authors over cryptographic algorithm designers. Given these
+priorities, the following recommendations are made:
+        </p>
+        <ul>
+          <li>
+Cryptographic algorithm designers are advised to minimize the number of
```
thereisnogabe: having this kind of suggestion in the spec really feels like the tail wagging the dog. I would be more comfortable focusing just on implementers of the spec which are often "cryptographic library implementers" and "application developers"

index.html
dlongley: > A developers lack of mathematical understanding of [S-box construction](https://en.wikipedia.org/wiki/Rijndael_S-box) likely does not impact their ability to use AES affectively.

True -- I think what we want to capture though are things like an application developer setting up their software to use AES-GCM, but because the chosen library also supports AES-CBC, an attacker can downgrade to it.

index.html
dlongley: Yeah, maybe this is meant to be "Cryptographic suite designers".

index.html
msporny: The titles need to be 3 words long or they wrap on most devices, we want something pithy. The title you provided is accurate, but might not result in something "sticky".

Suggesting "Prioritize Application Developers" as a "Security Consideration".

index.html
```suggestion
require the application developer to specifically allow use on a per-cryptographic suite
```

msporny: Good catch... it was meant to be "application", which I've included in the text now.

index.html
msporny: This section is intended to convey a full top-to-bottom philosophy. I'll note that this conversation has surfaced in CFRG before and [is currently being debated](https://mailarchive.ietf.org/arch/msg/cfrg/FpfQFAtqwn5gHPV619RLvMAu98I/). I believe that not including the foundational work that the IETF CFRG does would be problematic as there is a feedback cycle that exists between the high-level security protocols and the low-level security primitives. One of the problems with the current state of the industry is in the myriad of options that are provided at the low-level and then are needlessly exposed to application developers that don't know any better.

Explaining the complete philosophy is better than just explaining part of the philosophy, IMHO.

That said, we might be missing a layer, as "cryptographic library implementers" (libcrypto) are different from "cryptographic suite designers" (eddsa-2022), which are different from "cryptographic suite implementers" (di.js).

I'll try to take this into account and rewrite bits of this section. Suggestions at concrete text will help if you guys have any for this particular issue.

index.html
msporny: My concern with "Security practitioners" is that it's too broad... that's like the union of "cryptographic algorithm designers, cryptographic library implementers, cryptographic suite designers, cryptographic suite implementers", right?

If we break the "groups of people that do things" into too many subgroups, it might be hard to follow the argument... if we combine too many of them into too few groups, it might not resonate with readers ("something seems off with these groups").

Got any concrete language suggestions, given the above?

thereisnogabe: what about something like

```
Application developers that use cryptographic libraries often do not have the requisite cryptographic expertise and  knowledge necessary to appropriately select cryptographic parameters and options for a given application.
```

index.html
thereisnogabe: > not including the foundational work that the IETF CFRG does would be problematic as there is a feedback cycle that exists between the high-level security protocols and the low-level security primitives

Agree here. Is there a middle ground where we can reference (or directly quote) an IETF document to not seem as if we're making these prescriptions ourselves?

index.html
TallTed: moved to https://github.com/w3c/vc-data-integrity/pull/40#pullrequestreview-1085748275

index.html
```diff
@@ -1879,6 +1879,88 @@ <h3>Versioning Cryptography Suites</h3>
         </p>
       </section>
 
+      <section>
+        <h3>Prioritize Application Developers</h3>
```
TallTed: > [@msporny] The titles need to be 3 words long or they wrap on most devices, we want something pithy. The title you provided is accurate, but might not result in something "sticky".
>
> Suggesting "Prioritize Application Developers" as a "Security Consideration".

Hrm. That line wrap length is VERY limiting, and quite troubling! I'm revising my suggestion anyway. "Prioritize" still feels wrong. My revised suggestion of "Protecting Application Developers" doesn't feel quite right, either, as I think it's more like "Protecting Application Developers [and Users from Developer Footguns]"... Maybe saying this will spark inspiration in someone else.

```suggestion
        <h3>Protecting Application Developers</h3>
```

TallTed (changes requested):
That subsection title doesn't quite work for me yet...  Not a complete blocker but maybe it warrants an issue to be come back to later.

index.html
TallTed: This can be marked "resolved" (but I don't have that button)

index.html
msporny: Fixed in 3a788ad56421a496e0d6f0b8e91027ac8c3e9664.

index.html
msporny: Fixed in 1a508e62597d326f1ace75c594df29ec53eac56c.

msporny:

PR #43, opened 2022-08-21
Add Security Considerations regarding Data Opacity.

msporny:
This PR adds a security considerations section related to data opacity as it related to the design of cryptographic suites.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/43.html" title="Last updated on Aug 27, 2022, 7:38 PM UTC (f426230)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/43/b7c2833...f426230.html" title="Last updated on Aug 27, 2022, 7:38 PM UTC (f426230)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2108,34 +2108,35 @@ digitally signed data that has multiple nested base-encoded payloads.
 Similarly, a cryptographically protected payload that can be processed by
 multiple upstream networked systems increases the ability to properly layer
 security architectures. For example, if upstream systems do not have to
-repeatedly decode the incoming payload, it increases the ability for a system
-to distribute processing load by specializing upstream subsystems to actively
+repeatedly decode the incoming payload, it increases the ability for a system to
+distribute processing load by specializing upstream subsystems to actively
 combat attacks. While a digital signature needs to always be checked before
-taking action, other upstream checks can be performed &mdash; such as
-identifier-based rate limiting, signature expiration checking, or
-nonce/challenge checking &mdash; to reject obviously bad requests.
+taking substantive action, other upstream checks can be performed on transparent
+payloads &mdash; such as identifier-based rate limiting, signature expiration
+checking, or nonce/challenge checking &mdash; to reject obviously bad requests.
         </p>
         <p>
 Additionally, if a developer is not able to easily view data in a system, the
 ability to easily audit or debug system correctness is hampered. For example,
 requiring application developers to cut-and-paste base-encoded application data
 makes development more challenging and increases the chances that obvious bugs
-will be missed because every message needs to go through a base-decoding tool
-that is external to the system.
+will be missed because every message needs to go through a manually operated
+base-decoding tool.
         </p>
         <p>
 There are times, however, where the correct design decision is to make data
 opaque. Data that does not need to be processed by other application subsystems,
 as well as data that does not need to be modified or accessed by an application
-developer, can be stored in opaque formats. Examples include digital signature
-values, cryptographic key parameters, and other data fields that only need to be
-accessed by a cryptographic library and need not be modified by the application
-developer. There are also examples where data opacity is appropriate when the
-underlying subsystem does not expose the application developer to the underlying
-complexity of the opaque data, such as databases that perform encryption at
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2073,6 +2073,42 @@ <h3>Transformations</h3>
...
+that can be processed by multiple upstream networked systems without having to
+repeatedly decode the payload increases the ability for a system to share
+processing load. Additionally, if a developer is not able to easily view data in
+a system, the ability to easily audit or debug system correctness is hampered.
+        </p>
+        <p>
+There are times, however, where the correct design decision is to make data
+opaque. Data that does not need to be processed by other application subsystems
+or data that does not need to be modified or accessed by an application
+developer can be stored in opaque formats. Examples include digital signature
```
```suggestion
opaque. Data that does not need to be processed by other application subsystems,
as well as data that does not need to be modified or accessed by an application
developer, can be stored in opaque formats. Examples include digital signature
```

index.html
```diff
@@ -2073,6 +2073,42 @@ <h3>Transformations</h3>
...
+values, cryptographic key parameters, and other data fields that only need to
+be accessed by a cryptographic library and need not be modified by the
+application developer.
+        </p>
+        <p>
+This specification strives to provide an architecture where application data
+remains in its native format and is not made opaque while other cryptographic
```
```suggestion
remains in its native format and is not made opaque, while other cryptographic
```

index.html
```diff
@@ -2073,6 +2073,42 @@ <h3>Transformations</h3>
...
+data, such as digital signatures, are kept in their opaque binary encoded form.
+Cryptographic suite implementers are urged to consider data opacity when
+designing their suites and weigh the design trade-offs when making application
+data opaque and when providing access to cryptographic data at the application
```
```suggestion
designing their suites, and to weigh the design trade-offs when making application
data opaque versus providing access to cryptographic data at the application
```

TallTed (changes requested):

index.html
```diff
@@ -2073,6 +2073,42 @@ <h3>Transformations</h3>
...
+        <h3>Data Opacity</h3>
+
+        <p>
+The inspectability of application data has effects on system efficiency and
+developer productivity. When cryptographically protected application data, such
+as base-encoded binary data, is not easily processed by application subsystems,
+such as databases, it increases the effort of working with with the
+cryptographically protected information. For example, a cryptographically
+protected payload that can be natively stored and indexed by a database will
+result in a simpler system. Similarly, a cryptographically protected payload
```
dlongley: Not talking about the "unseen" complexities here by only mentioning a "simpler system" may not address people's concerns -- we'll see. Some of the "unseen" benefits include not losing integrity proofs when data is transformed:

* allowing application developers to avoid creating complex systems that duplicate data (having one copy that is data integrity protected and another that is essentially not... but is presumed to have been checked and is what the application actually runs on)
* allowing application developers to avoid each creating their own solutions (which may differ and therefore generate more work in aggregate) to solve the impedance mismatch of using the data vs. maintaining one or more data integrity proofs (with multiple proofs perhaps implying multiple nested copies of the data).

dlongley (approved):

index.html
TallTed: It may also be worth mentioning that there are encrypted-at-rest (EAR) DBMS out there (Virtuoso is optionally one such), such that the DBMS natively stores and indexes the data, but it's not extractable from that storage without the relevant key(s).  This can result in a deceptively simple system, i.e., a system which appears simple to the casual observer, but is significantly more complex below the surface.

msporny: @dlongley wrote:
> Not talking about the "unseen" complexities here by only mentioning a "simpler system" may not address people's concerns

Added language in 60a61cb2124da4822537cc7f74cb9c664159ca14 to address your concern.

@TallTed wrote:
> It may also be worth mentioning that there are encrypted-at-rest (EAR) DBMS out there

Added language in 007125d7d9e4ba2dc983c3b2c1fa1481d01851dd to address your concern.

TallTed (approved):

msporny:

PR #51, opened 2022-09-17
Rename cryptographic hardening to cryptographic layering.

msporny:
This PR renames "cryptographic hardening" to "cryptographic layering" based on some requests from reviewers from SRI and CFRG.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/51.html" title="Last updated on Sep 25, 2022, 9:37 AM UTC (f12d0c2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/51/9eddbc8...f12d0c2.html" title="Last updated on Sep 25, 2022, 9:37 AM UTC (f12d0c2)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -347,9 +347,9 @@ for further rationale.
 Since digital proof mechanisms might be compromised without warning due to
 technological advancements, it is important that <a>cryptographic suites</a>
 provide multiple layers of protection and can be rapidly upgraded. This
-specification provides for both algorithmic agility and cryptographic hardening,
+specification provides for both algorithmic agility and cryptographic layering,
 while still keeping the digital proof format easy for developers to understand
-and use. See section <a href="#agility-and-hardening"></a> to understand the
+and use. See section <a href="#agility-and-layering"></a> to understand the
 particulars.
           </dd>
           <dt>Progressive Extensibility</dt>
@@ -2087,7 +2087,7 @@ appropriate override flags.
       </section>
 
       <section>
-        <h3>Agility and Hardening</h3>
+        <h3>Agility and Layering</h3>
 
         <p>
 <dfn class="lint-ignore">Cryptographic agility</dfn> is a practice by which one designs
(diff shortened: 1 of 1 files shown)


dlongley (approved):
I wonder if "cryptographic reinforcement" would be another option.

thereisnogabe (approved):

TallTed (approved):

mprorock (approved):
layering is much better than hardening (which has its own definition) - I would note that we probably should be more explicit here, with language changed from:
`This specification provides for both algorithmic agility and cryptographic layering,
while still keeping the digital proof format easy for developers...` to `This
specification provides for both algorithmic agility and use of multiple cryptographic algorithms,
while still keeping the digital proof format easy for developers`

layering implies to me one method, alg, etc over the top of another, which is not typically what is going on here - mutliple signatures or proofs attached to a plaintext input is more what is going on - maybe `multi-sig` or something is better than layering?

mkhraisha:
Would something closer to `multi-signature` work? 
or would it have connotations that are associated with multiple parties providing independent signatures?

TallTed:
Continuing draft wordsmithing from @mprorock's https://github.com/w3c/vc-data-integrity/pull/51#pullrequestreview-1115533630... Maybe —

`This specification provides for both algorithmic agility (by which we mean easy switching from one algorithm to another) and cryptographic layering (by which we mean simultaneous use of multiple cryptographic algorithms, typically in parallel, such that any of those used to encrypt or sign may be used to decrypt or verify without reliance on or requirement of the others), while still keeping the digital proof format easy for developers.`

msporny:
> @mprorock and @mkhraisha suggested using multi-signatures and multi-proofs and noted "would it have connotations that are associated with multiple parties providing independent signatures?"

Yes, that's the primary reason I stayed away from the multisig wording, as it typically implies different parties signing the same data vs. the same party signing the same data using different mechanisms. We might want to avoid multisignatures because some entities might sign and then anchor (two proofs, one a signature, one an anchoring proof). Multiproofs might work, but then suggest that multiple parties are signing (as multisig typically does). 

I'm going to try applying something to the effect of what @TallTed suggested (as an improvement on what @mprorock suggested). PR update forthcoming.

msporny:
Implemented in https://github.com/w3c/vc-data-integrity/pull/51/commits/f12d0c2e42433f71316d88940f3d9b2e5dfdc2d7.

msporny:

index.html
```diff
@@ -2108,24 +2108,30 @@ <h3>Agility and Hardening</h3>
...
 provide a mechanism that could survive the failure of two of these three digital
 signature algorithms. When a particular cryptographic protection is compromised,
 such as an RSA digital signature using 768-bit keys, systems can still utilize
 the non-compromised cryptographic protections to continue to protect the
-information.
+information. Developers are urged to take advantage of this feature for all
+signed content that might need to be protected for a year or longer.
+        </p>
+        <p>
+This specification provides for both forms of agility. It provides for
```
TallTed, line 2126: @msporny — `This specification provides for both forms of agility.` doesn't quite make sense when those forms are `cryptographic agility` and `cryptographic layering` (which isn't so much "agile", to my mind).  Maybe `both forms of cryptographic enhancement`?

PR #52, opened 2022-09-17
Add content to section on unlinkability

msporny:
This PR replaces the issue marker about unlinkability with content describing what unlinkability is, what to consider wrt unlinkability, and how to achieve some level of it.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/52.html" title="Last updated on Oct 1, 2022, 8:26 AM UTC (5ddec64)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/52/1184fad...5ddec64.html" title="Last updated on Oct 1, 2022, 8:26 AM UTC (5ddec64)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2384,27 +2384,25 @@ full name and address are not being shared.
         </p>
 
         <p>
-The creator of an initial message that they expect to be selectively disclosed
-needs to use a capable cryptosuite and enable selective disclosure on the
-message in order for a recipient of that message to use the technique on future
-disclosures. The creator might also need to make it mandatory to disclose
-certain parts of a message. The recipient of the message that desires to
-selectively disclose the original message needs to then utilize software that is
-capable of performing selective disclosure. Not all software or cryptosuites are
-capable of providing selective disclosure. An example of a cryptosuite that does
-support selective disclosure is `bbs-2022`.
+Not all software or cryptosuites are capable of providing selective disclosure.
+If the author of a message wishes it to be selectively disclosable by
+its recipient, then they need to both use a capable cryptosuite and enable
+selective disclosure on the specific message. They might also make it mandatory
+to disclose certain parts of the message. A recipient that wants to selectively
+disclose the message needs to utilize software that is able to perform the
+technique. An example of a cryptosuite that supports selective disclosure is
+`bbs-2022`.
         </p>
 
         <p>
-It is important to realize that selective disclosure does not always mean that
-all aspects of privacy are preserved. It is possible to selectively disclose
-information in a way that does not preserve privacy. For example, one might want
-to disclose the inspection results related to a shipment, which include the
-shipment identifier or lot number, which might have to be correlatable due to
-regulatory requirements. However, disclosure of the entire inspection result
-might not be required as selectively disclosing just the pass/fail status could
-be deemed as adequate. For more information on disclosing information while
-preserving privacy, see Section <a href="#unlinkability"></a>.
+It is possible to selectively disclose information in a way that does not preserve
+unlinkability. For example, one might want to disclose the inspection results related
+to a shipment, which include the shipment identifier or lot number, which might
+have to be correlatable due to regulatory requirements. However, disclosure of
+the entire inspection result might not be required as selectively disclosing just
+the pass/fail status could be deemed as adequate. For more information on
+disclosing information while preserving privacy, see
+Section <a href="#unlinkability"></a>.
         </p>
       </section>
 


index.html
```diff
@@ -2316,11 +2316,39 @@ <h2>Privacy Considerations</h2>
...
-digital signature is a correlatable identifier. There are digital signature
-schemes that provide uncorrelatable digital signatures.
+        <p>
+When the contents of a digitally signed payload contains a correlatable
+identifier, that identifier can be used for tracking purposes by a verifier,
+which might lead to a violation of an expected level of privacy. An example of
+such an identifier is a shopping loyalty card number. However, some of these
+identifiers might not be immediately apparent, such as a static digital
+signature or a cryptographic hash of an image, which can also be used as a
+correlatable identifier.
```
dlongley: Some wordsmithing to:

* Bring in a definition for "correlatable" to replace perhaps redundant statement
* Better signal that some correlation might actually be expected (and not unwanton)
* Decouple a loyalty card number from the unwanton correlation case since people might want their purchasing behavior to be correlated in some fashion for better loyalty rewards

```suggestion
When a digitally-signed payload contains an identifier that is seen by multiple
verifiers, it becomes a point of correlation. An example of such an identifier is
a shopping loyalty card number. Correlatable identifiers can be used for tracking
purposes by verifiers, which can sometimes violate privacy expectations. The
fact that some identifiers can be used for tracking might not be immediately
apparent. Examples of such correlatable identifiers include, but are not limited
to, a static digital signature or a cryptographic hash of an image.
```

index.html
```diff
@@ -2316,11 +2316,39 @@ <h2>Privacy Considerations</h2>
...
+        </p>
+
+        <p>
+It is possible to create a digitally signed payload that does not have any
```
```suggestion
It is possible to create a digitally-signed payload that does not have any
```

index.html
```diff
@@ -2316,11 +2316,39 @@ <h2>Privacy Considerations</h2>
...
+correlatable tracking identifiers while also providing some level of assurance
+that the payload is trustworthy for a given interaction. This characteristic is
+called <dfn>unlinkability</dfn> which ensures that no correlatable identifiers
+are used in a digitally signed payload while providing some level of trust.
```
```suggestion
are used in a digitally-signed payload while providing some level of trust.
```

index.html
```diff
@@ -2316,11 +2316,39 @@ <h2>Privacy Considerations</h2>
...
+privacy such that it becomes practically impossible to correlate the entity that
+expects some level of privacy in the interaction.
+        </p>
+
+        <p>
+A variety of methods can be used to achieve unlinkability. These methods
+include the utilization of single use bearer tokens, carefully crafted
+messages that utilize attributes that ensure an adequate level of herd privacy,
+and the use of cryptosuites that enable the entity presenting a message to
+re-generate new signatures while not compromising the trust in the message
```
```suggestion
regenerate new signatures while not compromising the trust in the message
```

index.html
dlongley: I feel like we ought to say something like "Verifiers are encouraged to carefully analyze what level of trust they are able to achieve when accepting such payloads." "Some" could also mean "insufficient for the use case".

dlongley (approved):
Approving ... but with some suggestions. Thanks!

index.html
thereisnogabe: what about expanding past "identifier" to be "a piece or pieces of data that may be seen by multiple verifiers" -- foreseeably even if an `id` property is not shared, there could be n other fields that give sufficient information in order to correlate.

index.html
thereisnogabe: I feel like unlinkability is a potentially dangerous term. it is not solely up to the construction of the credential to provide this feature. there could be ways to construct data in a credential itself that make sharing between colluding verifiers less risky. of course this is up to the issuers to be aware of and using such a scheme.

index.html
TallTed: Upon re-reading, neither `regenerate new signatures` nor `re-generate new signatures` works. It should be either `generate new signatures` or `regenerate signatures`, of which I'm inclined toward the former.

index.html
TallTed: @dlongley — I *think* you meant [`and not unwanton` and `unwanton correlation`](https://github.com/w3c/vc-data-integrity/pull/52/files#r973764166) to say `and not unwanted` and `unwanted correlation`, respectively?

TallTed: Further on @dlongley's suggestion, in line 2321, I'd change `it becomes a point` to `that identifier becomes a point` (because the `it` *could* refer back to the entire payload, which I think is not intended).

index.html
```suggestion
are used in a digitally-signed payload while still providing some level of trust,
the sufficiency of which must be determined by each verifier.
```

index.html
```diff
@@ -2316,11 +2316,39 @@ <h2>Privacy Considerations</h2>
...
+correlatable tracking identifiers while also providing some level of assurance
+that the payload is trustworthy for a given interaction. This characteristic is
+called <dfn>unlinkability</dfn> which ensures that no correlatable identifiers
+are used in a digitally signed payload while providing some level of trust.
+        </p>
+
+        <p>
+There are at least two aspects to providing unlinkability. The first aspect is
+to 1) ensure that every data field used in the message is never repeated in a
+future message, or 2) ensure that any repeated field(s) provide adequate herd
```
TallTed: I can't tell whether the `at least two aspects` is meant to refer to the `1)` and `2)`, or to `the first` and an unstated `second`.  That knowledge is necessary to rephrase this paragraph, which rephrasing is currently necessary.

TallTed (changes requested):
It'll need another reading after the suggestions and comments above are addressed.

mprorock (changes requested):
we need some intro language for what kind of cases need unlinkability vs those that require the ability to link things

index.html
brentzundel: `unlinkability` is a term used in cryptographic literature to describe a process wherin events that occur, when observed by an attacker, do not appear any more related after the process completes than they did before. Before unlinkability can be claimed, the attacker model needs to be specified, i.e., we would need to be very specific about the context within which we are claiming unlinkability.

Avoidance of correlating identifiers is one aspect of unlinkability, but there is more to it than that, and it may be better to avoid the term, or hedge things by saying stuff like "avoiding correlating identifiers provides some degree of unlinkability"

index.html
mkhraisha: +1 to replace `identifier` with a more general term. "a piece or pieces of data", tried rewording Dave's statement
```suggestion
When a digitally-signed payload contains a piece or pieces of data that are seen by multiple
verifiers, it becomes a point of correlation. An example is
a shopping loyalty card number. Correlatable data can be used for tracking
purposes by verifiers, which can sometimes violate privacy expectations. The
fact that some pieces of data can be used for tracking might not be immediately
apparent. Examples of such correlatable data include, but are not limited
to, a static digital signature or a cryptographic hash of an image
```

index.html
```diff
@@ -2316,11 +2316,39 @@ <h2>Privacy Considerations</h2>
...
+identifiers might not be immediately apparent, such as a static digital
+signature or a cryptographic hash of an image, which can also be used as a
+correlatable identifier.
+        </p>
+
+        <p>
+It is possible to create a digitally signed payload that does not have any
+correlatable tracking identifiers while also providing some level of assurance
+that the payload is trustworthy for a given interaction. This characteristic is
+called <dfn>unlinkability</dfn> which ensures that no correlatable identifiers
```
mkhraisha: if we're changing from `correlatable identifiers` to `correlatable data` or similar, there are a few places for this change this comment is just a reminder to make all those changes.

index.html
```diff
@@ -2316,11 +2316,39 @@ <h2>Privacy Considerations</h2>
...
+are used in a digitally signed payload while providing some level of trust.
+        </p>
+
+        <p>
+There are at least two aspects to providing unlinkability. The first aspect is
```
mkhraisha: This reads like you need both, but the OR statement makes it seem like you need one or the other.

TallTed: Yes.  See [my previous comment](https://github.com/w3c/vc-data-integrity/pull/52#pullrequestreview-1114082060), above.

index.html
msporny: Fixed in b5df82a148c7cf11d804041fbe39bc0363526951.

index.html
msporny: Reworded in 4994c5b.

index.html
msporny: Reworded in 4994c5b.

msporny:
@mprorock wrote:
> we need some intro language for what kind of cases need unlinkability vs those that require the ability to link things

Fixed in db3cdb5.

index.html
```diff
@@ -2322,11 +2322,48 @@ <h2>Privacy Considerations</h2>
...
+It is possible to create a digitally-signed payload that does not have any
+correlatable tracking data while also providing some level of assurance that the
+payload is trustworthy for a given interaction. This characteristic is called
+<dfn>unlinkability</dfn> which ensures that no correlatable data are used in a
+digitally-signed payload while still providing some level of trust, the
+sufficiency of which must be determined by each verifier.
+        </p>
+
+        <p>
+It is important to understand that not all use cases require unlinkability.
```
```suggestion
It is important to understand that not all use cases require or even permit unlinkability.
```

index.html
```diff
@@ -2322,11 +2322,48 @@ <h2>Privacy Considerations</h2>
...
+regulatory or safety reasons, such as correlating organizations and individuals
+that are shipping and storing hazardous materials. Unlinkability is useful
+when there is an expectation of privacy for a particular interaction.
+        </p>
+
+        <p>
+There are at least two mechanisms that can provide some level of unlinkability.
+The first method is to ensure that every data value used in the message is
+never repeated in a future message. The second is to ensure that any repeated
+data value(s) provide adequate herd privacy such that it becomes practically
```
```suggestion
The first method is to ensure that no data value used in the message is
ever repeated in a future message. The second is to ensure that any repeated
data value provides adequate herd privacy such that it becomes practically
```

index.html
```diff
@@ -2322,11 +2322,48 @@ <h2>Privacy Considerations</h2>
...
+impossible to correlate the entity that expects some level of privacy in the
+interaction.
+        </p>
+
+        <p>
+A variety of methods can be used to achieve unlinkability. These methods include
+the utilization of single use bearer tokens, carefully crafted messages that
+utilize attributes that ensure an adequate level of herd privacy, and the use of
+cryptosuites that enable the entity presenting a message to regenerate new
+signatures while not compromising the trust in the message being presented.
```
```suggestion
A variety of methods can be used to achieve unlinkability. These methods include
crafting messages with a combination of single-use bearer tokens, attributes that
ensure an adequate level of herd privacy, and cryptosuites that enable the entity
presenting a message to regenerate new signatures while not compromising the trust
in the message being presented.
```

index.html
msporny: Applied in e0d07318018d369ec568f12b49598198e2b96f34.

index.html
msporny: Applied in e0d07318018d369ec568f12b49598198e2b96f34.

index.html
msporny: Applied in e0d07318018d369ec568f12b49598198e2b96f34.

msporny:
Editorial, multiple reviews, changes requested and made, no objections, merging (manually).

PR #57, opened 2022-10-01
Refactor the cryptosuite section.

msporny:
This PR refactors the cryptosuite section by:

1. Removing the old content about specifying cryptosuites using URL identifiers.
2. Adding a basic set of requirements for cryptosuite specifications, which could be used as registration criteria for a future registry.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/57.html" title="Last updated on Oct 9, 2022, 4:41 PM UTC (43e33f3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/57/b9c2d07...43e33f3.html" title="Last updated on Oct 9, 2022, 4:41 PM UTC (43e33f3)">Diff</a>


index.html
```diff
@@ -1489,90 +1489,53 @@ <h2>Cryptographic Suites</h2>
...
+The specification MUST detail the algorithms used to perform cryptographic
+protection of the data.
+        </li>
+        <li>
+The specification MUST detail the algorithms used to perform cryptographic
+verification of the data.
+        </li>
+        <li>
+The specification MUST contain a Security Considerations section detailing
+privacy considerations specific to the cryptographic suite.
```
```suggestion
security considerations specific to the cryptographic suite.
```

TallTed (changes requested):
Looks good, modulo one piece of internal inconsistency, changed here.

index.html
```diff
@@ -1489,90 +1489,53 @@ <h2>Cryptographic Suites</h2>
...
+        </li>
+        <li>
+The specification MUST contain a Privacy Considerations section detailing
+privacy considerations specific to the cryptographic suite.
+        </li>
+        <li>
+The specification MUST provide an interoperability test suite to determine if
+a given implementation is conformant with the specification.
```
dlongley: I don't know how the specification provides this -- unless this refers to inline test vectors, perhaps. This language might need to be tweaked a little, but approving for now.

dlongley (approved):

index.html
msporny: Yes, you're right, the spec can't provide this. What I was trying to get to was some sort of higher bar for registration -- where you have to demonstrate some level of interoperability to get into the registry.

Perhaps what we should do here is strike the requirement and make it a part of the registration process OR say that the specification MUST point to an interoperability test suite report (which you can do via ReSpec).

I'm going to strike this particular normative language for now because it'll probably lead to a contentious discussion.

Reworded as an issue in a6e348b39b3433d863d17c1418b0ebf3abb36e96.

TallTed (approved):

index.html
mprorock: we ok with a SHOULD here?

index.html
```diff
@@ -1489,90 +1489,60 @@ <h2>Cryptographic Suites</h2>
...
+The specification MUST detail the algorithms used to perform cryptographic
+hashing to the data to be protected.
+        </li>
+        <li>
+The specification MUST detail the algorithms used to perform cryptographic
+protection of the data.
+        </li>
+        <li>
+The specification MUST detail the algorithms used to perform cryptographic
+verification of the data.
```
```suggestion
The specification MUST detail the algorithms, parameters, and other necessary details used to perform cryptographic verification of the data.
```

mprorock (approved):
approved with minor suggestion

index.html
TallTed: I have no particular argument with this suggestion, but I submit that if this change is made, it should also be echoed in all the nearby list items.

index.html
msporny: I'd personally be ok w/ a SHOULD, though it lowers the bar to getting an entry into the registry AND it puts the registry maintainers in the critical path wrt. making "judgement calls", which have gotten us in trouble in the past. Given that this text is already in an issue marker, noting that it's controversial, I'm going to stick with the MUST for now and see if we get serious push-back. The text in the issue marker now suggests:

> The specification MUST provide a link to an interoperability test report to document which implementations are conformant with the cryptographic suite specification.

Given the nature of cryptosuites, not providing such a test report (with a single entry in it) feels like a low enough bar already. I would expect that we don't want registry entries to be taken up with cryptosuite designers that are unwilling to write conformance test suites for their cryptosuites, no?

msporny:
Substantive, normative, multiple reviews, changes requested and made, no objections, merging.

PR #59, opened 2022-10-01
Refactor examples to use DataIntegrityProof

msporny:
This PR updates the examples to align with the new `DataIntegrityProof` section.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/59.html" title="Last updated on Oct 9, 2022, 5:17 PM UTC (0bcd83c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/59/6bcc7f1...0bcd83c.html" title="Last updated on Oct 9, 2022, 5:17 PM UTC (0bcd83c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -549,7 +549,7 @@ by adding the parameters outlined in this section:
   {
     "@context": [
       {"title": "https://schema.org/title"},
-      "https://w3id.org/security/suites/data-integrity/v1"
+      "https://w3id.org/security/data-integrity/v1"
     ],
     "title": "Hello world!",
     "proof": {
@@ -601,7 +601,7 @@ ibsa-2022, and jws-2022.
 
       <pre class="example" title="A DataIntegrityProof example using a NIST ECDSA 2022 Cryptosuite">
 {
-  "@context": ["https://w3id.org/security/suites/data-integrity/v1"],
+  "@context": ["https://w3id.org/security/data-integrity/v1"],
   "type": "DataIntegrityProof",
   "cryptosuite": "ecdsa-2022",
   "created": "2022-11-29T20:35:38Z",
@@ -656,7 +656,7 @@ with the <code>proof</code> key in a document.
 {
   "@context": [
     {"title": "https://schema.org/title"},
-    "https://w3id.org/security/suites/data-integrity/v1"
+    "https://w3id.org/security/data-integrity/v1"
 ],
   "title": "Hello world!",
   "proof": [{
@@ -697,7 +697,7 @@ represented by associating an ordered list of proofs with the
 {
   "@context": [
     {"title": "https://schema.org/title"},
-    "https://w3id.org/security/suites/data-integrity/v1"
+    "https://w3id.org/security/data-integrity/v1"
 ],
   "title": "Hello world!",
   "proofChain": [{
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -519,10 +520,10 @@ <h3>Proofs</h3>
         </pre>
 
         <p>
-The proof example above uses the <code>JcsSignature2020</code>
-<a>proof type</a> to produce a verifiable digital proof by canonicalizing the
-input data using the JSON Canonicalization Scheme [[RFC8785]] and then
-digitally signing it using an Ed25519 elliptic curve signature.
+The proof example above suggests an exemplary <code>json-signature-2022</code>
```
```suggestion
The proof example above suggests an example <code>json-signature-2022</code>
```

TallTed (changes requested):
One word change. "Examplary" suggests "superlative" which I don't think is intended. "Example" does echo the earlier part of the sentence, but I think that's OK.

index.html
```suggestion
The proof example above suggests a fictitious <code>json-signature-2022</code>
```

@TallTed what about "ficticious" instead? It's closer to what I was intending. We're trying to convey that "the thing above doesn't exist, we're only using it as an example that presumptively does X and Y." 

On the other hand, I don't really like starting the spec out with a fictitious example.

As a related aside, one of the concepts that's attempting to be conveyed is "You don't have to use Linked Data to use Data Integrity. Data Integrity works for both plain 'ol JSON as well as JSON-LD (or YAML, CBOR, etc. -- but that's a stretch and we might tackle in a few years if DI is successful) ." ... but perhaps we should really spell that out instead of having people trying to read between the lines.

Thoughts?

msporny: See https://github.com/w3c/vc-data-integrity/pull/59/files#r986298535 for an alternate suggestion.

TallTed: I'm OK with `fictitious`, tho optimally that fiction would carry through all layers, so the fictitious example looks as real as possible, while NOT being real (and having any descriptions that exist for any entities used also indicate that fictitiousness).

This is one of the reasons I fight to use reserved `example` domain names, phone numbers, etc., wherever they're meant to appear functional for the functionality we need while not carrying through to the physical world.

msporny: What if we were to rename `json-signature-2022` to `example-signature-2022`? Would that be better?

TallTed: Much!

index.html
```diff
@@ -595,10 +597,10 @@ <h3>Proofs</h3>
 ibsa-2022, and jws-2022.
       </p>
 
-      <pre class="example" title="A DataIntegritySignature example using a NIST ECDSA 2022 Cryptosuite">
+      <pre class="example" title="A DataIntegrityProof example using a NIST ECDSA 2022 Cryptosuite">
 {
-  "@context": ["https://w3id.org/security/data-integrity/v1"],
-  "type": "DataIntegritySignature",
+  "@context": ["https://w3id.org/security/suites/data-integrity/v1"],
```
dlongley: The context URL: https://w3id.org/security/data-integrity/v1 is actually correct (today, anyway).

dlongley (approved):
Approving assuming Ted's suggestions will go in -- and please note the comment I made. Thanks!

mprorock (approved):
approved, but agree with Ted on example name changes

index.html
msporny: Fixed in f33680902c92e937fe7c629ac7263d011d576e7f.

index.html
msporny: Fixed in 0bcd83ce6237a68d217cd99441ec538d4efda5ca.

msporny:

PR #63, opened 2022-10-16
Remove the unnecessary Create Verify Hash Algorithm.

msporny:
This PR removes the unnecessary Create Verify Hash Algorithm, which was hard coded to presume digital signatures only. The hashing algorithm is now defined by cryptosuites, which greatly simplifies how much bouncing around is done between the algorithms in the Data Integrity specification and the cryptosuite specifications.

All of the text being removed here has been replaced with a single sentence in step 3 of section 4.1 Generate Proof: https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/63.html#generate-proof

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/63.html" title="Last updated on Oct 23, 2022, 8:14 PM UTC (33e83ae)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/63/57903ba...33e83ae.html" title="Last updated on Oct 23, 2022, 8:14 PM UTC (33e83ae)">Diff</a>


dlongley:
Hmm, is there a way we can keep this function so it can be reused by cryptosuites? It will be commonly reused and it would be a shame if every cryptosuite had to redefine it or cross link to some other cryptosuite that it would otherwise not depend on.

msporny:
> Hmm, is there a way we can keep this function so it can be reused by cryptosuites? It will be commonly reused and it would be a shame if every cryptosuite had to redefine it or cross link to some other cryptosuite that it would otherwise not depend on.

Every cryptosuite is going to have to define the cryptosuite-specific `generate proof algorithm`, `generate hash algorithm` , and `verify proof algorithm` anyway. Not all of them can share the `append the proof hash` algorithm, though I do agree that it /might/ be useful to define it in the Data Integrity spec. For other things, like for how BBS prepares its `hashData`, we are probably not going to allow the same algorithm used to hash the document to be used to hash the proof graph, right? 

What parts of the algorithm do you think are going to be common to most cryptosuites?

If I had to guess, it would be the part of the algorithm that transform + hashes the proof graph? I was going to try specifying that in the `eddsa-2022` spec and then see if I might be able to generalize that between that, ecdsa-2022 and bbs-2023... though, I guess I could try to guess and write that now in this PR? Is that what you're asking me to do?

index.html
```diff
@@ -1642,9 +1642,10 @@ <h3>Generate Proof</h3>
 as inputs to the algorithm.
         </li>
           <li>
-Let <var>hashData</var> be the result of running the
-<a href="#create-verify-hash-algorithm">Create Verify Hash Algorithm</a> with
-the <var>transformedDocument</var> and <var>options</var> parameters provided
+Let <var>hashData</var> be the result of
+<a href="#dfn-hashing">hashing</a> <var>transformedDocument</var>
```
```suggestion
<a href="#dfn-hashing">hashing</a> the <var>transformedDocument</var>
```

index.html
```diff
@@ -1677,13 +1678,13 @@ <h3>Generate Proof</h3>
...
-While the output of the
-<a href="#create-verify-hash-algorithm">Create Verify Hash Algorithm</a>
-can be a single value, implementers are advised that particular cryptographic
-suites might define `hashData` to be a comprised of multiple values that might
-be processed independently in the proof generation algorithm. For example, this
-approach is known to be taken in certain cryptographic suites that allow
-selective disclosure or unlinkability via the digital proof.
+While the output of the <a href="#dfn-hashing-algorithm">hashing algorithm</a>
+can be a single value, such as a 32 byte SHA2-256 value, implementers are
+advised that particular cryptographic suites might define `hashData` to be a
```
```suggestion
advised that some cryptographic suite(s) might define `hashData` to be
```

TallTed (changes requested):
A couple of words...

OR13 (approved):

TallTed (approved):

dlongley (approved):
Thanks! We don't need to hold this up based on the concern I raised, we'll figure out the right way to deal with it as we move forward.

mprorock (approved):

sbutterfield (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2022-10-19-vcwg#section2-2) on 2022-10-19

- no resolutions were taken

#### 2.2. Remove the unnecessary Create Verify Hash Algorithm. (pr vc-data-integrity#63)

_See github pull request [vc-data-integrity#63](https://github.com/w3c/vc-data-integrity/pull/63)._




**Manu Sporny:** still trying to figure out what the text might be for this.  
… we will remove the current wrong language.  
… key thing is that the way verify hash is created in BBS will be different to all other crypto suites.  
… its hard to predict what will happen with BBS so getting the language right now will be difficult.  

**Markus Sabadello:** might fit into RCH working group.  

**Manu Sporny:** we will remove the wrong text now. Worst case is that every crypto algorithm might have to use same text.  

</details>

msporny:
@dlongley wrote:
> is there a way we can keep this function so it can be reused by cryptosuites

I added an issue marker in 33e83ae2f639e64a1a7d1301ffaed6b9fac5a90a to track that we might define this later based on how cryptographic suite specification authoring goes.

msporny:

PR #64, opened 2022-10-17
Update algorithm for Verify Proof

msporny:
This PR updates the algorithm used to verify proofs to align with the language used to generate proofs (now merged into `main`).

At present, both the proof generation and proof verification algorithms presume only a single proof. The algorithms will need to be updated to support proof sets and proof chains in a future PR.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/64.html" title="Last updated on Oct 23, 2022, 8:50 PM UTC (520a357)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/64/47774a7...520a357.html" title="Last updated on Oct 23, 2022, 8:50 PM UTC (520a357)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1742,21 +1742,24 @@ the `proof` value removed.
 Let <var>transformedDocument</var> be the result of
 <a href="#dfn-transformation">transforming</a> the <var>unsecuredDocument</var>
 according to a <a>transformation algorithm</a> associated with the
-<var>cryptosuite</var> specified in <var>proof</var> and the <var>options</var>
-parameters provided as inputs to the algorithm.
+<a>cryptographic suite</a> specified in <var>proof</var> and the
+<var>options</var> parameters provided as inputs to the algorithm.
+The type of <a>cryptographic suite</a> is specified by the
+<var>proof.type</var> value and MAY be further described by
+<a>cryptographic suite</a>-specific properties expressed in <var>proof</var>.
         </li>
           <li>
 Let <var>hashData</var> be the result of
 <a href="#dfn-hashing">hashing</a> the <var>transformedDocument</var>
 according to a <a>hashing algorithm</a> associated with the
-<var>cryptosuite</var> specified in the <var>proof</var> and <var>options</var>
-parameters provided as inputs to the algorithm.
+<a>cryptographic suite</a> specified in the <var>proof</var> and
+<var>options</var> parameters provided as inputs to the algorithm.
           </li>
           <li>
 Let <var>isProofVerified</var> be the result of running the
-<a>proof verification algorithm</a> associated with the <var>cryptosuite</var>
-with the <var>hashData</var> and <var>options</var> parameters provided
-as inputs to the algorithm.
+<a>proof verification algorithm</a> associated with the
+<a>cryptographic suite</a> with the <var>hashData</var> and <var>options</var>
+parameters provided as inputs to the algorithm.
           </li>
           <li>
 If the <var>proof</var>.<var>proofPurpose</var> value does not match
--- a/index.html
+++ b/index.html
@@ -1739,16 +1739,16 @@ the `proof` value removed.
           </li>
           <li>
 Let <var>transformedDocument</var> be the result of
-<a href="#dfn-transformation">transforming</a> <var>unsecuredDocument</var>
+<a href="#dfn-transformation">transforming</a> the <var>unsecuredDocument</var>
 according to a <a>transformation algorithm</a> associated with the
 <var>cryptosuite</var> specified in <var>proof</var> and the <var>options</var>
 parameters provided as inputs to the algorithm.
         </li>
           <li>
 Let <var>hashData</var> be the result of
-<a href="#dfn-hashing">hashing</a> <var>transformedDocument</var>
+<a href="#dfn-hashing">hashing</a> the <var>transformedDocument</var>
 according to a <a>hashing algorithm</a> associated with the
-<var>cryptosuite</var> specified in <var>proof</var> and the <var>options</var>
+<var>cryptosuite</var> specified in the <var>proof</var> and <var>options</var>
 parameters provided as inputs to the algorithm.
           </li>
           <li>
@@ -1783,12 +1783,12 @@ Return <var>isProofValid</var> as the <a>verification result</a>.
 
         <p class="issue" title="Define how verification material is retrieved">
 Specify how <var>proof</var>.<var>verificationMethod</var> can be obtained,
-through some out-of-band process and passed in or it can be retrieved by
-derefencing its URL identifier, etc. It is expected that the <a>public key</a>
+e.g., through some out-of-band process, by
+dereferencing its URL identifier, etc. It is expected that the <a>public key</a>
 is retrieved by dereferencing its URL identifier in the <code>proof</code> node
 of the default graph of <var>securedDocument</var>. The algorithm needs to
-confirm that the <a>verification method</a> that describes cryptographic
-material, such as the <a>public key</a>, specifies its <a>controller</a> and
+confirm that the <a>verification method</a> that describes the cryptographic
+material, such as the <a>public key</a>, specifies its <a>controller</a>, and
 that its controllers's URL identifier can be dereferenced to reveal a
 bi-directional link back to the key. Ensure that the key's controller is a
 trusted entity before proceeding to the next step.
--- a/index.html
+++ b/index.html
@@ -1735,6 +1735,11 @@ is not set, a `MALFORMED_PROOF_ERROR`
 MUST be raised.
           </li>
           <li>
+If the <var>proof</var>.<var>proofPurpose</var> value does not match
+<var>options</var>.<var>expectedProofPurpose</var>, a
+`MISMATCHED_PROOF_PURPOSE_ERROR` MUST be raised.
+          </li>
+          <li>
 Let <var>unsecuredDocument</var> be a copy of <var>securedDocument</var> with
 the `proof` value removed.
           </li>
@@ -1760,11 +1765,6 @@ Let <var>isProofVerified</var> be the result of running the
 <a>proof verification algorithm</a> associated with the
 <a>cryptographic suite</a> with the <var>hashData</var> and <var>options</var>
 parameters provided as inputs to the algorithm.
-          </li>
-          <li>
-If the <var>proof</var>.<var>proofPurpose</var> value does not match
-<var>options</var>.<var>proofPurpose</var>, a `MISMATCHED_PROOF_PURPOSE_ERROR`
-MUST be raised.
           </li>
           <li>
 If the <var>proof</var>.<var>created</var> is set and it deviates more than
--- a/index.html
+++ b/index.html
@@ -1753,7 +1753,7 @@ according to a <a>hashing algorithm</a> associated with the
 parameters provided as inputs to the algorithm.
           </li>
           <li>
-Let <var>isProofValid</var> be the result of running the
+Let <var>isProofVerified</var> be the result of running the
 <a>proof verification algorithm</a> associated with the <var>cryptosuite</var>
 with the <var>hashData</var> and <var>options</var> parameters provided
 as inputs to the algorithm.
@@ -1778,7 +1778,7 @@ If <var>options</var>.<var>challenge</var> is set and it does not match
 raised.
           </li>
           <li>
-Return <var>isProofValid</var> as the <a>verification result</a>.
+Return <var>isProofVerified</var> as the <a>verification result</a>.
           </li>
         </ol>
 
--- a/index.html
+++ b/index.html
@@ -1729,8 +1729,8 @@ If the <var>proof</var>.<var>type</var>,
 `MALFORMED_PROOF_ERROR` MUST be raised.
           </li>
           <li>
-If the cryptographic suite requires the 
-<var>proof</var>.<var>created</var> value, and it 
+If the cryptographic suite requires the
+<var>proof</var>.<var>created</var> value, and it
 is not set, a `MALFORMED_PROOF_ERROR`
 MUST be raised.
           </li>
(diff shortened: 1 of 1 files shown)


iherman:
A minor comment, though not strictly to §4.2: do we have to say somewhere what "raise XYZ ERROR" means (e.g., in 2.1), or do we intend to leave it at that and leave it to the implementations?

iherman (commented):
There is an inconsistency in the value references. $4.2/(2) uses "dotted" term references like _proof.type_, _proof.verificationMethod_, etc.  However, §4.1/(5) uses references to the same terms without the "dotted" formalism.  (I did not check all items, there may be such discrepancies elsewhere, too.) Even §4.2 is not fully consistent; shouldn't (5) and (6) say _proof.cryptosuite_ rather than "the _cryptosuite_ specified in _proof_"?

Technically, it is not clear to me where the parameters like _options_ come from. Or is this the ISSUE right after the algorithm (that only refers to _verificationMethod_)?

TallTed:
I would reword your original comment slightly, from `algorithms only presume a single proof` to `algorithms presume only a single proof`.

index.html
```diff
@@ -1690,59 +1690,93 @@ <h3>Generate Proof</h3>
...
+<var>proof</var>.<var>created</var> value is not set, a `MALFORMED_PROOF_ERROR`
+MUST be raised.
+          </li>
+          <li>
+Let <var>unsecuredDocument</var> be a copy of <var>securedDocument</var> with
+the `proof` value removed.
+          </li>
+          <li>
+Let <var>transformedDocument</var> be the result of
+<a href="#dfn-transformation">transforming</a> <var>unsecuredDocument</var>
```
```suggestion
<a href="#dfn-transformation">transforming</a> the <var>unsecuredDocument</var>
```

index.html
```diff
@@ -1690,59 +1690,93 @@ <h3>Generate Proof</h3>
...
+according to a <a>transformation algorithm</a> associated with the
+<var>cryptosuite</var> specified in <var>proof</var> and the <var>options</var>
+parameters provided as inputs to the algorithm.
+        </li>
+          <li>
+Let <var>hashData</var> be the result of
+<a href="#dfn-hashing">hashing</a> <var>transformedDocument</var>
```
```suggestion
<a href="#dfn-hashing">hashing</a> the <var>transformedDocument</var>
```

index.html
```diff
@@ -1690,59 +1690,93 @@ <h3>Generate Proof</h3>
...
+according to a <a>hashing algorithm</a> associated with the
+<var>cryptosuite</var> specified in <var>proof</var> and the <var>options</var>
```
```suggestion
<var>cryptosuite</var> specified in the <var>proof</var> and <var>options</var>
```

index.html
```diff
@@ -1690,59 +1690,93 @@ <h3>Generate Proof</h3>
...
-and the <a>public key</a> to the <a>proof verification algorithm</a>.
-Return the resulting boolean value.
+Return <var>isProofValid</var> as the <a>verification result</a>.
           </li>
         </ol>
+
+        <p class="issue" title="Define how verification material is retrieved">
+Specify how <var>proof</var>.<var>verificationMethod</var> can be obtained,
+through some out-of-band process and passed in or it can be retrieved by
+derefencing its URL identifier, etc. It is expected that the <a>public key</a>
```
```suggestion
e.g., through some out-of-band process, by
dereferencing its URL identifier, etc. It is expected that the <a>public key</a>
```

index.html
```diff
@@ -1690,59 +1690,93 @@ <h3>Generate Proof</h3>
...
+is retrieved by dereferencing its URL identifier in the <code>proof</code> node
+of the default graph of <var>securedDocument</var>. The algorithm needs to
+confirm that the <a>verification method</a> that describes cryptographic
+material, such as the <a>public key</a>, specifies its <a>controller</a> and
```
```suggestion
confirm that the <a>verification method</a> that describes the cryptographic
material, such as the <a>public key</a>, specifies its <a>controller</a>, and
```

TallTed (changes requested):
Some minor rephrasing

OR13 (approved):

index.html
```diff
@@ -1690,59 +1690,93 @@ <h3>Generate Proof</h3>
...
+Let <var>proof</var> be set to <var>securedDocument</var>.<var>proof</var>.
+          </li>
+          <li>
+If the <var>proof</var>.<var>type</var>,
+<var>proof</var>.<var>verificationMethod</var>, or
+<var>proof</var>.<var>proofPurpose</var> values are not set, a
+`MALFORMED_PROOF_ERROR` MUST be raised.
+          </li>
+          <li>
+If the cryptographic suite does not support unlinkability and the
```
dlongley: Maybe we just want to say "if the cryptographic suite requires `proof.created`" rather than make it hinge on the unlinkability property.

index.html
```diff
@@ -1690,59 +1690,93 @@ <h3>Generate Proof</h3>
...
+<var>proof</var>.<var>created</var> value is not set, a `MALFORMED_PROOF_ERROR`
+MUST be raised.
```
dlongley: This error description is fine for now -- but we will want to see if we can improve / draw from other specs for better error patterns and error type / code reuse.

index.html
```diff
@@ -1690,59 +1690,93 @@ <h3>Generate Proof</h3>
...
+          <li>
+Let <var>unsecuredDocument</var> be a copy of <var>securedDocument</var> with
+the `proof` value removed.
+          </li>
+          <li>
+Let <var>transformedDocument</var> be the result of
+<a href="#dfn-transformation">transforming</a> <var>unsecuredDocument</var>
+according to a <a>transformation algorithm</a> associated with the
+<var>cryptosuite</var> specified in <var>proof</var> and the <var>options</var>
+parameters provided as inputs to the algorithm.
```
dlongley: Can we say how the cryptosuite is specified, i.e., via `type` (and perhaps other cryptosuite type-specific properties)?

```suggestion
parameters provided as inputs to the algorithm. Note: The <var>cryptosuite</var>
is specified by <var>proof.type</var> and MAY be further described by
cryptosuite-specific properties.
```

Seeing as `the <var>cryptosuite</var> specified in <var>proof</var>` is mentioned in other steps as well -- we might want the above language more generally expressed in an early step.

index.html
```diff
@@ -1690,59 +1690,93 @@ <h3>Generate Proof</h3>
...
+          <li>
+Let <var>hashData</var> be the result of
+<a href="#dfn-hashing">hashing</a> <var>transformedDocument</var>
+according to a <a>hashing algorithm</a> associated with the
+<var>cryptosuite</var> specified in <var>proof</var> and the <var>options</var>
+parameters provided as inputs to the algorithm.
           </li>
           <li>
-Let <var>document</var> be a copy of <var>signed document</var>.
+Let <var>isProofValid</var> be the result of running the
```
dlongley: The ever-raging war of verification vs. validation strikes again. I think "verified" is better than "valid" here because it's the result of the "proof verification algorithm" and because whether or not the proof is valid for a particular use case is a business decision.

```suggestion
Let <var>isProofVerified</var> be the result of running the
```

index.html
```diff
@@ -1690,59 +1690,93 @@ <h3>Generate Proof</h3>
...
+<a>proof verification algorithm</a> associated with the <var>cryptosuite</var>
+with the <var>hashData</var> and <var>options</var> parameters provided
+as inputs to the algorithm.
+          </li>
+          <li>
+If the <var>proof</var>.<var>proofPurpose</var> value does not match
+<var>options</var>.<var>proofPurpose</var>, a `MISMATCHED_PROOF_PURPOSE_ERROR`
+MUST be raised.
```
dlongley: This step should be run before any effort is made to verify the proof, i.e., the expected proof purpose should be checked first to find a matching proof. We might want the options name to be "expectedProofPurpose" as well, for clarity.

index.html
```diff
@@ -1690,59 +1690,93 @@ <h3>Generate Proof</h3>
...
-passing the information in <var>proof</var>.
+If <var>options</var>.<var>challenge</var> is set and it does not match
+<var>proof</var>.<var>challenge</var>, an `INVALID_CHALLENGE_ERROR` MUST be
+raised.
           </li>
           <li>
-Pass the <a>proofValue</a>, <var>tbv</var>,
-and the <a>public key</a> to the <a>proof verification algorithm</a>.
-Return the resulting boolean value.
+Return <var>isProofValid</var> as the <a>verification result</a>.
```
```suggestion
Return <var>isProofVerified</var> as the <a>verification result</a>.
```

dlongley: I think we need a cleaner separation of cryptosuite verification and purpose verification. We should:

1. Find a proof that matches the expected purpose; if none is found, raise an error.
2. Verify the proof using the cryptosuite. This returns `true` or `false` or throws if the operation cannot be executed.
3. If cryptosuite verification failed, return that the proof is not verified, otherwise...
4. Verify the proof against the expected purpose's specific checks (checking 'domain' and 'challenge' and so on are run here for authentication proofs, allowing for other purposes to be slotted in, i.e., "authentication" isn't the only known proof purpose).
5. Finally, return whether the proof is verified or not.

I'm writing this down so it can be turned into a separate issue to further iterate on the algorithm, it doesn't have to go into this PR if that's not desirable at this time.

dlongley (approved):
Thanks! Approving to keep things moving, but see suggestions (and +1 to @TallTed's suggestions).

mprorock (approved):

OR13 (approved):

sbutterfield (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2022-10-19-vcwg#section2-3) on 2022-10-19

- no resolutions were taken

#### 2.3. Update algorithm for Verify Proof (pr vc-data-integrity#64)

_See github pull request [vc-data-integrity#64](https://github.com/w3c/vc-data-integrity/pull/64)._




**Manu Sporny:** this is just updating the language.  

</details>

msporny:
> A minor comment, though not strictly to §4.2: do we have to say somewhere what "raise XYZ ERROR" means (e.g., in 2.1), or do we intend to leave it at that and leave it to the implementations?

Ultimately, it's up to the implementations to expose those errors. Exactly how each one does it is implementation specific at the moment. We might want to do something like what the [JSON-LD API](https://www.w3.org/TR/json-ld11-api/#error-handling) did, though some have complained about the use of text strings instead of error codes.

We do plan to expose those error types in protocols such as the VC API via a `code` field, but that's out of scope for this spec (probably). In short, we're still working things out. I'm trying to be consistent with the error codes so that when it comes time to expose them further up the software stack, people aren't caught off guard.

> There is an inconsistency in the value references. $4.2/(2) uses "dotted" term references like _proof.type_, _proof.verificationMethod_, etc. However, §4.1/(5) uses references to the same terms without the "dotted" formalism. (I did not check all items, there may be such discrepancies elsewhere, too.) Even §4.2 is not fully consistent; shouldn't (5) and (6) say _proof.cryptosuite_ rather than "the _cryptosuite_ specified in _proof_"?

Yes, thanks for spotting that, it is an oversight and I'll fix that in an upcoming commit. I was going back and forth on the easiest way to mark up these variables.

> Technically, it is not clear to me where the parameters like _options_ come from. Or is this the ISSUE right after the algorithm (that only refers to _verificationMethod_)?

`options` is specified as an input to the algorithm in the [opening paragraph](https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/64.html#verify-proof). I'm still trying to figure out if the ReSpec `var` parameter approach is more helpful than harmful... I think it's helping, but the github preview doesn't show the markup, so it's difficult to see it without checking out the source and rendering the document locally. The `var` parameter highlights everywhere the variable is used, so in theory, you should be able to click on the word `options` and it will highlight everywhere the variable is used in the algorithm. In short, if you had a properly rendered document, this might have been easier to find for you (or I screwed up the markup and I'll fix it :P ).

index.html
```suggestion
If the cryptographic suite requires the 
<var>proof</var>.<var>created</var> value, and it 
is not set, a `MALFORMED_PROOF_ERROR`
MUST be raised.
```

index.html
msporny: This presumes the `cryptosuite` property is in the `proof`, which it isn't for anything other than `DataIntegrityProof`... If it is, that'll have to be fixed. I'm also hesitating to clarify this in the body of the algorithm. I'll try to think about the best way to integrate this into this section and agree that we want to say something about cryptosuites being able to define other properties in a proof.

index.html
msporny: Ok, will look into rewriting the algorithm in this way.

index.html
msporny: Raised an issue here: https://github.com/w3c/vc-data-integrity/issues/66

index.html
msporny: I misunderstood what you were going for the first time I read your comments. I've applied your note and clarified the thing that threw me off the first time in f0732e50c6c450ff6265022981f5a47367343dcf.

index.html
msporny: Fixed in f3a3d2c92ce1688626b2a423bb6a83da0d0578e4.

msporny:
@iherman wrote:
> > There is an inconsistency in the value references. $4.2/(2) uses "dotted" term references like _proof.type_, _proof.verificationMethod_, etc. However, §4.1/(5) uses references to the same terms without the "dotted" formalism. (I did not check all items, there may be such discrepancies elsewhere, too.) Even §4.2 is not fully consistent; shouldn't (5) and (6) say _proof.cryptosuite_ rather than "the _cryptosuite_ specified in _proof_"?
> 
> Yes, thanks for spotting that, it is an oversight and I'll fix that in an upcoming commit. I was going back and forth on the easiest way to mark up these variables.

Fixed in 520a357a4454f9a2342ab7220ae43831a3a57535.

msporny:
Normative, multiple reviews, changes requested and made, no objections.

PR #69, opened 2022-11-10
Set up auto-publish.yml

iherman:
This is to run ECHIDNA. Publishing happens when merging into 'main'

(It is a copy-paste of the similar script in the VCDM repo, with the change on the short name.)

### DIFF
--- a/.github/auto-publish.yml
+++ b/.github/auto-publish.yml
@@ -1,0 +1,20 @@
+# .github/workflows/pr-push.yml
+name: CI
+on:
+  pull_request: {}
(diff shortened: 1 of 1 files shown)


msporny (approved):
LGTM. This PR just mechanizes the process that the WG made a [resolution](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2022-10-26-vcwg#resolution1) to put in place.

msporny:
Implementation of process agreed to by existing WG resolution, merging.

PR #79, opened 2023-01-17
Attempt to synchronize the security vocabulary with the FPWD spec

iherman:
This is a _**draft**_ PR to engage a discussion. Some of you may see it as a friendly provocation...

We have now an official draft for [VC Data Integrity](https://www.w3.org/TR/vc-data-integrity/). There is also a document for the [security vocabulary](https://w3c.github.io/vc-data-integrity/vocab/security/vocabulary.html) which is the reformatted version of the [CCG security vocabulary](https://w3c-ccg.github.io/security-vocab/) minus the examples. (At the time of creating the current vocabulary document the toolkit did not allow for the examples; the current toolkit has this feature). However, looking at that document it is fairly messy. It is not clear how the various terms (classes, properties, etc) relate to one another, the descriptions are inconsistent or simply missing, the relationship to the Data Integrity spec is unclear, etc. It looks like a notebook, a collection of ideas.

As FPWD for the integrity spec is out, I tried to clean up and, mainly, to sync the FPWD with the vocabulary definitions. Caveat (but maybe at this point it is an advantage): I am not a domain expert in security, so there are many things whose details I do not understand. Nevertheless, I immediately found a set problems that was easy to settle: some terms referred to in the specification were not present in the vocabulary in the first place. These are `Ed25519Signature2020`, `created`, `proofChain`, and `keyAgreement`. No big deal to add those. I could also took over some text from the spec to refresh the term descriptions.

But I got lost in trying to make sense of the vocabulary, so I took a very radical step. I started "fresh". I did the following:

- I took all the classes and properties that are defined or, at least, mentioned in the spec. That was actually relatively straightforward, except for one minor thing: I introduced a `VerificationMethod` class to make the vocabulary cleaner: it is the range of `verificationMethod` and the domain of `controller`. It is also the range of `authentication` and friends. (I was actually wondering whether the `VerificationMethod` as I used it is the same as the `Key` in the CCG vocabulary, but it did not sound right.)

    Also, just as in the VCDM, I've created a `ProofGraph` to differentiate an RDF Graph that holds a (single) `Proof`. This is necessary for properties whose values are, well, graphs (like `proof`).

- I then took all the other terms and _added them as deprecated_ (some had already been deprecated in the CCG, I added a temporary note to the label for those). My logic was: the security vocabulary is defined in the security spec, and anything that is not defined by the spec should not be in the vocabulary. On the other hand, we do not want to remove terms; hence move them into the 'deprecated' sections. (I could be convinced to keep terms that are defined more properly in, say, a note, some other spec, or a proper text. At the moment, there is no definition for most of those…)

The HTML result is [on this branch](https://raw.githack.com/w3c/vc-data-integrity/vocab/sync-with-fpwd/vocab/security/vocabulary.html), (with the yml, ttl, and json-ld alongside the html file). For my own understanding, I have also created and [ad-hoc diagram](https://docs.google.com/drawings/d/1NWi4I3eBo2_kheX9VwSFoBTw72czjo9kIKtJH0DlJls/edit) of the class/range/domain relationships; it may help in understanding the current status.

While doing all this, I've also found some other issues that need discussions (probably spawning into separate issues when we get there):

- There is a set of "Suites", which are all (directly or indirectly) subclasses of "Proof" per the CCG spec. I kept "Ed25519Signature2020" in the non-deprecated list because it is mentioned in the spec. But all these sound to me like identifiers for things that are defined elsewhere; in other words, this looks more like entries in a registry. Is it really the right place to hold these in the formal vocabulary and not having some registry somewhere?

    My ontology instinct also says that these should not be _Classes_ but _Individuals_ of some types. The problem is that the current VCs already use the idiom of `type:"Ed25519Signature2020"`, which means that the entity is of type of a specific class. I.e., if we change these to individuals, we may have a backward compatibility problem.

- There is a whole area in the vocabulary for signatures ("Signature", separation between XXXSignatureXXX and XXXVerificationXXX, etc), with `Signature` as as a subclass of `Proof`. There is no mention of that in the specification (hence they are all in the deprecated list). Do we plan to add those? Do we need that?

- An aside: I also realized that the definition of `authentication` and friends come, sometimed verbatim, from the DID spec (see, e.g., [VC Integrity §2.3.2.1](https://w3c.github.io/vc-data-integrity/#authentication) vs. [DID Core §5.3.1.](https://www.w3.org/TR/did-core/#authentication)). I understand why that is so, but I wonder whether we should really do that. Instead, the spec should somehow mention these and refer to the DID spec for the normative content. 

I am sure there will be some terms that we want to move back into the regular list of terms after a proper specification, which is fine; that is why I provoked you all. You tell me. But then there should be a proper definition for each of those in the specification…

I hope I could raise your interest :-)


dlongley:
Note: The `created` term maps to `http://purl.org/dc/terms/created`, so it's part of the dublin core vocab.

iherman:
> Note: The `created` term maps to `http://purl.org/dc/terms/created`, so it's part of the dublin core vocab.

Ah! So it should be removed from this document altogether, right?

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-01-18-vcwg#section2-5) on 2023-01-18

- no resolutions were taken

#### 2.5. Attempt to synchronize the security vocabulary with the FPWD spec (pr vc-data-integrity#79)

_See github pull request [vc-data-integrity#79](https://github.com/w3c/vc-data-integrity/pull/79)._




**Manu Sporny:** there are open PRs for data integrity, but I have not had chance to process them.  
… keep an eye on Data Integrity #79 ... this relates to long standing issues with the security vocabulary.  
… status list 2021 has new issues, but no PRs.  
… see the new issues related to ordering of statuses, and caching.  

**Kristina Yasuda:** are there any other updates?.  
… reminder, there is a conversation on the mailing list regarding vc-jws-2020.  
… see the comments on the list.  

</details>

vocab/security/template.html
```diff
@@ -108,16 +131,40 @@ <h2>Namespaces</h2>
...
+    </section>   
 
-    <section id="individual_definitions">
-      <h1>Definitions for individuals</h1>
-    </section>
+    <section id="deprecated_term_definitions">
+      <h1>Deprecated term definitions</h1>
+
+      <p class="annoy">All terms in this section are <em><strong>deprecated</strong></em>, and are only kept in this vocabulary for backward compatibility. 
+        <br>New applications should not use them.
```
TallTed: A double newline provides better readability than a single newline.
```suggestion
        <br><br>New applications should not use them.
```

TallTed:
Please confirm which document(s) are generated from which source(s), i.e., which `vocab/security/vocabulary.{html|yml|jsonld|ttl}` document(s) in https://github.com/w3c/vc-data-integrity/pull/79/files should be reviewed and edited for content (where changes will be automatically reflected in the other documents), and which document(s) should be ignored.

iherman:
> Please confirm which document(s) are generated from which source(s), i.e., which `vocab/security/vocabulary.{html|yml|jsonld|ttl}` document(s) in https://github.com/w3c/vc-data-integrity/pull/79/files should be reviewed and edited for content (where changes will be automatically reflected in the other documents), and which document(s) should be ignored.

The only document that should be reviewed is the yml file for the vocabulary items. If the introductory HTML text is to be changed, then the source for that is the template.html file. All the other files are generated.

(This ambiguity will disappear when the automatic generation for the vocabulary will be installed in this repo as well, just as we have done it for the VCDM vocabulary.)

msporny:
Hey @iherman, it's not clear which of these things you'd like to handle in this PR and which ones could be future issues, so I'm trying to do a full review here:

In general, we should pull this PR in sooner than later as it doesn't do anything horrible that can't be undone, so I'd like to address most of these issues in future PRs and just deal with the "philosophy of how we manage this vocabulary" here and make any changes related to those things in this PR (not sure if there are any, though). 

As for the rest of your questions: 

> I took a very radical step. I started "fresh". 

I think that's fine. This vocabulary has been maintained in an ad-hoc manner since 2012... and it was meant to be more of a registry than a formally defined vocabulary. When we needed a term, and we couldn't find a good one anywhere, we added it to this document.

> I took all the classes and properties that are defined or, at least, mentioned in the spec. That was actually relatively straightforward, except for one minor thing: I introduced a VerificationMethod class to make the vocabulary cleaner: it is the range of verificationMethod and the domain of controller. 

I think that's ok, though wonder if we expect verificationMethods to be on things other than controllers. For now, I don't think we do, so it's fine as is.

> It is also the range of authentication and friends. (I was actually wondering whether the VerificationMethod as I used it is the same as the Key in the CCG vocabulary, but it did not sound right.)

Verification Methods are different from Keys. Verification Methods can use/be associated with Keys. I'm wondering if we should update Key to be CryptographicKey to avoid confusion.

> Also, just as in the VCDM, I've created a ProofGraph to differentiate an RDF Graph that holds a (single) Proof. This is necessary for properties whose values are, well, graphs (like proof).

I think that's fine for now, nothing jumps out as that being a bad thing.

> I then took all the other terms and added them as deprecated (some had already been deprecated in the CCG, I added a temporary note to the label for those). My logic was: the security vocabulary is defined in the security spec, and anything that is not defined by the spec should not be in the vocabulary. On the other hand, we do not want to remove terms; hence move them into the 'deprecated' sections. (I could be convinced to keep terms that are defined more properly in, say, a note, some other spec, or a proper text. At the moment, there is no definition for most of those…)

I think that's fine for now... my concern here is that some people might assert that this vocabulary is only for Verifiable Credentials and so anything not related to VCs shouldn't go in there. That will, of course, break the ecosystem, so we should be very careful not to remove things unless we know for sure that no one is using them... and since this is a decentralized vocabulary, it's not easy to know that.

> For my own understanding, I have also created and [ad-hoc diagram](https://docs.google.com/drawings/d/1NWi4I3eBo2_kheX9VwSFoBTw72czjo9kIKtJH0DlJls/edit) of the class/range/domain relationships; it may help in understanding the current status.

The diagram is great! We should use it! :)

The only way I know of defining ProofGraph and CryptographicKey is to have a normative appendix (weird, I know) that defines these things for formal correctness... almost no one is going to care, except for the people that want to understand the formal model for the classes, properties, domains, ranges, and their relationships. I do think it's important to define those, but again, most will not care, so perhaps we put it in an appendix in the spec called "Formal Data Model" or something to that effect.

> There is a set of "Suites", which are all (directly or indirectly) subclasses of "Proof" per the CCG spec. I kept "Ed25519Signature2020" in the non-deprecated list because it is mentioned in the spec. But all these sound to me like identifiers for things that are defined elsewhere; in other words, this looks more like entries in a registry. Is it really the right place to hold these in the formal vocabulary and not having some registry somewhere?

Yes, well, the security vocabulary WAS the place that people would list these things. We don't necessarily need them here, but at the same time, they're useful. Perhaps we can use your "see also" feature in `yml2vocab` to point people to other vocabularies?

> My ontology instinct also says that these should not be Classes but Individuals of some types. The problem is that the current VCs already use the idiom of type:"Ed25519Signature2020", which means that the entity is of type of a specific class. I.e., if we change these to individuals, we may have a backward compatibility problem.

Also note that `DataIntegrityProof` changed the above... we're going to end up deprecating `Ed25519Signature2020` as well and use strings like `eddsa-2022` to identify cryptosuites from now on (so we don't end up in the situation we're in right now). That does raise a new question, which is, where do people register strings like `eddsa-2022`?

> There is a whole area in the vocabulary for signatures ("Signature", separation between XXXSignatureXXX and XXXVerificationXXX, etc), with Signature as as a subclass of Proof. There is no mention of that in the specification (hence they are all in the deprecated list). Do we plan to add those? Do we need that?

Yes, we need to keep them, probably as deprecated things. These are created/registered via the Cryptosuites, but again, most cryptosuites won't need new classes/properties after everything shifts to `DataIntegrityProof`.

> An aside: I also realized that the definition of authentication and friends come, sometimed verbatim, from the DID spec (see, e.g., [VC Integrity §2.3.2.1](https://w3c.github.io/vc-data-integrity/#authentication) vs. [DID Core §5.3.1.](https://www.w3.org/TR/did-core/#authentication)). I understand why that is so, but I wonder whether we should really do that. Instead, the spec should somehow mention these and refer to the DID spec for the normative content.

The problem is cross-referencing between DID WG and VC WG, and neither WG being in charge of the other's terminology. I had initially set up terms.html so that we could just have ONE glossary shared between DIDWG/VCWG/CCG, but then there was some gnashing of teeth on "who is the rightful controller of that vocabulary"... so, while we figure that out, we have at least two different `terms.html` files... and it's proliferating beyond that to one per spec now. ReSpec does allow us to pull in terminology from external files, but we can't combine terminology from `N`-many locations, IIRC. All that to say, we're going to have to continue to manage this stuff in `terms.html` until someone adds those features to ReSpec.

vocab/security/vocabulary.yml
```diff
@@ -8,49 +8,73 @@ prefix:
...
+        url: https://www.w3.org/TR/vc-data-integrity/#dataintegrityproof
+
+  - id: Ed25519Signature2020
+    label: ED2559 Signature Suite, 2020 version
+    upper_value: sec:Proof
+    comment: T.B.D.
+
+# These are the class definitions in the CCG documents that are not defined in the VCWG document; they are all deprecated
+
+  - id: Key
```
msporny, line 49:
```suggestion
  - id: CryptographicKey
```

vocab/security/vocabulary.yml
```diff
@@ -527,55 +619,55 @@ property:
...
   - id: invocationTarget
+    deprecated: true
     label: Invocation target
     comment: An invocation target identifies where a capability may be invoked, and identifies the target object for which the root capability expresses authority.
     see_also:
       - label: Authorization Capabilities
         url: https://w3c-ccg.github.io/zcap-spec/#root-capability
 
   - id: invoker
+    deprecated: true
```
msporny, line 670: Most of this isn't deprecated, as it's a part of the Authorization Capabilities work: https://w3c-ccg.github.io/zcap-spec/

msporny (changes requested):
I suggest this PR gets rid of the vocabulary.[html|ttl|jsonld] files, we can activate the new publishing mechanism now that it works in vc-data-model.

TallTed:
> [@iherman] I took a very radical step. I started "fresh".

Hoping it's relatively easy to do, and if it's not covered by — 

> I then took all the other terms and added them as deprecated

— I think it may be worth running a comparison between lists of terms defined in the previous vocab, and those defined in the "fresh" version — to help confirm that nothing important has been accidentally omitted.

TallTed:
>> [@iherman] For my own understanding, I have also created and [ad-hoc diagram](https://docs.google.com/drawings/d/1NWi4I3eBo2_kheX9VwSFoBTw72czjo9kIKtJH0DlJls/edit) of the class/range/domain relationships; it may help in understanding the current status.
> 
> [@msporny] The diagram is great! We should use it! :)

YES!  IMHO, a diagram like this would optimally be included with every vocab produced by any CG, WG, or otherwise.  It would help both the group, and eventual users of the group's output.

vocab/security/vocabulary.yml
```diff
@@ -215,51 +252,171 @@ class:
...
+  - id: proofValue
+    label: Proof value
+    domain: sec:Proof
+    range: xsd:string
+    comment: A string value that contains the data necessary to verify the digital proof using the `verificationMethod` specified
+
+  - id: proof
+    label: Proof sets
+    range: sec:ProofGraph
+    comment: The value of the `proof` property MUST identify <a href=”#ProofGraph”>`ProofGraph` instances</a> (informally, it indirectly identifies <a href=”#Proof”>`Proof` instances</a> each contained in a separate graph). The property is used to associate a proof with a graph of information. The proof property is typically not included in the canonicalized graphs that are then digested, and digitally signed. The order of the proofs is not relevant.
```
```suggestion
    comment: The value of the `proof` property MUST identify <a href=”#ProofGraph”>`ProofGraph` instances</a> (informally, it indirectly identifies <a href=”#Proof”>`Proof` instances</a>, each contained in a separate graph). The property is used to associate a proof with a graph of information. The proof property is typically not included in the canonicalized graphs that are then digested and digitally signed. The order of the proofs is not relevant.
```

vocab/security/vocabulary.yml
```diff
@@ -215,51 +252,171 @@ class:
...
+
+  - id: proofChain
+    label: Proof chain
+    range: rdf:List
+    comment: |
+      The value of the `proofChain` property MUST identify a sequence of <a href=”#ProofGraph”>`ProofGraph` instances</a> (informally, it indirectly identifies a series of <a href=”#Proof”>`Proof` instances</a>, each contained in a separate graph). The property is used to provide a sequence of multiple proofs where the order of when the proofs occurred matters. The proof chain property is typically not included in the canonicalized graphs that are then separately digested, and digitally signed.
```
```suggestion
      The value of the `proofChain` property MUST identify a sequence of <a href=”#ProofGraph”>`ProofGraph` instances</a> (informally, it indirectly identifies a series of <a href=”#Proof”>`Proof` instances</a>, each contained in a separate graph). The property is used to provide a sequence of multiple proofs where the order of when the proofs occurred matters. The proof chain property is typically not included in the canonicalized graphs that are then separately digested and digitally signed.
```

vocab/security/vocabulary.yml
```diff
@@ -215,51 +252,171 @@ class:
...
+
+  - id: controller
+    label: Controller
+    domain: sec:VerificationMethod
+    range: IRI
+    comment: |
+      A controller is an entity that claims control over a particular resource. Note that control is best validated as a two-way relationship where the controller claims control over a particular resource, and the resource clearly identifies its controller.
```
```suggestion
      A controller is an entity that claims control over a particular resource. Note that control is best validated as a two-way relationship, where the controller claims control over a particular resource, and the resource clearly identifies its controller.
```

vocab/security/vocabulary.yml
```diff
@@ -215,51 +252,171 @@ class:
...
+    range: VerificationMethod
+    comment: An `assertionMethod` property is used to specify a URL that contains information about a `verificationMethod` used for assertions.
+
+  - id: capabilityDelegation
+    label: Capability Delegation Method
+    range: VerificationMethod
+    comment: |
+      A `capabilityDelegation` property is used to express that one or more `verificationMethods` are authorized to verify cryptographic proofs that were created for the purpose of delegating capabilities.
+      A `verificationMethod` may be referenced by its identifier (a URL) or expressed in full.
+      The aforementioned proofs are created to prove that some entity is delegating the authority to take some action to another entity. A verifier of the proof should expect the proof to express a `proofPurpose` of `capabilityDelegation` and reference a `verificationMethod` to verify it. The dereferenced `verificationMethod` MUST have a controller property that has a property of `capabilityDelegation` that references the `verificationMethod`. This indicates that the controller has authorized it for the expressed `proofPurpose`.
```
TallTed, line 321: These are clearly meant to be distinct paragraphs, but that is not communicated by single line-breaks. I am not fluent in YML, and do not know how to embed blank lines in a value, which would *probably* be the easiest solution. (Such blank lines are often taller or shorter than `<p>` spacing in a given CSS, but they do make the paragraph break clear.)

iherman:
> 
> In general, we should pull this PR in sooner than later as it doesn't do anything horrible that can't be undone, so I'd like to address most of these issues in future PRs and just deal with the "philosophy of how we manage this vocabulary" here and make any changes related to those things in this PR (not sure if there are any, though).

I am fine with this. What we could do is

- agree on the proposed changes of @TallTed, so that they would be lost (domain experts should approve them
- merge the PR
- **_Install the automatic procedure for vocabulary generation, just as we have for VCDM_**
- take it from there...
- 
> 
> As for the rest of your questions:
> 

…

> 
> > I took all the classes and properties that are defined or, at least, mentioned in the spec. That was actually relatively straightforward, except for one minor thing: I introduced a VerificationMethod class to make the vocabulary cleaner: it is the range of verificationMethod and the domain of controller.
> 
> I think that's ok, though wonder if we expect verificationMethods to be on things other than controllers. For now, I don't think we do, so it's fine as is.

Let us do that later

> 
> > It is also the range of authentication and friends. (I was actually wondering whether the VerificationMethod as I used it is the same as the Key in the CCG vocabulary, but it did not sound right.)
> 
> Verification Methods are different from Keys. Verification Methods can use/be associated with Keys. I'm wondering if we should update Key to be CryptographicKey to avoid confusion.

I am not sure whether it makes a big difference, but I rely on domain experts for this...

> 
> > Also, just as in the VCDM, I've created a ProofGraph to differentiate an RDF Graph that holds a (single) Proof. This is necessary for properties whose values are, well, graphs (like proof).
> 
> I think that's fine for now, nothing jumps out as that being a bad thing.

It is the same as VCDM...

> 
> > I then took all the other terms and added them as deprecated (some had already been deprecated in the CCG, I added a temporary note to the label for those). My logic was: the security vocabulary is defined in the security spec, and anything that is not defined by the spec should not be in the vocabulary. On the other hand, we do not want to remove terms; hence move them into the 'deprecated' sections. (I could be convinced to keep terms that are defined more properly in, say, a note, some other spec, or a proper text. At the moment, there is no definition for most of those…)
> 
> I think that's fine for now... my concern here is that some people might assert that this vocabulary is only for Verifiable Credentials and so anything not related to VCs shouldn't go in there. That will, of course, break the ecosystem, so we should be very careful not to remove things unless we know for sure that no one is using them... and since this is a decentralized vocabulary, it's not easy to know that.
> 
> > For my own understanding, I have also created and [ad-hoc diagram](https://docs.google.com/drawings/d/1NWi4I3eBo2_kheX9VwSFoBTw72czjo9kIKtJH0DlJls/edit) of the class/range/domain relationships; it may help in understanding the current status.
> 
> The diagram is great! We should use it! :)

I am happy to keep it afloat, but these types of diagrams tend to become too complicated when the vocabulary becomes large and complex...

> 
> The only way I know of defining ProofGraph and CryptographicKey is to have a normative appendix (weird, I know) 

Normative appendices are fine.

> that defines these things for formal correctness... almost no one is going to care, except for the people that want to understand the formal model for the classes, properties, domains, ranges, and their relationships. I do think it's important to define those, but again, most will not care, so perhaps we put it in an appendix in the spec called "Formal Data Model" or something to that effect.
> 
> > There is a set of "Suites", which are all (directly or indirectly) subclasses of "Proof" per the CCG spec. I kept "Ed25519Signature2020" in the non-deprecated list because it is mentioned in the spec. But all these sound to me like identifiers for things that are defined elsewhere; in other words, this looks more like entries in a registry. Is it really the right place to hold these in the formal vocabulary and not having some registry somewhere?
> 
> Yes, well, the security vocabulary WAS the place that people would list these things. We don't necessarily need them here, but at the same time, they're useful. Perhaps we can use your "see also" feature in `yml2vocab` to point people to other vocabularies?

This vocabulary should cover the vocabularies that are defined in the spec and only them. Everything else should be somewhere else...

> 
> > My ontology instinct also says that these should not be Classes but Individuals of some types. The problem is that the current VCs already use the idiom of type:"Ed25519Signature2020", which means that the entity is of type of a specific class. I.e., if we change these to individuals, we may have a backward compatibility problem.
> 
> Also note that `DataIntegrityProof` changed the above... we're going to end up deprecating `Ed25519Signature2020` as well and use strings like `eddsa-2022` to identify cryptosuites from now on (so we don't end up in the situation we're in right now). That does raise a new question, which is, where do people register strings like `eddsa-2022`?

Yes, I was wondering about that. And I think it would indeed simplify things. 

The registry issue is for another day...

> 
> > An aside: I also realized that the definition of authentication and friends come, sometimed verbatim, from the DID spec (see, e.g., [VC Integrity §2.3.2.1](https://w3c.github.io/vc-data-integrity/#authentication) vs. [DID Core §5.3.1.](https://www.w3.org/TR/did-core/#authentication)). I understand why that is so, but I wonder whether we should really do that. Instead, the spec should somehow mention these and refer to the DID spec for the normative content.
> 
> The problem is cross-referencing between DID WG and VC WG, and neither WG being in charge of the other's terminology. I had initially set up terms.html so that we could just have ONE glossary shared between DIDWG/VCWG/CCG, but then there was some gnashing of teeth on "who is the rightful controller of that vocabulary"... so, while we figure that out, we have at least two different `terms.html` files... and it's proliferating beyond that to one per spec now. ReSpec does allow us to pull in terminology from external files, but we can't combine terminology from `N`-many locations, IIRC. All that to say, we're going to have to continue to manage this stuff in `terms.html` until someone adds those features to ReSpec.

But it is not only a matter of terms. It is a matter of full (normative!) descriptive texts!

My proposal would be, if we keep the vocabulary here, we refer to the existing DID Recommendation from this text as the normative description.

vocab/security/vocabulary.yml
iherman: If it is there then the vocabulary must be there in a separate document. This vocabulary is for terms defined in this specification and this specification only imho.

iherman:
> > [@iherman] I took a very radical step. I started "fresh".
> 
> Hoping it's relatively easy to do, and if it's not covered by —
> 
> > I then took all the other terms and added them as deprecated
> 
> — I think it may be worth running a comparison between lists of terms defined in the previous vocab, and those defined in the "fresh" version — to help confirm that nothing important has been accidentally omitted.

@TallTed, the original is here: https://w3c-ccg.github.io/security-vocab/. I would welcome if you (or someone) made a check, indeed. More pair of eyes is always welcome.

vocab/security/vocabulary.yml
```diff
@@ -215,51 +252,171 @@ class:
...
+
+  - id: capabilityInvocation
+    label: Capability Invocation Method
+    range: VerificationMethod
+    comment: |
+      A `capabilityInvocation` property is used to express that one or more `verificationMethods` are authorized to verify cryptographic proofs that were created for the purpose of invoking capabilities.
+      A `verificationMethod` MAY be referenced by its identifier (a URL) or expressed in full.
+      The aforementioned proofs are created to prove that some entity is attempting to exercise some authority they possess to take an action. A verifier of the proof should expect the proof to express a `proofPurpose` of `capabilityInvocation` and reference a `verificationMethod` to verify it. The dereferenced `verificationMethod` MUST have a controller property that, when dereferenced, has a property of `capabilityInvocation` that references the `verificationMethod.` This indicates that the controller has authorized it for the expressed `proofPurpose`.
```
TallTed, line 329: Again, clearly meant to be distinct paragraphs...

vocab/security/vocabulary.yml
```diff
@@ -215,51 +252,171 @@ class:
...
+
+  - id: publicKeyMultibase
+    label: Public key multibase
+    domain: sec:VerificationMethod
+    range: xsd:string
+    comment: |
+      The public key multibase property is used to specify the multibase-encoded version of a public key. The contents of the property are defined by specifications such as ED25519-2020 and listed in the Linked Data Cryptosuite Registry. Most public key type definitions are expected to:
+      • Specify only a single encoding base per public key type as it reduces implementation burden and increases the chances of reaching broad interoperability.
+      • Specify a multicodec header on the encoded public key to aid encoding and decoding applications in confirming that they are serializing and deserializing an expected public key type.
+      • Use compressed binary formats to ensure efficient key sizes.
```
TallTed, line 353: These bullets (`<ul>`) may need some special treatment, to make multiple lines indent properly.

vocab/security/vocabulary.yml
iherman: @TallTed I am not sure what you refer to, but I _think_ you have not looked at the generated HTML file of the vocabulary. The [preview of the section](https://raw.githack.com/w3c/vc-data-integrity/vocab/sync-with-fpwd/vocab/security/vocabulary.html#capabilityInvocation) clearly separates the distinct paragraphs imho...

Could you check your formatting comments against the generated HTML? Thx.

TallTed: Ah, you're right!  My bad.

vocab/security/vocabulary.yml
TallTed: *These* do need further attention, possibly to change the `•` to `*`, possibly to use whatever other markup makes the bullet-items "hang" properly.

TallTed:
> these types of diagrams tend to become too complicated when the vocabulary becomes large and complex

No matter how complicated the diagram may get, it will be easier to parse than the text alone — especially "when the vocabulary becomes large and complex"!

vocab/security/vocabulary.yml
iherman: Let me take this separately, because this is an issue on the tool and not the vocabulary. I have raised a separate issue: https://github.com/w3c/yml2vocab/issues/8 and will resolve this particular conversation. It should not be a show-stopper for this particular vocabulary.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-01-25-vcwg#section4-8) on 2023-01-25

- no resolutions were taken

#### 4.8. Attempt to synchronize the security vocabulary with the FPWD spec (pr vc-data-integrity#79)

_See github pull request [vc-data-integrity#79](https://github.com/w3c/vc-data-integrity/pull/79)._




**Manu Sporny:** 79 is taken sec vocab and refactored to match VC data model.  
… would be first revision, won't have all details right on first pass, but is a good first step.  
… does not belive anything provacative.  

</details>

vocab/security/vocabulary.yml
iherman: To allow progression, I will merge this (and similar) editorial changes as proposed by @TallTed, so that we could merge this PR in a subsequent step. We can always come back and wordsmith the definitions.

iherman:
@msporny,

out of the steps I proposed in https://github.com/w3c/vc-data-integrity/pull/79#issuecomment-1400601222


>>In general, we should pull this PR in sooner than later as it doesn't do anything horrible that can't be undone, so I'd like to address most of these issues in future PRs and just deal with the "philosophy of how we manage this vocabulary" here and make any changes related to those things in this PR (not sure if there are any, though).
> I am fine with this. What we could do is
>
> - agree on the proposed changes of @TallTed, so that they would be lost (domain experts should approve them
> - merge the PR
> - Install the automatic procedure for vocabulary generation, just as we have for VCDM
> - take it from there...

I have made step (1) above and turned off the 'Draft' flag on the PR. Per the [call yesterday](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-01-25-vcwg#section4-8), this PR could be merged, and then the automatic procedure installed. However, I will be out of circulation for a while for medical reasons, so I would prefer you pick up the next steps at this point...

msporny:
@iherman wrote:
> I will be out of circulation for a while for medical reasons, so I would prefer you pick up the next steps at this point.

Ok, I can pick this PR up from here, will do some rebasing to clean up the commit history, and will process it during the weekend. I hope your upcoming medical stuff goes well. :)

vocab/security/vocabulary.yml
dlongley: Not sure how we'll want to resolve this, but we do need to keep the terms around somehow and have a way to incubate new terms.

msporny (approved):

msporny:
Editorial, multiple reviews, changes requested and made, no objections, future work needed, merging.

msporny:
@iherman wrote:
> Install the automatic procedure for vocabulary generation, just as we have for VCDM

This is now done and working:

https://w3c.github.io/vc-data-integrity/vocab/security/vocabulary.html

The Echidna auto-publication works now as well:

https://github.com/w3c/vc-data-integrity/actions/runs/4033945598/jobs/6934785892

Proof of auto-publication is here:

https://www.w3.org/TR/2023/WD-vc-data-integrity-20230128/

1azteca1:
> * took all the classes and properties that are defined or, at least, mentioned in the spec. That was actually relatively straightforward, except for one minor thing: I introduced a `VerificationMethod` class to make the vocabulary cleaner: it is the range of `verificationMethod` and the domain of `controller`. It is also the range of `authentication` and friends. (I was actually wondering whether the `VerificationMethod` as I used it is the same as the `Key` in the CCG vocabulary, but it did not sound right.)

I kill for that kind of información máster. i need💡

PR #85, opened 2023-02-25
Replace `proofChain` property with `previousProof` property.

msporny:
This PR finalizes the mechanism for performing proof chaining to a simpler property-based mechanism in a proof (rather than requiring an additional change to the top-level data model that might conflict with the existence of the `proof` property).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/85.html" title="Last updated on Mar 4, 2023, 4:46 PM UTC (f46d696)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/85/b698759...f46d696.html" title="Last updated on Mar 4, 2023, 4:46 PM UTC (f46d696)">Diff</a>


index.html
```diff
@@ -487,6 +495,14 @@ <h3>Proofs</h3>
 <em>proof value</em> generated by the
 <a href="#generate-proof">Generate Proof Algorithm</a>.
           </dd>
+
+          <dt>previousProof</dt>
+          <dd>
+An optional string value that identifies another <a>data integrity proof</a>
+that MUST verify before the current proof is processed. This property is used by
```
```suggestion
that MUST verify before the current proof is processed. This property is used
```

index.html
```diff
@@ -422,6 +422,14 @@ <h3>Proofs</h3>
         </p>
 
         <dl style="margin-left: 1em;">
+          <dt>id</dt>
+          <dd>
+An optional identifier for the proof, which MUST be a URL [[URL]], such as
+a UUID (`urn:uuid:6a1676b8-b51f-11ed-937b-d76685a20ff5`).
```
```suggestion
a UUID as a URN (`urn:uuid:6a1676b8-b51f-11ed-937b-d76685a20ff5`).
```

Could reference this too: https://www.rfc-editor.org/rfc/rfc4122

index.html
```diff
@@ -695,9 +711,10 @@ <h3>Proof Chains</h3>
 A proof chain is useful when the same data needs to be signed by
 multiple entities and the order of when the proofs occurred matters,
 such as in the case of a notary counter-signing a proof that had been
-created on a document. A proof chain, where order needs to be preserved, is
-represented by associating an ordered list of proofs with the
-`proofChain` key in a document.
+created on a document. A proof chain, where proof order needs to be preserved,
+is expressed by providing at least one proof with an `id`, such as a UUID,
```
```suggestion
is expressed by providing at least one proof with an `id`, such as a UUID as a URN,
```

dlongley (approved):
Thanks! I think this is a great start.

index.html
```diff
@@ -706,7 +723,8 @@ <h3>Proof Chains</h3>
     "https://w3id.org/security/data-integrity/v1"
 ],
   "title": "Hello world!",
-  "proofChain": [{
+  "proof": [{
+    "id": "urn:uuid:60102d04-b51e-11ed-acfe-2fcd717666a7",
```
OR13, line 727: OMG! proofs that are not blank nodes! blasphemy!

index.html
```diff
@@ -723,7 +741,8 @@ <h3>Proof Chains</h3>
       DpuwRBokeWcmrgHxUXfnncxiZP",
     "proofPurpose": "assertionMethod",
     "proofValue": "z6Qnzr5CG9876zNht8BpStWi8H2Mi7XCY3inbLrZrm955QLBrp19KiWXerb8
-      ByPnAZ9wujVFN8PDsxxXeMoyvDqhZ"
+      ByPnAZ9wujVFN8PDsxxXeMoyvDqhZ",
+    "previousProof": "urn:uuid:60102d04-b51e-11ed-acfe-2fcd717666a7"
```
OR13, line 745: very clever.

OR13 (approved):
Makes me with `proof.id` was always present.

index.html
msporny: Time will tell. :)

msporny:
@OR13 wrote?
> Makes me with `proof.id` was always present.

Trying to understand this... with -> wish?

OR13:
Yes, "wish"... Blank nodes + `@container` in proof has some effects that I suspect most people are not expecting unless they are RDF / JSON-LD experts.

vocab/security/vocabulary.yml
```diff
@@ -272,6 +272,12 @@ property:
     range: xsd:string
     comment: The challenge property is used to associate a challenge with a proof, for use with a `proofPurpose` such as `authentication`. This string value SHOULD be included in a proof if a `domain` is specified.
 
+  - id: previousProof
+    label: Previous proof
+    domain: sec:Proof
+    range: xsd:anyURI
```
iherman:
```suggestion
    range: sec:Proof
```

My understanding, per description, is that we refer to another proof that must be checked before the current one. Ie, the object of the property must be a proof as well, right?

The minor downside is that this change makes it possible to create "embedded" proofs, ie, that the previous proof is in a blank node. I do not think it is actually a real problem but if it is, this can be explicitly restricted in the description within the specification.

index.html
```diff
@@ -487,6 +495,14 @@ <h3>Proofs</h3>
 <em>proof value</em> generated by the
 <a href="#generate-proof">Generate Proof Algorithm</a>.
           </dd>
+
+          <dt>previousProof</dt>
+          <dd>
+An optional string value that identifies another <a>data integrity proof</a>
+that MUST verify before the current proof is processed. This property is used
```
TallTed, line 502: The order of verification seems backward. I would think an enclosing proof should be verified before an enclosed (previous) proof. Likewise for enclosing encryption before enclosed (previous) encryption... I may be conflating encryption and signing/proving, which I think means we need to clarify how these will/do interoperate.
```suggestion
that MUST verify before the current proof is processed. This property is used
```

index.html
```diff
@@ -695,9 +711,10 @@ <h3>Proof Chains</h3>
 A proof chain is useful when the same data needs to be signed by
 multiple entities and the order of when the proofs occurred matters,
```
TallTed, line 712: Ties into the above — outer/later signing must be checked before inner/earlier signing, and the "required order" will be the reverse of outermost-to-innermost order.

vocab/security/vocabulary.yml
```diff
@@ -272,6 +272,12 @@ property:
...
+    comment: The` previousProof` property is used to identify a proof that MUST be verified before the proof that contains this property.
```
TallTed: Again, this seems backwards or inverted.
```suggestion
    comment: The `previousProof` property is used to identify a proof that MUST be verified before the proof that contains this property.
```

TallTed (changes requested):
As noted...

vocab/security/vocabulary.yml
mprorock: > My understanding, per description, is that we refer to another proof that must be checked before the current one. Ie, the object of the property must be a proof as well, right?
> 
> The minor downside is that this change makes it possible to create "embedded" proofs, ie, that the previous proof is in a blank node. I do not think it is actually a real problem but if it is, this can be explicitly restricted in the description within the specification.

i am good with this adjustment as noted by @iherman - marking approved for now, as I think we can open an issue and address, even if we take this PR as is (since it is a clear improvement)

mprorock (approved):

brentzundel (approved):

index.html
OR13: @brentzundel now is your chance to object to the use of `id` ; )

index.html
msporny: > I may be conflating encryption and signing/proving

Yes, this purely has to do with signing and not encryption.

Fundamentally, with chained signatures, you need to verify the entire chain so from that perspective, it doesn't really matter where you start.

One argument for where one should prefer to start is to try and determine where an attacker would likely strike if they were trying to DoS you. Presuming you need to fetch some of these keys from the network (like if people are using `did:web` or any ledger-based DID Method), an attacker could attack you by providing a bunch of bogus keys in a chain of proofs. So, of that chain, where are they most likely to attack? 

Would an attacker choose to provide a chain where the attack happens at the beginning or the end of a chain? That might suggest where we should start verification.

Fundamentally, if the first proof in a chain is invalid, the rest of the chain is invalid. If the last proof in a chain is invalid, that might be ok from a verifier perspective... remember, with the `previousProof` property... you can have directed acyclic graphs of signatures, where only a subset might need to verify for the Verifier to have confidence in the signed information. Now, I know that this PR does not talk about any of that... and those use cases are experimental (and we probably won't try to standardize them for years)... but when we get there, it hints at first-to-last verification order having a slight (theoretical) advantage over last-to-first. Maybe what we want is good caching behaviour, where we cache as much as we can early on, and bail as early on in the chain as we can?

So, I think the current argument is: If we do first to last, or last to first verification doesn't matter... and we'll put language in there that says that as long as the result of an alternate algorithm is the same as the one documented in the spec, it's conformant (which is typical language to put in an algorithms section).

The future argument might be: We optimize for caching the signature check of the head of the chain, because we might have multi-branching in these proof chains (effectively, proof trees) and so we should ensure that the root of the proof tree is valid before continuing down the branches. 

In any case, even if we were to pick the wrong thing here, I don't think it prevents us from changing it in the future in a way that doesn't cause backwards-incompatible changes.

... and to assuage people's concerns, we might want to start out with very strong "every proof in the chain MUST verify" language and then weaken it later once people provide good examples for why it's ok for some proofs in the chain to not verify.

index.html
msporny: See response here: https://github.com/w3c/vc-data-integrity/pull/85/files#r1125499681

TL;DR: Which order we go in doesn't matter at this point in time, and we can change it in the future if it does matter w/o creating backwards-incompatible changes.

vocab/security/vocabulary.yml
msporny: > My understanding, per description, is that we refer to another proof that must be checked before the current one. Ie, the object of the property must be a proof as well, right?

Yes, and I was trying to keep things simple so that we always deal w/ references and a flat proof array, and not a mixture of embedded proofs and references to proofs.

We can say `sec:Proof` and restrict the property in the spec to a URL (which I think it already is).

msporny:

PR #86, opened 2023-02-26
Add algorithm for retrieving a Verification Method.

msporny:
This PR adds the algorithm for retrieving a verification method, such as a cryptographic public key, from a controller document, such as a DID Document, given a verification method URL.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/86.html" title="Last updated on Mar 17, 2023, 5:56 PM UTC (a46507a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/86/7cdb5e2...a46507a.html" title="Last updated on Mar 17, 2023, 5:56 PM UTC (a46507a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1854,14 +1854,24 @@ If <var>vmIdentifier</var> is not a valid URL, an
 `INVALID_VERIFICATION_METHOD_URL` error MUST be raised.
           </li>
           <li>
+Let <var>controllerDocumentUrl</var> be the result of parsing
+<var>vmIdentifier</var> according to the rules of the URL scheme and extracting
+the primary resource identifier (without the fragment identifier).
+          </li>
+          <li>
 Let <var>vmFragment</var> be the result of parsing <var>vmIdentifier</var>
 according to the rules of the URL scheme and extracting the secondary
-resource identifier.
+resource identifier (the fragment identifier).
           </li>
           <li>
 Let <var>controllerDocument</var> be the result of dereferencing
-<var>vmIdentifier</var>, without <var>vmFragment</var>, according to the rules
+<var>controllerDocumentUrl</var>, according to the rules
 of the URL scheme and using the supplied <var>options</var>.
+          </li>
+          <li>
+If <var>controllerDocument</var>.<var>id</var> does not match the
+<var>controllerDocumentUrl</var>, a `INVALID_CONTROLLER_DOCUMENT_ID` error MUST
+be raised.
           </li>
           <li>
 If <var>controllerDocument</var> is not a valid <a>controller document</a>,
--- a/index.html
+++ b/index.html
@@ -1832,8 +1832,80 @@ trusted entity before proceeding to the next step.
       <section class="normative">
         <h3>Retrieve Verification Method</h3>
 
-      </section>
+        <p>
+The following algorithm specifies how to safely retrieve a verification method,
+such as a cryptographic public key, by using a <a>verification method</a>
+identifier contained in a <a>data integrity proof</a>. Required inputs are a
+<a>data integrity proof</a> (<var>proof</var>) and a set of
+<strong>dereferencing options</strong> (<var>options</var>). A
+<strong>verification method</strong> is produced as output.
+        </p>
 
+        <ol class="algorithm">
+          <li>
+Let <var>verificationMethod</var> be set to a `null` value.
+          </li>
+          <li>
+Let <var>vmIdentifier</var> be set to
+<var>proof</var>.<var>verificationMethod</var>.
+          </li>
+          <li>
+Let <var>vmPurpose</var> be set to <var>proof</var>.<var>proofPurpose</var>.
+          </li>
+          <li>
+If <var>vmIdentifier</var> is not a valid URL, an
+`INVALID_VERIFICATION_METHOD_URL` error MUST be raised.
+          </li>
+          <li>
+Let <var>controllerDocument</var> be the result of dereferencing
+<var>vmIdentifier</var>.
+          </li>
+          <li>
+If <var>controllerDocument</var> is not a valid <a>controller document</a>,
+an `INVALID_CONTROLLER_DOCUMENT` error MUST be raised.
+          </li>
+          <li>
+Let <var>verificationMethodArray</var> be an empty array. If the value of
+the <var>vmPurpose</var> property in the <var>controllerDocument</var> is a
+single value, add it to the  <var>verificationMethodArray</var>. If the value of
+the <var>vmPurpose</var> property in the <var>controllerDocument</var> is an
+array, add each item in the array to <var>verificationMethodArray</var>.
+          </li>
+          <li>
+For each <var>item</var> in <var>verificationMethodArray</var>:
+            <ol class="algorithm">
+              <li>
+If <var>item</var> is a string, replace <var>item</var> with the corresponding
+<var>expandedItem</var> in the
+<var>controllerDocument</var>.<var>verificationMethod</var> array. To determine
+if an <var>item</var> corresponds to an <var>expandedItem</var>, check for
+string equivalence between <var>item</var> and
+<var>expandedItem</var>.<var>id</var>. If a corresponding
+<var>expandedItem</var> does not exist, an `INVALID_VERIFICATION_METHOD`
+error MUST be raised.
+              </li>
+              <li>
+If the <var>item</var>.<var>id</var> value is equivalent to
+<var>vmIdentifier</var>, and <var>item</var>.<var>controller</var> is
+equivalent to <var>controllerDocument</var>.<var>id</var>, then set
+<var>verificationMethod</var> to <var>item</var>.
+              </li>
+            </ol>
+          </li>
+          <li>
+If <var>verificationMethod</var> is set to `null`, a
+`VERIFICATION_METHOD_NOT_FOUND` error MUST be raised.
+          </li>
+          <li>
+If <var>verificationMethod</var> is not a conforming <a>verification method</a>,
+a `INVALID_VERIFICATION_METHOD` error MUST be raised.
+          </li>
+          <li>
+Return <var>verificationMethod</var> as the
+<strong>verification method</strong>.
+          </li>
+        </ol>
+      </section>
     </section>
 
     <section>
--- a/index.html
+++ b/index.html
@@ -1843,9 +1843,6 @@ identifier contained in a <a>data integrity proof</a>. Required inputs are a
 
         <ol class="algorithm">
           <li>
-Let <var>verificationMethod</var> be set to a `null` value.
-          </li>
-          <li>
 Let <var>vmIdentifier</var> be set to
 <var>proof</var>.<var>verificationMethod</var>.
           </li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1812,8 +1812,80 @@ <h3>Verify Proof</h3>
...
+Let <var>controllerDocument</var> be the result of dereferencing
+<var>vmIdentifier</var>.
+          </li>
+          <li>
+If <var>controllerDocument</var> is not a valid <a>controller document</a>,
+an `INVALID_CONTROLLER_DOCUMENT` error MUST be raised.
+          </li>
+          <li>
+Let <var>verificationMethodArray</var> be an empty array. If the value of
+the <var>vmPurpose</var> property in the <var>controllerDocument</var> is a
```
dlongley: I think we will also need to talk about the acceptable contexts in the `controllerDocument` (or even a default context if no `@context` is present). We can turn this comment into an issue to handle it in another PR if we don't want to do it here.

dlongley (approved):
Thanks! Another good start to this. I left a comment about dealing with possible contexts in controller docs that we might want to turn into an issue for a later PR.

index.html
OR13: Yes, and an example would probably make this clearer.

OR13 (changes requested):
I am super supportive of this PR.

I want to see examples added.

I want to see the "context" issue addressed, can context for VM be omitted?

msporny:
@OR13 wrote:
> I want to see examples added.

Can you be more specific? A list of examples with what each example should show will help me create them.

> I want to see the "context" issue addressed, can context for VM be omitted?

Yes, I'm trying to get to a place where you don't need to use `@context` for controller documents and VMs (we might be able to get away with static type/field matching). I feel like it's going to take multiple iterations and discussion to get it right and explore the attack surface if we make `@context` optional for VMs and controller documents (for the core algorithms). Could we raise this as an issue and move the discussion there?

OR13:
I'm fine noting that context does not need to be present, if that indeed is the case.

The examples would make the context issue clearer as well.

Regarding examples, mostly want to be clear on the response shape for `controller` vs `verificationMethod`... Ideally something like:

https://example.com/controllers/123#verification-method-456

To go along with the algorithm description.

TallTed:
I request and strongly advise that we avoid overloaded abbreviations like VM (and PR), even if it seems like their meaning should be clear in context. For instance, years of reading "VM" as "Virtual Machine" overpower months of "Verification Method" [among others](https://acronyms.thefreedictionary.com/VM). "PR" is extra challenging with "Proposed Recommenation" and "Pull Request" among [hundreds of others](https://acronyms.thefreedictionary.com/PR).

index.html
```diff
@@ -1812,8 +1812,80 @@ <h3>Verify Proof</h3>
       <section class="normative">
         <h3>Retrieve Verification Method</h3>
 
-      </section>
+        <p>
+The following algorithm specifies how to safely retrieve a verification method,
```
TallTed, line 1836:
```suggestion
The following algorithm specifies how to safely retrieve a <a>verification method</a>,
```

index.html
```diff
@@ -1812,8 +1812,80 @@ <h3>Verify Proof</h3>
...
+<var>vmIdentifier</var>.
+          </li>
+          <li>
+If <var>controllerDocument</var> is not a valid <a>controller document</a>,
+an `INVALID_CONTROLLER_DOCUMENT` error MUST be raised.
+          </li>
+          <li>
+Let <var>verificationMethodArray</var> be an empty array. If the value of
+the <var>vmPurpose</var> property in the <var>controllerDocument</var> is a
+single value, add it to the  <var>verificationMethodArray</var>. If the value of
```
```suggestion
single value, add it to the <var>verificationMethodArray</var>. If the value of
```

TallTed (changes requested):
minor...

peacekeeper:
I find this algorithm strange.. It seems to retrieve the entire controller document and then search for one or more verification methods by purpose (e.g. `assertionMethod`). Maybe there are scenarios where that is useful, but in general I would instead expect this algorithm to retrieve a specific verification method directly by URL. The algorithm should still verify that the purpose matches, but I don't think it should use the purpose as the primary key for finding the verification method.

The algorithm starts with:

> Let _vmIdentifier_ be set to _proof.verificationMethod_.

So for example in the case of DIDs, _vmIdentifier_ could be `did:ex:123#key1`

And then it says:

> Let _controllerDocument_ be the result of dereferencing _vmIdentifier_.

I think this is a mistake, since the result of dereferencing _vmIdentifier_ should already be a single verification method, rather than a controller document.

OR13:
@peacekeeper yes, I agree with your comment... I think examples will make it clearer.

`resolve` of `did:example:123#key-456` produces a `DID Document`.
`dereference` of `did:example:123#key-456` produces a `VerificationMethod`

msporny:
> I find this algorithm strange

Keep in mind that any approach that results in an equivalent outcome is seen as valid in most specs. I need to add that language to the conformance section. 

That said, I agree that the language in the algorithm needs to improve. 

I'm struggling with which spec to refer to wrt. referencing. 

* If we refer to the URI spec, it's not specific enough wrt. dereferencing.
* If we refer to the Fetch spec, while it's how implementations actually work, I expect some to groan about it being too browser centric.
* We can't refer to the DID Resolution spec because non-DIDs can be used.

> I would instead expect this algorithm to retrieve a specific verification method directly by URL.

Remember that in HTTP, you retrieve a document and all fragment identifiers are dereferenced locally. That is, the server MUST completely ignore the hash (if it's sent) and often, the fragment identifier isn't sent up to the server. 

Another consideration is that directly retrieving the verification method by URL bypasses an important security check (to ensure that the proof purpose and controller are checked). I know that you mentioned some of this above, and I'm highlighting this for others in this thread that might not be aware that there are mandatory security checks that must be performed to avoid key misuse attacks, controller misuse attacks, and indirection attacks.

I'll try to rework the PR with the input above... what would help me in doing that is understanding which specification I should be referring to when speaking about "resolving" and "dereferencing". Remember, the solution has to work with all schemes that support both concepts... DIDs and HTTPS are just two of the schemes that we have to consider.

peacekeeper:
I think the algorithm should basically be like this:
1. Dereference the vmIdentifier minus its fragment to the controller document (the "primary resource"), according to the rules of the URI scheme of vmIdentifier.
2. Dereference the fragment to the verification method in the controller document (the "secondary resource"), according to the rules of the media type of the controller document.
3. Verify that the proof purpose and controller match what's expected.

This specification shouldn't "override" how steps 1. and 2. above work, i.e. the dereferencing part is mandated by URI schemes and media types.

peacekeeper:
This raises an interesting question.. Do we require that vmIdentifier always has a fragment? Or could it e.g. be an HTTP URL without fragment that dereferences directly to a verification method, rather than a controller document? In that case, how would proof purpose and controller be verified?

OR13:
@peacekeeper in general "sub resources" such as "VerificationMethod"s are represented with "fragments / relative identifiers"...

I have see no interoperability for dereferencing an external VM with an id like:

`did:example:123/verification-methods/456?query=latest#kid`... but I suppose that could be a legal verification method id.

msporny:
@peacekeeper wrote:
> I think the algorithm should basically be like this:

Ok, I'll rewrite the algorithm in that form, I do agree that it makes more sense to do it that way.

msporny:
> This raises an interesting question.. Do we require that vmIdentifier always has a fragment? 

We might want to start out requiring that, because...

> Or could it e.g. be an HTTP URL without fragment that dereferences directly to a verification method, rather than a controller document? In that case, how would proof purpose and controller be verified?

We could do this, but then the question becomes, when you go from the verification method (HTTP_DOCUMENT_A) to the controller document (HTTP_DOCUMENT_B), how do you ensure that the controller document lists the verification method as one it's in control of? The answer is, well, it has to be listed in the controller document somewhere by reference... and if you require that, then what benefit are you getting by putting the verification method in one place and the reference from the controller document to the verification method in another place?

I'm trying to think of a use case where you'd want one security domain managing a key, and another security domain managing the controller document. Like, you have INDUSTRY_BODY managing a controller document, but it references verification methods for a few INDUSTRY_MEMBERs, so the INDUSTRY_MEMBERs can update/manage the cryptographic material? Feels like a recipe for confusion/disaster... do we want to support that? Are there any legitimate/popular use cases there?

IOW, we could do that... and it would complicate the retrieval algorithm... and it's not clear that it's worth doing that. So, I think for now, let's just keep it simple: verification method identifiers are of the form URL#FRAGMENT, and the complete verification method MUST be expressed in the controller document.

I'll update the algorithm to reflect the above (and look for objections to the approach in the next couple of days).

msporny:
@peacekeeper I have implemented your request in 2cb981d65ead46dc3c9775b7c726bb4a2eaa4cbd

Your update simplified the algorithm by shifting the dereferencing mechanisms to the URL schemes. 

@OR13 I have implemented your request in 764e27e5c260e205f4a61da12e8c9d604cbafb05. I tried to keep the example minimal and avoid the `publicKeyJwk/publicKeyMultibase` permathread by just saying `// public cryptographic material goes here`.

At present, `@context` is not required in both controller documents and verification methods if JSON-LD processing is not being done. Otherwise, one must use a default context (or equivalent context to the default context plus other term declarations, if preferred). That language isn't in the spec yet, and I raised issue #88 to track that concern.

index.html
msporny: Implemented in 764e27e5c260e205f4a61da12e8c9d604cbafb05.

index.html
```diff
@@ -1831,8 +1831,92 @@ <h3>Verify Proof</h3>
...
+be raised.
+          </li>
+          <li>
+If <var>controllerDocument</var> is not a valid <a>controller document</a>,
+an `INVALID_CONTROLLER_DOCUMENT` error MUST be raised.
+          </li>
+          <li>
+Let <var>verificationMethod</var> be the result of dereferencing the
+<var>vmFragment</var> from the <var>controllerDocument</var> according to the
+rules of the associated URL scheme and using the supplied <var>options</var>.
```
peacekeeper:
```suggestion
rules of the media type of the <var>controllerDocument</var>.
```

I think this is important.. In the DID WG we discussed several times that dereferencing the fragment depends only on the media type and can't be re-defined by the URL scheme.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-03-08-vcwg#section2-4) on 2023-03-08

- no resolutions were taken

#### 2.4. Add algorithm for retrieving a Verification Method. (pr vc-data-integrity#86)

_See github pull request [vc-data-integrity#86](https://github.com/w3c/vc-data-integrity/pull/86)._




**Manu Sporny:** so, please take a look at it, comment, approve, etc..  

**Orie Steele:** just to comment on JCS canon scheme - we discussed it at the f2f, and one of the reason we withdrew the JWS work item, is we knew that JCS was going to be proposed.  
… I'm pretty strongly opposed to calling that suite 'JSON' when it requires the JCS dependency.  
… lots of devs use JSON but don't know about JCS..  
… I'm vocal on the PR regarding the namechange.  

**Brent Zundel:** we'll discuss.  
… but I'm very supporting of /using/ JCS.  


</details>

index.html
```diff
@@ -1831,8 +1831,92 @@ <h3>Verify Proof</h3>
       <section class="normative">
         <h3>Retrieve Verification Method</h3>
 
-      </section>
+        <p>
+The following algorithm specifies how to safely retrieve a verification method,
+such as a cryptographic public key, by using a <a>verification method</a>
+identifier contained in a <a>data integrity proof</a>. Required inputs are a
+<a>data integrity proof</a> (<var>proof</var>) and a set of
```
```suggestion
<strong><a>data integrity proof</a></strong> (<var>proof</var>) and a set of
```

index.html
```diff
@@ -1831,8 +1831,92 @@ <h3>Verify Proof</h3>
...
+resource identifier (the fragment identifier).
+          </li>
+          <li>
+Let <var>controllerDocument</var> be the result of dereferencing
+<var>controllerDocumentUrl</var>, according to the rules
+of the URL scheme and using the supplied <var>options</var>.
+          </li>
+          <li>
+If <var>controllerDocument</var>.<var>id</var> does not match the
+<var>controllerDocumentUrl</var>, a `INVALID_CONTROLLER_DOCUMENT_ID` error MUST
```
```suggestion
<var>controllerDocumentUrl</var>, an `INVALID_CONTROLLER_DOCUMENT_ID` error MUST
```

index.html
```diff
@@ -1831,8 +1831,92 @@ <h3>Verify Proof</h3>
...
+<var>vmFragment</var> from the <var>controllerDocument</var> according to the
+rules of the associated URL scheme and using the supplied <var>options</var>.
+          </li>
+          <li>
+If <var>verificationMethod</var> is not a valid <a>verification method</a>,
+an `INVALID_VERIFICATION_METHOD` error MUST be raised.
+          </li>
+          <li>
+If <var>verificationMethod</var> is not associated with the array of
+<var>vmPurpose</var>s in the <var>controllerDocument</var>, either by reference
```
TallTed: Something else might also need to be changed for ReSpec to properly handle this plural.
```suggestion
<var>vmPurposes</var> in the <var>controllerDocument</var>, either by reference
```

index.html
```diff
@@ -1831,8 +1831,92 @@ <h3>Verify Proof</h3>
...
+  "id": "https://controller.example/123",
+  "verificationMethod": [{
+    "id": "https://controller.example/123#key-456",
+    "type": "ExampleVerificationMethodType",
+    "controller": "https://controller.example/123",
+    <span class="comment">// public cryptographic material goes here</span>
+  }],
+  "authentication": ["#key-456"]
+}
+        </pre>
```
TallTed, line 1918: The line breaks in this `<pre>` element are not preserved in the preview. The doc should be checked after this PR is merged, to ensure that the line breaks are preserved there.

TallTed (changes requested):
A few small tweaks and requests.

peacekeeper (approved):

index.html
msporny: I just checked on my local copy, the formatting survives... this is probably an issue w/ the PR-Preview tool.

msporny:

PR #91, opened 2023-04-12
Use Specref VC-DATA-MODEL-2.0 ref.

davidlehn:
Specref has VC-DATA-MODEL-2.0 but not VC-DATA-MODEL-2. The use of VC-DATA-MODEL-2 in terms.html is an issue for other vc-di-* specs that include it directly. They can't find the Specref reference. This patch removes the localBiblio entry and adjusts references to the known Specref name.

An alternate approach would be to add VC-DATA-MODEL-2 to Specref.  I'm not sure what best practice is in this case.

***
<a href="https://pr-preview.s3.amazonaws.com/davidlehn/vc-data-integrity/pull/91.html" title="Last updated on Apr 12, 2023, 9:35 PM UTC (9b776b2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/91/47f3a4c...davidlehn:9b776b2.html" title="Last updated on Apr 12, 2023, 9:35 PM UTC (9b776b2)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -138,17 +138,6 @@
               ],
               status: "CG-DRAFT",
               publisher: "Credentials Community Group"
-            },
-            "VC-DATA-MODEL-2": {
-              title: "Verifiable Credentials Data Model v2.0",
-              href: "https://www.w3.org/TR/vc-data-model-2.0/",
-              authors: [
-                "Manu Sporny", "Dave Longley", "Grant Noble", "Dan Burnett",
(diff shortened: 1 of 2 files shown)


msporny (approved):
[[VC-DATA-MODEL-2.0]] makes me sad. I don't know what the proper thing to do here is, I'd prefer to just use the major version.

Editorial, one review, no objections, merging.

PR #105, opened 2023-06-30
Fix examples and point to specifications

msporny:
This PR fixes the introductory examples and points to the respective specifications so that people can find the cryptosuites used in the examples.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/105.html" title="Last updated on Jul 9, 2023, 6:00 PM UTC (3b1631f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/105/096eb30...3b1631f.html" title="Last updated on Jul 9, 2023, 6:00 PM UTC (3b1631f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -547,8 +547,8 @@ by adding the parameters outlined in this section:
     "title": "Hello world!",
     "proof": {
       "type": "DataIntegrityProof",
-      "cryptosuite": "example-signature-2022",
-      "created": "2020-11-05T19:23:24Z",
+      "cryptosuite": "jcs-eddsa-2022",
+      "created": "2023-03-05T19:23:24Z",
       "verificationMethod": "https://di.example/issuer#z6MkjLrk3gKS2nnkeWcmcxi
         ZPGskmesDpuwRBorgHxUXfxnG",
       "proofPurpose": "assertionMethod",
@@ -559,11 +559,10 @@ by adding the parameters outlined in this section:
         </pre>
 
         <p>
-The example proof above suggests a ficticious
-`example-signature-2022` cryptography suite that produces a
-verifiable digital proof by presumptively transforming the input data using the
-JSON Canonicalization Scheme [[?RFC8785]] and then digitally signing it using an
-Ed25519 elliptic curve signature.
+The proof above uses the `jcs-eddsa-2022` cryptography suite [[?DI-EDDSA]] that
+produces a verifiable digital proof by transforming the input data using the
+JSON Canonicalization Scheme (JCS) [[?RFC8785]] and then digitally signing it
+using an Edwards Digital Signature Algorithm (EdDSA).
         </p>
 
         <p>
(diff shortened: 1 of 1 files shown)


brentzundel (approved):

dlongley (approved):

Wind4Greg (approved):
Very nice update with examples from both ECDSA and EdDSA signatures and two different canonicalization schemes.

index.html
```diff
@@ -559,11 +559,10 @@ <h3>Proofs</h3>
         </pre>
 
         <p>
-The example proof above suggests a ficticious
-`example-signature-2022` cryptography suite that produces a
-verifiable digital proof by presumptively transforming the input data using the
-JSON Canonicalization Scheme [[?RFC8785]] and then digitally signing it using an
-Ed25519 elliptic curve signature.
+The proof above uses the `jcs-eddsa-2022` cryptography suite [[?DI-EDDSA]] that
```
TallTed: I generally find things are better comprehended when the description (and any relevant warnings, etc.) precedes (rather than follows) the examples. This will take some minutes to adjust document-wide, along with changing all the `above` to `below`. I can do, within this PR, or as a standalone.....

msporny: Agree, @TallTed, thank you! Feel free to do that in this PR or one that layers on top of this one. Up to you, I'll take the changes either way.

msporny: Nevermind, @TallTed, I went ahead an implemented your suggestion in https://github.com/w3c/vc-data-integrity/pull/105/commits/7575750ed5c24343e2e9152bcd86169956648dfe.

index.html
```diff
@@ -531,85 +531,78 @@ <h3>Proofs</h3>
...
+  "title": "Hello world!",
+  "proof": {
+    "type": "DataIntegrityProof",
+    "cryptosuite": "jcs-eddsa-2022",
+    "created": "2023-03-05T19:23:24Z",
+    "verificationMethod": "https://di.example/issuer#z6MkjLrk3gKS2nnkeWcmcxi
+      ZPGskmesDpuwRBorgHxUXfxnG",
+    "proofPurpose": "assertionMethod",
+    "proofValue": "zQeVbY4oey5q2M3XKaxup3tmzN4DRFTLVqpLMweBrSxMY2xHX5XTYV8nQA
+      pmEcqaqA3Q1gVHMrXFkXJeV6doDwLWx"
```
TallTed: This makes it look like the inline whitespace in the `verificationMethod` and `proofValue` will be ignored by tools that make use of these values. I'm pretty sure that's *not* the reality, so I think the above should become the following:
```suggestion
    "verificationMethod": "https://di.example/issuer#z6MkjLrk3gKS2nnkeWcmcxiZPGskmesDpuwRBorgHxUXfxnG",
    "proofPurpose": "assertionMethod",
    "proofValue": "zQeVbY4oey5q2M3XKaxup3tmzN4DRFTLVqpLMweBrSxMY2xHX5XTYV8nQApmEcqaqA3Q1gVHMrXFkXJeV6doDwLWx"
```
It *might* make sense to put `&nbsp;` between the keys and the values, so that the wraps are guaranteed to be in the long strings....

msporny: I attempted a fix in 3b1631f38ad34248a622f88ba44408f902c9aa86. The benefit is that people can now just copy-paste the examples AND the line wrapping happens automatically on small screens. The downside is that the examples are less readable w/o the indentation. 

If we want to do better here, we will need to write a respec plugin that automatically wraps and indents long lines in &lt;pre> HTML tags, which would be a non-trivial undertaking if we wanted highlighting to work as well. CSS is hard.

msporny:

index.html
TallTed: `wbr` works for me

PR #111, opened 2023-07-05
Add Internationalization Considerations section.

msporny:
This PR adds an internationalization considerations section, even though the spec is largely a cryptographic enveloping mechanism (and thus doesn't interact with individuals directly). There is one consideration related to date/time values that was worth mentioning, so that's been added.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/111.html" title="Last updated on Jul 13, 2023, 9:41 PM UTC (bae11c6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/111/32a436b...bae11c6.html" title="Last updated on Jul 13, 2023, 9:41 PM UTC (bae11c6)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2415,5 +2415,20 @@ Section <a href="#unlinkability"></a>.
 
     </section>
 
+    <section>
+      <h2>Internationalization Considerations</h2>
+      <p>
+The following section describes internationalization considerations that
+developers implementing this specification are urged to consider in order to
+ensure that their software is usable by people from different cultures. As a
+general rule, this specification is used by system software and does not
+directly expose individuals to information subject to internationalization
+considerations. However, there are instances where individuals might be
+indirectly exposed to information expressed by this specification and thus the
+guidance below is provided for those situations.
+      </p>
+
+
+    </section>
 </body>
 </html>
--- a/index.html
+++ b/index.html
@@ -606,6 +606,20 @@ Signature Algorithm (ECDSA).
 }
         </pre>
 
+        <p class="note" title="Representing time values to individuals">
+This specification enables the expression of dates and times, such as through
+the `created` and `expires` properties. This information might be indirectly
+exposed to an individual if a proof is processed and is detected to be outside
+an allowable time range. When displaying date and time values related to the
+validity of cryptographic proofs, implementers are advised to respect the
+<a date-cite="LTLI#locale" href="https://www.w3.org/TR/ltli/#locale">locale</a>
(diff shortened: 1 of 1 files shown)


iherman (approved):

TallTed (approved):

index.html
```diff
@@ -2487,5 +2487,33 @@ <h3>Selective Disclosure</h3>
 
     </section>
 
+    <section>
+      <h2>Internationalization Considerations</h2>
```
aphillips: The I18N WG generally does not request the creation of "Internationalization Considerations" sections unless there is a really specific message that needs to be communicated. We generally prefer that linguistic, cultural, regional, and other considerations be dealt with in the section(s) that are applicable (and where guidance will most likely be seen by implementers). 

I would be much happier if you put a note block into the "representing time" section (the original one linked to but the new "representing time" sub-section in this PR) that says something like:

```html
<!-- note this links to a term in the i18n-glossary -->
<div class="note">
<p>When displaying date and time values related to the
    validity of cryptographic proofs, implementers should 
    respect the <a>locale</a> and possibly local calendar
    preferences of the user. 
    Conversion of timestamps to local time values should
    consider the time zone expectations of the user.</p>
</div>
```

aphillips (changes requested):

Wind4Greg (approved):

index.html
Wind4Greg: So we could just put your recommended text alongside the **created** parameter in the  **Proofs** (section 2.1) which is the only place we have a date/time.

aphillips: @Wind4Greg exactly so.

Note that I18N in in the process of doing a review and there is a `pending` issue related to this (with different textual recommendations in it) [here](https://github.com/w3c/i18n-activity/issues/1729)

Wind4Greg: Thanks @aphillips we'll want to be consistent with the VC Data Model 2.0 and I particularly like your point about handling cases with/without timezone information!

dlongley (approved):
+1 and happy for this to be reconfigured in the way @aphillips suggested.

index.html
msporny: Ok, I'll make the change as you suggest @aphillips -- thank you. I'll also try to cross-link to the vc-data-model section on time representation so we take advantage of the text over there as well.

msporny: Fixed in bae11c651e9964ff1f8c7dc585aa65cc54bce0d9.

msporny:

Issue #66, opened 2022-10-21
Cleanly separate cryptosuite verification and purpose verification algorithms


msporny:
@dlongley wrote:

I think we need a cleaner separation of cryptosuite verification and purpose verification. We should:

1. Find a proof that matches the expected purpose; if none is found, raise an error.
2. Verify the proof using the cryptosuite. This returns `true` or `false` or throws if the operation cannot be executed.
3. If cryptosuite verification failed, return that the proof is not verified, otherwise...
4. Verify the proof against the expected purpose's specific checks (checking 'domain' and 'challenge' and so on are run here for authentication proofs, allowing for other purposes to be slotted in, i.e., "authentication" isn't the only known proof purpose).
5. Finally, return whether the proof is verified or not.

_Originally posted by @dlongley in https://github.com/w3c/vc-data-integrity/pull/64#discussion_r997527870_

TallTed:
> 3. If cryptosuite verification failed, return that the proof is not verified, otherwise...

For clarity in working through this, I think that should be reworded to

> 3. If cryptosuite verification does not return `true` (in other words, if cryptosuite verification returns `false` or throws an error), return that the proof is not verified, otherwise...

At this point, there will be some debate about the return (i.e., just "not verified", or some detail like "not verified due to error xyz").  I think the level of returned detail should be left to the implementer and/or deployer (better the latter), and that any errors thrown should be recorded/returned to the server-admin, whether via logs, configurable messaging, or otherwise.

mprorock:
should we consider using [RFC 7807](https://www.rfc-editor.org/rfc/rfc7807) for error responses?

msporny:
> should we consider using [RFC 7807](https://www.rfc-editor.org/rfc/rfc7807) for error responses?

It feels like it would be a good idea to consider RFC7807 for use in VC API, and should figure out a way to map what's happening in the low-level libraries that implement Data Integrity to something that can be exposed through that mechanism.

The thing I'm concerned about are things like C/C++ libraries that only have integer return values -- those implementers have traditionally railed against not being able to pick their own error codes (as integers).

msporny:
@dlongley The specification now contains a section on verification method retrieval and validation here:

https://www.w3.org/TR/vc-data-integrity/#retrieve-verification-method

It also has a section on proof purpose validation here:

https://www.w3.org/TR/vc-data-integrity/#verify-proof

Are these thing enough to address your initial issue raised?

dlongley:
[...]

PR #134, opened 2023-07-29
Explain use of each verify proof option.

msporny:
This PR addresses issue #66 by explaining what each verify proof option is used for.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/134.html" title="Last updated on Aug 5, 2023, 8:32 PM UTC (4d802f7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/134/28a168f...4d802f7.html" title="Last updated on Aug 5, 2023, 8:32 PM UTC (4d802f7)">Diff</a>


Wind4Greg (approved):

index.html
```diff
@@ -1962,8 +1962,20 @@ <h3>Verify Proof</h3>
 integrity of a <a>secured data document</a> by verifying its
 digital proof. Required inputs are a
 <a>secured data document</a> (<var>securedDocument</var>) and
-<a>proof options</a> (<var>options</var>). A <dfn>verification result</dfn> is
-produced as output.
+<a>proof options</a> (<var>options</var>). The <a>proof options</a> MAY contain
+the following values that help protect against relay and replay attacks:
+<var>expectedProofPurpose</var>,
+<var>acceptableCreatedTimeDeviationInSeconds</var>, <var>domain</var>, and
```
dlongley: I'd prefer to see this renamed to something like `maxTimestampDelta` so it's both shorter and can be used for any time comparisons, both for `created` and `expires` generally.

index.html
```diff
@@ -1962,8 +1962,20 @@ <h3>Verify Proof</h3>
...
+<var>challenge</var>. The <var>expectedProofPurpose</var> value is used to
+ensure that the <var>proof</var> was generated by the sender for the expected
+reason by the receiver, such as "authentication". The
```
dlongley: I don't think `sender` is quite right terminology. If we just look at the usage of DI with VCs for example, the issuer will create a proof that they send to the holder, but then the holder becomes the sender when presenting it to a verifier. The holder would not get to change the proof purpose for the proof created by the issuer in that case. Perhaps "proof creator" would be better here -- and `receiver` should then be `verifier`.

```suggestion
ensure that the <var>proof</var> was generated by the proof creator for the expected
reason by the verifier, such as "authentication". The
```

index.html
```diff
@@ -1962,8 +1962,20 @@ <h3>Verify Proof</h3>
...
+<var>acceptableCreatedTimeDeviationInSeconds</var> value is used to ensure that
+the <var>created</var> value is timely. The <var>domain</var> value is used
+by the sender to lock a proof to a particular security domain, and used by
```
```suggestion
by the proof creator to lock a proof to a particular security domain, and used by
```

index.html
```diff
@@ -1962,8 +1962,20 @@ <h3>Verify Proof</h3>
...
+the receiver to ensure that a proof is not being used across different security
```
```suggestion
the verifier to ensure that a proof is not being used across different security
```

index.html
```diff
@@ -1962,8 +1962,20 @@ <h3>Verify Proof</h3>
...
+domains. The <var>challenge</var> value is used by the receiver to ensure that
```
```suggestion
domains. The <var>challenge</var> value is used by the verifier to ensure that
```

index.html
```diff
@@ -1962,8 +1962,20 @@ <h3>Verify Proof</h3>
...
+the sender is not replaying previously created proofs.
```
```suggestion
an attacker is not replaying previously created proofs.
```

dlongley (approved):
Thanks! Approved with some suggestions.

index.html
dlongley: In fact, I recommend we remove being prescriptive about checking times in the algorithms and instead mention that implementations may provide interfaces to allow applications to use business rules to check the timeliness of proofs. Otherwise, we'll end up having more complex algorithms where the "current time" can be passed in and so on -- all checks that implementations can decide how to / not to expose information. What's important for our specs is that cryptosuites secure over these parameters so they are tamper-evident -- such that applications / implementations can check them as they see fit.

iherman (approved):
(This is a purely editorial comment, can be done later): I wonder whether it is worth to collect, eg, in an informational appendix, all the error conditions with references. It may help implementors.

msporny:
> (This is a purely editorial comment, can be done later): I wonder whether it is worth to collect, eg, in an informational appendix, all the error conditions with references. It may help implementors.

Yes, agreed. I've raised issue #136 to track this request. @mprorock requested something similar in https://github.com/w3c/vc-data-integrity/issues/66#issuecomment-1300669373 .

index.html
msporny: Ok, I'll remove it in the next iteration.

TallTed (approved):
Might need a tracking issue for removal of `acceptableCreatedTimeDeviationInSeconds` and adding some of @dlongley's language

dmitrizagidulin (approved):
:+1:

index.html
msporny: Removed in 4d802f7cd6bcaa039b1835d4ded917b1224bd30f.

msporny:
> Might need a tracking issue for removal of acceptableCreatedTimeDeviationInSeconds and adding some of @dlongley's language

I got rid of it in 4d802f7cd6bcaa039b1835d4ded917b1224bd30f, no need to track it. :)

Issue #130, opened 2023-07-28
Add JsonWebKey definition in Data Integrity


msporny:
Since the `JsonWebSignature2020` specification has been pulled, we need to define `JsonWebKey` so that implementers can use it if they so choose to use `publicKeyJwk`. This is just a tracking issue to make sure we do that.

msporny:
Sorry, @dmitrizagidulin -- you were assigned to this, I was addressing issue #73 and accidentally ended up writing the PR for this issue in the process.

msporny:
PR #135 has been raised to address this issue. This issue will be closed once #135 is merged.

msporny:
PR #135 has been merged, closing.

Issue #73, opened 2022-12-29
domain of publicKeyJWK should be rdf:JSON


bblfish:
The json-ld contexts for publicKeyJWK make the value be a rdf:JSON, which cleverly allows the unobtrusive integration of the key as a literal into JSON. But the spec has the range be `rdfs:string`.

So in the [Http Signature protocol](https://github.com/bblfish/authentication-panel/blob/main/proposals/HttpSignature.md) I am developing
I wrote the key in JSON-LD like this

```json
{
"@context": [
    "https://w3id.org/security/v1",
    { "ex": "http://example.org/vocab#" }
  ],
  "id": "#",
  "controller": "/people/alice#i",
  "publicKeyJwk": { 
        "kty":"RSA",
        "e":"AQAB",
        "n" : "hAKYdtoeoy8zcAcR874L8cnZxKzAGwd7v36APp7Pv6Q2jdsPBRrwWEBnez6d0UDKDwGbc6nxfEXAy5mbhgajzrw3MOEt8uA5txSKobBpKDeBLOsdJKFqMGmXCQvEG7YemcxDTRPxAleIAgYYRjTSd_QBwVW9OwNFhekro3RtlinV0a75jfZgkne_YiktSvLG34lw2zqXBDTC5NHROUqGTlML4PlNZS5Ri2U4aCNx2rUPRcKIlE0PuKxI4T-HIaFpv8-rdV6eUgOrB2xeI1dSFFn_nnv5OoZJEIB-VmuKn3DCUcCZSFlQPSXSfBDiUGhwOw76WuSSsf1D4b_vLoJ10w",
        "alg":"PS512",
        "kid":"2021-04-01-laptop"
  }
}
```
Which I worked out to be equivalent to the following turtle

```Turtle
@prefix security: <https://w3id.org/security#> .

</keys/alice#>
     security:controller </people/alice#i> ;
     security:publicKeyJwk """{
            "alg":"PS512",
            "e":"AQAB",
            "kid":"2021-04-01-laptop",
            "kty":"RSA",
[...]

PR #135, opened 2023-07-29
Add JSON Web Key to specification

msporny:
This PR addresses issue #130 and issue #73 by adding the `JsonWebKey` type to the specification and the vocabulary, and ensuring that the range of `publicKeyJwk` is JSON. This value was going to be defined in the JsonWebSignature2020 specification, but since that specification has been pulled, we have to define it somewhere else and the Data Integrity specification is probably the best place for it since it also defined controller documents and verification methods.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/135.html" title="Last updated on Aug 12, 2023, 7:26 PM UTC (1852514)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/135/b809599...1852514.html" title="Last updated on Aug 12, 2023, 7:26 PM UTC (1852514)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -115,6 +115,13 @@
               status:   "WD",
               publisher:  "W3C Verifiable Credentials Working Group"
             },
+            "JOSE-REGISTRIES": {
+              title:    "The JSON Object Signing and Encryption (JOSE) Registries",
+              href:     "https://www.iana.org/assignments/jose/jose.xhtml",
+              authors:  ["The Internet Assigned Numbers Authority"],
+              status:   "REC",
+              publisher:  "The Internet Assigned Numbers Authority"
+            },
             "SECURITY-VOCABULARY": {
               title:    "The Security Vocabulary",
               href:     "https://w3id.org/security",
@@ -157,7 +164,7 @@
               ],
               status: "ED",
               publisher: "W3C Verifiable Credentials Working Group"
-            },
+            }
           },
           postProcess: [restrictRefs],
           otherLinks: [{
@@ -1037,7 +1044,7 @@ both properties above is shown below.
       <span class="comment">...</span>
       "verificationMethod": [{
         "id": "did:example:123#_Qq0UL2Fq651Q0Fjd6TvnYE-faHiOpRlPVQcY_-tA4A",
-        "type": "JsonWebKey2020", <span class="comment">// external (property value)</span>
+        "type": "JsonWebKey", <span class="comment">// external (property value)</span>
         "controller": "did:example:123",
         "publicKeyJwk": {
           "crv": "Ed25519", <span class="comment">// external (property name)</span>
@@ -1108,6 +1115,82 @@ which is the [[?MULTIBASE]] header that conveys that the binary data is
 base58-encoded using the Bitcoin base-encoding alphabet. The decoded binary data
 [[?MULTICODEC]] header is `0x1300`, which specifies that the remaining data
 is a 32-byte raw Ed25519 private key.
+            </p>
+
+          </section>
+
+          <section>
+            <h3>JsonWebKey</h3>
+            <p>
+The JSON Web Key data model is a specific type of <a>verification method</a>
+that utilizes the JWK specification [[RFC7517]] to encode key types into a
+set of parameters. To encode a JSON Web Key, the <a>verification method</a>
+`type` MUST be set to `JsonWebKey` and the `publicKeyJwk` value MUST be a
+valid JWK value as described in [[RFC7517]]. An example of a JSON Web Key is
+provided below:
+            </p>
+
+            <pre class="example nohighlight"
+              title="JSON Web Key encoding of a Ed25519 public key">
+{
+  "@context": ["https://w3id.org/security/jwk/v1"],
+  "id": "did:example:123456789abcdefghi#keys-1",
+  "type": "JsonWebKey",
+  "controller": "did:example:123456789abcdefghi",
+  "publicKeyJwk": {
+    "kty": "OKP",
+    "alg": "EdDSA"
+    "crv": "Ed25519",
+    "kid": "did:example:123456789abcdefghi#keys-1",
+    "x": "_1EiHquO2aUx9JARSu0P8jdYT_OVneYxYOnOMAmUcFI",
+  }
+}
+            </pre>
+
+            <p>
+In the example above, the `publicKeyJwk` value contains the JSON Web Key.
+The `kty` property encodes the key type of "OKP", which means
+"Octet string key pairs". The `alg` property identifies the algorithm intended
+for use with the public key. The `crv` property identifies the particular
+curve type of the public key. The `kid` property specifies how the public key
+might be referenced in software systems; if present, the value SHOULD match the
+`id` property of the encapsulating `JsonWebKey` object. Finally, the `x`
+property specifies the point on the Ed25519 curve that is associated with the
+public key.
+            </p>
+
+            <p>
+The `publicKeyJwk` property MUST NOT contain any property marked as
+"Private" in any registry contained in the JOSE Registries [[JOSE-REGISTRIES]].
+            </p>
+
+            <p>
+The JSON Web Key data model is also capable of encoding secret keys, sometimes
+referred to as <em>private keys</em>.
+            </p>
+
+            <pre class="example nohighlight"
+              title="JSON Web Key encoding of a Ed25519 secret key">
+{
+  "@context": ["https://w3id.org/security/jwk/v1"],
+  "id": "did:example:123456789abcdefghi#keys-1",
+  "type": "JsonWebKey",
+  "controller": "did:example:123456789abcdefghi",
+  "secretKeyJwk": {
+    "kty": "OKP",
+    "alg": "EdDSA"
+    "crv": "Ed25519",
+    "kid": "did:example:123456789abcdefghi#keys-1",
+    "d": "Q6JwjCUdThSnoxfXHSFt5C1nVFycY_ZpW7qVzK644_g",
+    "x": "_1EiHquO2aUx9JARSu0P8jdYT_OVneYxYOnOMAmUcFI",
+  }
+}
+            </pre>
+
+            <p>
+The example above is identical to the previous example, except that the
+information is stored in the `secretKeyJwk` property and its value encodes the
+private key value in the `d` property.
             </p>
 
           </section>
--- a/index.html
+++ b/index.html
@@ -1122,8 +1122,8 @@ is a 32-byte raw Ed25519 private key.
           <section>
             <h3>JsonWebKey</h3>
             <p>
-The JSON Web Key data model is a specific type of <a>verification method</a>
-that utilizes the JWK specification [[RFC7517]] to encode key types into a
+The JSON Web Key (JWK) data model is a specific type of <a>verification method</a>
+that uses the JWK specification [[RFC7517]] to encode key types into a
 set of parameters. To encode a JSON Web Key, the <a>verification method</a>
 `type` MUST be set to `JsonWebKey` and the `publicKeyJwk` value MUST be a
 valid JWK value as described in [[RFC7517]]. An example of a JSON Web Key is
@@ -1131,7 +1131,7 @@ provided below:
             </p>
 
             <pre class="example nohighlight"
-              title="JSON Web Key encoding of a Ed25519 public key">
+              title="JSON Web Key encoding of an Ed25519 public key">
 {
   "@context": ["https://www.w3.org/ns/security/jwk/v1"],
   "id": "did:example:123456789abcdefghi#key-1",
@@ -1153,7 +1153,7 @@ The `kty` property encodes the key type of "OKP", which means
 "Octet string key pairs". The `alg` property identifies the algorithm intended
 for use with the public key. The `crv` property identifies the particular
 curve type of the public key. The `kid` property specifies how the public key
-might be referenced in software systems; if present, the value SHOULD match the
+might be referenced in software systems; if present, the `kid` value SHOULD match the
 `id` property of the encapsulating `JsonWebKey` object. Finally, the `x`
 property specifies the point on the Ed25519 curve that is associated with the
 public key.
@@ -1165,12 +1165,12 @@ The `publicKeyJwk` property MUST NOT contain any property marked as
             </p>
 
             <p>
-The JSON Web Key data model is also capable of encoding secret keys, sometimes
+The JSON Web Key data model is also capable of encoding <em>secret keys</em>, sometimes
 referred to as <em>private keys</em>.
             </p>
 
             <pre class="example nohighlight"
-              title="JSON Web Key encoding of a Ed25519 secret key">
+              title="JSON Web Key encoding of an Ed25519 secret key">
 {
   "@context": ["https://www.w3.org/ns/security/jwk/v1"],
   "id": "did:example:123456789abcdefghi#key-1",
@@ -1188,9 +1188,11 @@ referred to as <em>private keys</em>.
             </pre>
 
             <p>
-The example above is identical to the previous example, except that the
-information is stored in the `secretKeyJwk` property and its value encodes the
-private key value in the `d` property.
+The private key example above is almost identical to the previous example of the
+public key, except that the information is stored in the `secretKeyJwk` property
+(rather than the `publicKeyJwk`), and the private key value is encoded in the `d`
+property thereof (alongside the `x` property, which still specifies the point on
+the Ed25519 curve that is associated with the public key).
             </p>
 
           </section>
--- a/index.html
+++ b/index.html
@@ -960,28 +960,25 @@ using the `<a>controller</a>` property at the highest level of the
 
             <p>
 Verification material is any information that is used by a process that applies
-a <a>verification method</a>. The `type` of a <a>verification
-method</a> is expected to be used to determine its compatibility with such
-processes. Examples of verification material properties are
-`publicKeyJwk` or `publicKeyMultibase`. A
-<a>cryptographic suite</a> specification is responsible for specifying the
-<a>verification method</a> `type` and its associated verification
-material. For example, see <a href="https://w3c-ccg.github.io/lds-jws2020/">JSON
-Web Signature 2020</a> and <a
-href="https://w3c-ccg.github.io/lds-ed25519-2020/">Ed25519 Signature 2020</a>.
-For all registered <a>verification method</a> types and associated verification
-material available for <a>controllers</a>, please see the Data Integrity
-Specification Registries [TBD - DIS-REGISTRIES].
+a <a>verification method</a>. The `type` of a <a>verification method</a> is
+expected to be used to determine its compatibility with such processes. Examples
+of verification material include `JsonWebKey` and `Multikey`.
+A <a>cryptographic suite</a> specification is responsible for specifying the
+<a>verification method</a> `type` and its associated verification material
+format. For examples, see
+<a href="https://www.w3.org/TR/vc-di-ecdsa/">the Data Integrity ECDSA
+Cryptosuites</a> and <a href="https://w3c-ccg.github.io/lds-ed25519-2020/">
+the Data Integrity EdDSA Cryptosuites</a>. For a list of <a>verification
+method</a> types please see the [[?SECURITY-VOCABULARY]].
             </p>
 
             <p>
 To increase the likelihood of interoperable implementations, this specification
-limits the number of formats for expressing verification material in a <a>controller
-document</a>. The fewer formats that implementers have to
+limits the number of formats for expressing verification material in a
+<a>controller document</a>. The fewer formats that implementers have to
 implement, the more likely it will be that they will support all of them. This
 approach attempts to strike a delicate balance between ease of implementation
 and supporting formats that have historically had broad deployment.
-Two supported verification material properties are listed below:
             </p>
 
             <p>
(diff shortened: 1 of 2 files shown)


mkhraisha (approved):

Wind4Greg (approved):

dlongley (approved):

vocab/security/vocabulary.yml
```diff
@@ -3,6 +3,8 @@ vocab:
   value: https://w3id.org/security#
 
 prefix:
+  - id: rdf
+    value: http://www.w3.org/1999/02/22-rdf-syntax-ns#
```
iherman: Strictly speaking, this is not necessary (although it does not harm). The tool has a number default prefixes which include rdf, rdfs, dc, owl, xsd.

```suggestion
```

iherman (approved):
Just an editorial comment, possibly with an editorial note. It may be worth emphasizing that the `publicKeyJwk` having a `JSON` datatype means that all json terms _within_ the target are not subject to the JSON-LD context processing. Ie, the terms like `kid` or `alg` are _not_ mapped to any URL.

index.html
```diff
@@ -1083,6 +1090,82 @@ <h3>Multikey</h3>
...
+
+            <p>
+The JSON Web Key data model is also capable of encoding secret keys, sometimes
+referred to as <em>private keys</em>.
+            </p>
+
+            <pre class="example nohighlight"
+              title="JSON Web Key encoding of a Ed25519 secret key">
+{
+  "@context": [""https://www.w3.org/ns/security/jwk/v1""],
```
```suggestion
  "@context": ["https://www.w3.org/ns/security/jwk/v1"],
```

vocab/security/vocabulary.yml
```diff
@@ -43,6 +45,11 @@ class:
       - label: EdDSA Cryptosuite v2022
         url: https://www.w3.org/TR/vc-di-eddsa/#multikey
 
+  - id: JsonWebKey
+    label: JSON Web Key Verification Method
+    upper_value: sec:VerificationMethod
+    defined_by: https://www.w3.org/TR/vc-data-integrity/#jsonwebkey
```
iherman, line 49: I did not see the `<dfn>` statement in the text...

iherman (changes requested):

vocab/security/vocabulary.yml
msporny: Ah! Thank you, I checked the yml2vocab tool and didn't see the definition. Great that it exists, removing the `rdf` definition now.

vocab/security/vocabulary.yml
msporny: ReSpec auto-generates it based on this line: https://github.com/w3c/vc-data-integrity/pull/135/files#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051R1094

thereisnogabe (approved):

index.html
```diff
@@ -112,6 +112,13 @@
               status:   "WD",
               publisher:  "W3C Verifiable Credentials Working Group"
             },
+            "JOSE-REGISTRIES": {
+              title:    "The JSON Object Signing and Encryption (JOSE) Registries",
+              href:     "https://www.iana.org/assignments/jose/jose.xhtml",
```
OR13:
```suggestion
              href:     "https://www.iana.org/assignments/jose",
```

OR13: Possible this will change in the future, better to not ask for XML by file extension

index.html
```diff
@@ -1083,6 +1090,82 @@ <h3>Multikey</h3>
...
+    "kid": "key-1",
+    "d": "Q6JwjCUdThSnoxfXHSFt5C1nVFycY_ZpW7qVzK644_g",
+    "x": "_1EiHquO2aUx9JARSu0P8jdYT_OVneYxYOnOMAmUcFI",
+  }
+}
+            </pre>
+
+            <p>
+The example above is identical to the previous example, except that the
+information is stored in the `secretKeyJwk` property and its value encodes the
```
OR13: hmmm... `privateKeyJwk` vs `secretKeyJwk`... Is the intention to make this work with MAC Signatures? and thats why you choose `secretKeyJwk` ?

index.html
```diff
@@ -1083,6 +1090,82 @@ <h3>Multikey</h3>
...
+private key value in the `d` property.
```
OR13: This seems possibly not true for all secret keys.

vocab/security/vocabulary.yml
```diff
@@ -237,7 +242,7 @@ property:
 
   - id: publicKeyJwk
```
OR13, line 272: Expecting to see `secretKeyJwk` defined in this vocabulary as well... otherwise how will people understand how to use it?

OR13 (changes requested):
We should get more reviews on the `secretKeyJwk` part... will it be in the vocabulary? can it be used with HMAC?.

Concrete change request is to define `secretKeyJwk` in the vocabulary or remove the example.

I'm not attached to `privateKeyJwk` vs `secretKeyJwk`... perhapse it would be good to ask the experts on the JOSE and COSE mailing lists how they feel about this?

index.html
mprorock: this is definitely not true for all secret keys, and may yet be changing - why would we define any of this here rather than just point at rfc7517?

mprorock:
+1 @OR13 - definitely want feedback in from the jose and cose lists

mprorock (changes requested):
lets get feedback requests out to appropriate experts - clean up some naming / guidance or just point to appropriate RFCs, otherwise i think this good

index.html
```diff
@@ -1083,6 +1090,82 @@ <h3>Multikey</h3>
 
           </section>
 
+          <section>
+            <h3>JsonWebKey</h3>
+            <p>
+The JSON Web Key data model is a specific type of <a>verification method</a>
+that utilizes the JWK specification [[RFC7517]] to encode key types into a
```
```suggestion
that uses the JWK specification [[RFC7517]] to encode key types into a
```

index.html
```diff
@@ -1083,6 +1090,82 @@ <h3>Multikey</h3>
 
           </section>
 
+          <section>
+            <h3>JsonWebKey</h3>
+            <p>
+The JSON Web Key data model is a specific type of <a>verification method</a>
```
```suggestion
The JSON Web Key (JWK) data model is a specific type of <a>verification method</a>
```

index.html
```diff
@@ -1083,6 +1090,82 @@ <h3>Multikey</h3>
...
+}
+            </pre>
+
+            <p>
+In the example above, the `publicKeyJwk` value contains the JSON Web Key.
+The `kty` property encodes the key type of "OKP", which means
+"Octet string key pairs". The `alg` property identifies the algorithm intended
+for use with the public key. The `crv` property identifies the particular
+curve type of the public key. The `kid` property specifies how the public key
+might be referenced in software systems; if present, the value SHOULD match the
```
```suggestion
might be referenced in software systems; if present, the `kid` value SHOULD match the
```

index.html
```diff
@@ -1083,6 +1090,82 @@ <h3>Multikey</h3>
...
+public key.
+            </p>
+
+            <p>
+The `publicKeyJwk` property MUST NOT contain any property marked as
+"Private" in any registry contained in the JOSE Registries [[JOSE-REGISTRIES]].
+            </p>
+
+            <p>
+The JSON Web Key data model is also capable of encoding secret keys, sometimes
```
```suggestion
The JSON Web Key data model is also capable of encoding <em>secret keys</em>, sometimes
```

index.html
```diff
@@ -1083,6 +1090,82 @@ <h3>Multikey</h3>
...
+The JSON Web Key data model is a specific type of <a>verification method</a>
+that utilizes the JWK specification [[RFC7517]] to encode key types into a
+set of parameters. To encode a JSON Web Key, the <a>verification method</a>
+`type` MUST be set to `JsonWebKey` and the `publicKeyJwk` value MUST be a
+valid JWK value as described in [[RFC7517]]. An example of a JSON Web Key is
+provided below:
+            </p>
+
+            <pre class="example nohighlight"
+              title="JSON Web Key encoding of a Ed25519 public key">
```
```suggestion
              title="JSON Web Key encoding of an Ed25519 public key">
```

index.html
```diff
@@ -1083,6 +1090,82 @@ <h3>Multikey</h3>
...
+"Private" in any registry contained in the JOSE Registries [[JOSE-REGISTRIES]].
+            </p>
+
+            <p>
+The JSON Web Key data model is also capable of encoding secret keys, sometimes
+referred to as <em>private keys</em>.
+            </p>
+
+            <pre class="example nohighlight"
+              title="JSON Web Key encoding of a Ed25519 secret key">
```
```suggestion
              title="JSON Web Key encoding of an Ed25519 secret key">
```

index.html
```suggestion
The private key example above is almost identical to the previous example of the
public key, except that the information is stored in the `secretKeyJwk` property
(rather than the `publicKeyJwk`), and the private key value is encoded in the `d`
property thereof (alongside the `x` property, which still specifies the point on
the Ed25519 curve that is associated with the public key).
```

TallTed (changes requested):
Mostly small editorial stuff. One larger thing which I think is also editorial and hopefully not controversial.

dmitrizagidulin (approved):
:+1: overall, (agree with Mike about not defining the specific key properties like `d` in the spec, and just pointing to the RFC)

vocab/security/vocabulary.yml
dmitrizagidulin: Yes, secretKeyJwk is added to the vocab in PR https://github.com/w3c/vc-data-integrity/pull/148

OR13:
I asked the IETF lists: https://mailarchive.ietf.org/arch/msg/cose/fuTzZtvbvSF3GQFVjI-JYOyX7Lk/

I suggest we leave the PR open for another week in case they have feedback.

EDIT: also please apply @TallTed 's suggestions.

index.html
OR13: See https://datatracker.ietf.org/doc/draft-ietf-cose-key-thumbprint/

for concrete examples...

https://mailarchive.ietf.org/arch/msg/cose/T4Od25YoPmgOuALuuIGf1m8N4wo/

index.html
OR13: maybe better to refer to the spec language.... which calls this "private information"....

https://datatracker.ietf.org/doc/html/rfc7517#section-8.1

> It also records whether the  parameter conveys public or private information.

OR13 (changes requested):
Please apply suggestions so that re-review is possible.

index.html
```diff
@@ -1083,6 +1090,82 @@ <h3>Multikey</h3>
...
+            </p>
+
+            <p>
+The `publicKeyJwk` property MUST NOT contain any property marked as
+"Private" in any registry contained in the JOSE Registries [[JOSE-REGISTRIES]].
+            </p>
+
+            <p>
+The JSON Web Key data model is also capable of encoding secret keys, sometimes
+referred to as <em>private keys</em>.
```
TallTed: Integrating [the suggestion](https://github.com/w3c/vc-data-integrity/pull/135/files#r1284562689) from @OR13 --
```suggestion
The JSON Web Key data model is also capable of encoding private information,
including <em>private keys</em> and/or <em>secret keys</em>.
```

OR13: ^ I support the language above, but in case we want to be even more precise, "private information" is not "secrete keys" or "private keys".... its parameters of "secrete keys" or "private keys"... for example:

https://www.iana.org/assignments/jose/jose.xhtml#web-key-set-parameters

Screen Shot 2023-08-04 at 5 10 31 PM

Just look at all those RSA things that are not "d" ... that are private... that is why I tried to correct the language here: 

- https://github.com/w3c/vc-data-integrity/pull/99

The advice given in data integrity regarding "d" does not actually protect people... since RSA does not use "d"... and lots of people still use RSA saddly.

TallTed: Updated language. 

> `The JSON Web Key data model is also capable of encoding private information,
including <em>private keys</em> and/or <em>secret keys</em>.`

dlongley: > I support the language above, but in case we want to be even more precise, "private information" is not "secrete keys" or "private keys".... its parameters of "secrete keys" or "private keys"... for example:

Yes, and I believe this is one of the challenges with some key formats. I think it's better for app developers to be able to treat key values as opaque and completely omit secret keys without having to know the details around validating and parsing (potentially) secret keys to look for specific internal components, etc. This is best left to another layer.

dlongley: > The advice given in data integrity regarding "d" does not actually protect people... since RSA does not use "d"... and lots of people still use RSA saddly.

This might have been a typo; RSA does use "d" -- that's where it came from originally:

https://en.wikipedia.org/wiki/RSA_(cryptosystem)#Operation

OR13: Yes, I meant to say... omitting ONLY `d` (which is the current guidance) is bad advice... 

Just like it would be bad advice to say that only uranium is radioactive... There are other radioactive elements... and there are other "private information"  parameters that should not be seen in public keys.

index.html
selfissued: I agree that privateKeyJwk is a better name than secretKeyJwk.

index.html
dlongley: > Yes, I meant to say... omitting ONLY `d` (which is the current guidance) is bad advice...

Agreed -- it's good as an example but that example should say that it's insufficient to just omit that one thing and that `<reference X>` must be checked and every possible private or secret (whichever terminology is appropriate) component listed similarly omitted.

dlongley:
> I'm not attached to `privateKeyJwk` vs `secretKeyJwk`... perhaps it would be good to ask the experts on the JOSE and COSE mailing lists how they feel about this?

Sure, but note that we're defining the interface / expression here where the most important audience is app developers who should be thoroughly made aware of "secret" information so it can be dealt with appropriately. Whether it's the best name in the opinion of a particular cryptographer or a perfect fit for a particular cryptographic scheme is not right the measure for the name, IMO. It doesn't really matter what the underlying crypto is; the value it holds is "secret" and the type for the value is a JWK -- any JWK. That's where the JOSE (or COSE if it were `secretKeyCwk`) layer would come in -- and we should just reuse those specs for the values of the properties.

index.html
msporny: > Is the intention to make this work with MAC Signatures? and thats why you choose secretKeyJwk ?

Yes, that's one of the reasons.

The other reason is due to the English definition of the word "private" vs. the English definition of the word "secret":

private: intended for or restricted to the use of a particular person, group, or class
secret: kept from knowledge or view; something kept hidden or unexplained

IOW, the word "secret", that is -- "do not share this with others", is closer to the intended usage of the property than "private", which can be misinterpreted as "its ok to share this information with a group" (which is almost never the case with a symmetric or a private key).

index.html
msporny: > This seems possibly not true for all secret keys.

The language you are pointing to is providing a description of a specific example. It is not claiming that the example is true for all secret keys.

> this is definitely not true for all secret keys, and may yet be changing - why would we define any of this here rather than just point at rfc7517?

Please re-read the /description of the example/. We are not defining anything here, merely explaining what an example is expressing.

msporny:
> We should get more reviews on the secretKeyJwk part... will it be in the vocabulary? can it be used with HMAC?.

Yes, it will be in the vocabulary. Yes, it can be used w/ HMAC (and anything else JWK can do).

> Concrete change request is to define secretKeyJwk in the vocabulary or remove the example.

That is being done in PR https://github.com/w3c/vc-data-integrity/pull/148.

> I'm not attached to privateKeyJwk vs secretKeyJwk... perhapse it would be good to ask the experts on the JOSE and COSE mailing lists how they feel about this?

I see that you've sent an email to those mailing lists. I will hold this PR an additional week for feedback and then merge. If we get feedback at any point beyond that, we can raise another PR to apply the feedback.

> Please apply suggestions so that re-review is possible.

Done.

@mprorock wrote:
> just point to appropriate RFCs
@dmitrizagidulin wrote:
> agree with Mike about not defining the specific key properties like d in the spec, and just pointing to the RFC

Can you guys point to where we're defining things that are in the RFC in this specification? We /do/ point to the RFC with this language (and DO NOT re-define properties in this spec and/or PR):

> The JSON Web Key (JWK) data model is a specific type of <a>verification method</a> that uses the JWK specification [[RFC7517]] to encode key types into a set of parameters. To encode a JSON Web Key, the <a>verification method</a> `type` MUST be set to `JsonWebKey` and the `publicKeyJwk` or `secretKeyJwk` value MUST be a valid JWK value as described in [[RFC7517]].

What am I missing?

Requesting a re-review from each of you.

vocab/security/vocabulary.yml
iherman: That is correct for the class specification. However, #148, that adds both the `secretKeyJwk` and `publicKeyJwk` properties rely on the presence of `<dfn`> for those. I see it for `publicKeyJwk` at https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/135.html#dfn-publickeyjwk, I did not find it for `secretKeyJwk` :-(

iherman (changes requested):

vocab/security/vocabulary.yml
msporny: Ah, right. I'll raise another PR to refactor the sections and put the definitions where they belong (in the sections defining both the `JsonWebKey` type and the `Multikey` type.

msporny: PR #152 has been raised to address this issue.

vocab/security/vocabulary.yml
msporny: Definitions added in PR #152

TallTed (approved):

OR13:
I agree with @iherman  that `secretKeyJwk` or `privateKeyJwk` needs to be defined for this to be merged... and the same applies to `publicKeyMultibase` or `secretKeyMultibase` or `privateKeyMultibase`.

See also https://github.com/w3c/vc-data-integrity/pull/150/files#diff-af0ab1b57396186235708fdf500251dfc3f9b18aeeb992f927979d9630889c28R21

and https://github.com/w3c/vc-data-integrity/pull/149/files#diff-be8086ba9b3e09e91dd045c04051b6e1d0d2a10105db9d380835cb1151fd5470R29

Seems blocked pending https://github.com/w3c/vc-data-integrity/pull/152#pullrequestreview-1568433288

OR13 (approved):
Pending merge / approval of https://github.com/w3c/vc-data-integrity/pull/152#pullrequestreview-1568433288

index.html
```diff
@@ -1083,6 +1090,84 @@ <h3>Multikey</h3>
...
+  "type": "JsonWebKey",
+  "controller": "did:example:123456789abcdefghi",
+  "publicKeyJwk": {
+    "kty": "OKP",
+    "alg": "EdDSA"
+    "crv": "Ed25519",
+    "kid": "key-1",
+    "x": "_1EiHquO2aUx9JARSu0P8jdYT_OVneYxYOnOMAmUcFI",
+  }
+}
```
iherman, line 1163: Why do we need yet another context here? The definition of `publicKeyJwk` stipulates that its object is of datatype JSON, meaning that the properties within that object are ignored by any JSON-LD processing. If so, then I do not see any reason using a different context (which I have not seen defined anywhere).

Proposal: just use the usual security context here.

index.html
```diff
@@ -1083,6 +1090,84 @@ <h3>Multikey</h3>
...
+  "type": "JsonWebKey",
+  "controller": "did:example:123456789abcdefghi",
+  "secretKeyJwk": {
+    "kty": "OKP",
+    "alg": "EdDSA"
+    "crv": "Ed25519",
+    "kid": "key-1",
+    "d": "Q6JwjCUdThSnoxfXHSFt5C1nVFycY_ZpW7qVzK644_g",
+    "x": "_1EiHquO2aUx9JARSu0P8jdYT_OVneYxYOnOMAmUcFI",
+  }
```
iherman, line 1202: Same comment as above. Except that we use here `secretKeyJwk`, but I presume the range of that property is also JSON

iherman (approved):
I have made some inline comments.

I must say I am lost among all the PRs that refer to various property names, anchors, ranges, etc. I am of the opinion at this point that all these PRs should be merged asap to allow for a comprehensive review of the vocabulary, domains, ranges, etc, possibly leading to new PRs. The current situation is simply too chaotic in my view.

index.html
OR13: We don't need another context here.

I will object if one is required... however, I think its fine to ship many smaller contexts from `w3id.org` (which is not the W3C)... IMO `W3C` examples should use `W3C` context URLs.

OR13: Also, seems we are following a different style here than the core context, where we seem to prefer bundling, that inconsistency, feels... not good to me.

dlongley: @OR13,

The `@vocab` in the core v2 context is what is driving the need to bundle. Whenever that isn't used, bundling should be avoided (unless there's some other very good reason to do it). That's the consistent rule we're applying.

msporny: Raised https://github.com/w3c/vc-data-integrity/issues/164 to track @OR13's concern, which seems to be more of a misunderstanding than a misalignment.

index.html
msporny: Same response as above, yes, range is JSON.

msporny:
PR #152 has been merged into this PR. It has been two weeks since we asked for feedback from the COSE mailing list (and no feedback has been given). If we get feedback in the future, we can always raise an issue and modify this via a future PR. In the meantime, this PR needs to go in so we have the appropriate anchors to finish up the vocabulary and JSON Web Key sections of the specification.

msporny:

Issue #88, opened 2023-03-04
Specify a default context if JSON-LD processing is required


msporny:
The Data Integrity specification does not require the use of JSON-LD. That said, there are times when a controller document, or a verification method, might need to be interpreted in a JSON-LD Context. For those instances, we might want to define a default context (or the use of a JSON-LD Context that is equivalent to the default context).

msporny:
Also mention that the data-integrity/v1 context does not need to be injected if the VC 2.0 context is used.

msporny:
PR #138 has been raised to address this issue. This issue will be closed when that PR is merged.

msporny:
PR #138 has been merged, closing.

PR #138, opened 2023-07-30
Add section on context injection.

msporny:
This PR attempts to address issue #88 by adding a section on context injection. This section ensures that all documents, whether they are processed as JSON or JSON-LD, end up using the same semantics for Data Integrity terms.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/138.html" title="Last updated on Aug 5, 2023, 9:28 PM UTC (ea872a1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/138/a665eb7...ea872a1.html" title="Last updated on Aug 5, 2023, 9:28 PM UTC (ea872a1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1655,6 +1655,33 @@ Credentials v2.0: Base Context</a> and <a data-cite="?VC-DATA-MODEL-2.0#vocabula
 Verifiable Credentials v2.0: Vocabularies</a>.
         </p>
 
+        <section>
+          <h3>Context Injection</h3>
+
+          <p>
+The `@context` property is used to ensure that implementations are using the
+same semantics when terms in this specification are processed. For example, this
+can be important when properties like `type` are processed and its value, such
+as `DataIntegrityProof`, are used.
+          </p>
+
+          <p>
+If a `@context` property is not provided in a document that is being secured or
+verified, or the Data Integrity terms used in the document are not defined by
+existing values in the `@context` property, implementations MUST inject or add
+an `@context` property with a value of
+`https://www.w3.org/ns/data-integrity/v1`.
+          </p>
+
+          <p>
+An example of when context injection is unnecessary is when the Verifiable
+Credential Data Model v2.0 context (`https://www.w3.org/ns/credentials/v2`)
+exists as a value in the `@context` property, as that context defines all of the
+necessary Data Integrity terms defined by
+`https://www.w3.org/ns/data-integrity/v1`.
+          </p>
+        </section>
+
       </section>
 
     </section>
--- a/index.html
+++ b/index.html
@@ -1666,18 +1666,18 @@ as `DataIntegrityProof`, are used.
           </p>
 
           <p>
-If a `@context` property is not provided in a document that is being secured or
-verified, or the Data Integrity terms used in the document are not defined by
+If an `@context` property is not provided in a document that is being secured or
+verified, or the Data Integrity terms used in the document are not mapped by
 existing values in the `@context` property, implementations MUST inject or add
 an `@context` property with a value of
 `https://www.w3.org/ns/data-integrity/v1`.
           </p>
 
           <p>
-An example of when context injection is unnecessary is when the Verifiable
+One example of when context injection is expected to be unnecessary is when the Verifiable
 Credential Data Model v2.0 context (`https://www.w3.org/ns/credentials/v2`)
-exists as a value in the `@context` property, as that context defines all of the
-necessary Data Integrity terms defined by
+exists as a value in the `@context` property, as that context maps all of the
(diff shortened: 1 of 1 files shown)


Wind4Greg (approved):
Helpful clarification on @context usage.

index.html
```diff
@@ -1655,6 +1655,33 @@ <h2>Contexts and Vocabularies</h2>
...
+an `@context` property with a value of
+`https://www.w3.org/ns/data-integrity/v1`.
+          </p>
+
+          <p>
+An example of when context injection is unnecessary is when the Verifiable
+Credential Data Model v2.0 context (`https://www.w3.org/ns/credentials/v2`)
+exists as a value in the `@context` property, as that context defines all of the
+necessary Data Integrity terms defined by
+`https://www.w3.org/ns/data-integrity/v1`.
```
iherman: Just flagging for ourselves: before going to CR we will have to crosscheck the (1) credential vocabulary (2) credential context (3) data integrity context and (4) data integrity vocabulary that all these statements are indeed correct. These vocabularies change almost constantly these days, and it is easy to miss out something...

iherman (approved):

dlongley (approved):

index.html
```diff
@@ -1655,6 +1655,33 @@ <h2>Contexts and Vocabularies</h2>
...
+
+          <p>
+The `@context` property is used to ensure that implementations are using the
+same semantics when terms in this specification are processed. For example, this
+can be important when properties like `type` are processed and its value, such
+as `DataIntegrityProof`, are used.
+          </p>
+
+          <p>
+If a `@context` property is not provided in a document that is being secured or
```
```suggestion
If an `@context` property is not provided in a document that is being secured or
```

index.html
```diff
@@ -1655,6 +1655,33 @@ <h2>Contexts and Vocabularies</h2>
...
+verified, or the Data Integrity terms used in the document are not defined by
+existing values in the `@context` property, implementations MUST inject or add
+an `@context` property with a value of
+`https://www.w3.org/ns/data-integrity/v1`.
+          </p>
+
+          <p>
+An example of when context injection is unnecessary is when the Verifiable
```
```suggestion
One example of when context injection is expected to be unnecessary is when the Verifiable
```

index.html
```diff
@@ -1655,6 +1655,33 @@ <h2>Contexts and Vocabularies</h2>
...
+Credential Data Model v2.0 context (`https://www.w3.org/ns/credentials/v2`)
+exists as a value in the `@context` property, as that context defines all of the
+necessary Data Integrity terms defined by
```
```suggestion
exists as a value in the `@context` property, as that context maps all of the
necessary Data Integrity terms that were previously mapped by
```

index.html
```diff
@@ -1655,6 +1655,33 @@ <h2>Contexts and Vocabularies</h2>
...
+          <p>
+The `@context` property is used to ensure that implementations are using the
+same semantics when terms in this specification are processed. For example, this
+can be important when properties like `type` are processed and its value, such
+as `DataIntegrityProof`, are used.
+          </p>
+
+          <p>
+If a `@context` property is not provided in a document that is being secured or
+verified, or the Data Integrity terms used in the document are not defined by
```
```suggestion
verified, or the Data Integrity terms used in the document are not mapped by
```

TallTed (changes requested):

index.html
dmitrizagidulin: I don't think the `@` symbol is pronounced, so I don't think this needs an `an`

dmitrizagidulin (approved):

index.html
TallTed: If the `@` isn't pronounced, how do you tell the difference between `context` and `@context`?

TallTed: Also, if we're going to use `a @context` here, line 1672 must be updated to match, as well as numerous instances across other documents.

I do noting that even [the JSON-LD 1.1 spec](https://www.w3.org/TR/json-ld11/#scoped-contexts) is inconsistent about this! At the link, you'll find `include a @context property` in the first paragraph of text in §4.1.8, and `Defining an @context` in the titles of `Example 45` and `Example 46`.

That said, searching for `an @context` throughout that spec finds 6 instances, while `a @context` finds only 1 instance. I think that's a clear enough indication of the intention of the spec authors, and I'll continue to press for `an @context` here.

msporny: The omission was an error. I pronounce the "@" symbol and always strive to put "an @" when writing. I agree with the world that @TallTed wants to live in and denounce the grammatically lawless world, devoid of strange English rules that take a half a lifetime to learn, that @dmitrizagidulin is attempting to lull us into. :P

msporny:

PR #140, opened 2023-07-30
Add errors with anchors to the specification. 

msporny:
This PR is an attempt to address issue #136 by adding a table of error definitions to the specification with anchors that can be used by RFC 7807.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/140.html" title="Last updated on Aug 5, 2023, 9:40 PM UTC (9abc28c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/140/a665eb7...9abc28c.html" title="Last updated on Aug 5, 2023, 9:40 PM UTC (9abc28c)">Diff</a>

PR #141, opened 2023-07-31
Added the error terms to the vocabulary

iherman:
**This PR is not against the `main` branch but, rather, against the branch behind https://github.com/w3c/vc-data-integrity/pull/140**

The PR extends the official security vocabulary to include all the error terms, as RDF individuals, in the security vocabulary.

Note that I have **not** defined the terms described in that version ("code", "title", etc.) that are in the proposed specification. It would require to add them as bona fide properties with the `Errors` class as a domain, but I do not believe those (JSON) objects will ever appear as Linked Data. If I am wrong, then these terms should be added to the vocabulary as well (although in that case, it is a possible problem to see that the terms like 'title' or 'code' are extremely generic, hopefully it will not lead to clashes...)

### DIFF
--- a/vocab/security/vocabulary.yml
+++ b/vocab/security/vocabulary.yml
@@ -58,9 +58,13 @@ class:
     upper_value: sec:Proof
     defined_by: https://www.w3.org/TR/vc-di-eddsa/#ed25519signature2020
 
+  - id: Errors
+    label: Processing errors
+    defined_by: https://www.w3.org/TR/vc-di-eddsa/#errors
 
-# These are the class definitions in the CCG documents that are not defined in a VCWG document; they are all deprecated
-# In some cases a ccg document was found and used for the definition, but in some cases even that is missing...
+
+# These are the class definitions in the CCG documents that are not defined in a VCWG document; they are all deprecated.
+# In some cases, a CCG document was found and used for the definition, but in other cases, even that is missing...
 
 
   - id: EcdsaSecp256k1Signature2019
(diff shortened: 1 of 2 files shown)


Wind4Greg (approved):

vocab/security/vocabulary.yml
```diff
@@ -58,6 +58,10 @@ class:
     upper_value: sec:Proof
     defined_by: https://www.w3.org/TR/vc-di-eddsa/#ed25519signature2020
 
+  - id: Errors
+    label: Processing errors
+    defined_by: https://www.w3.org/TR/vc-di-eddsa/#errors
+
 
 # These are the class definitions in the CCG documents that are not defined in a VCWG document; they are all deprecated
 # In some cases a ccg document was found and used for the definition, but in some cases even that is missing...
```
```suggestion
# These are the class definitions in the CCG documents that are not defined in a VCWG document; they are all deprecated.
# In some cases, a CCG document was found and used for the definition, but in other cases, even that is missing...
```

iherman: @TallTed I will merge this but... this is only a comment in the YAML file...

dmitrizagidulin (approved):

iherman:
@msporny I have the impression that the change conflicts are related to some other problems. This PR, originally, was meant to be merged into another branch (merged since) and the rebase to main changed things.

Because the merge conflicts are in the index.html file, ie, the core spec, I would prefer you take care of those. Thx.

msporny:
> @msporny I have the impression that the change conflicts are related to some other problems. This PR, originally, was meant to be merged into another branch (merged since) and the rebase to main changed things.

The branch it was intended to be merged into (the one that added the error table) was merged into main, at which point, Github automatically changes the old target branch (the one that added the error table) into the branch that the old target branch got merged into (main).

> Because the merge conflicts are in the index.html file, ie, the core spec, I would prefer you take care of those. Thx.

I'll take care of it. :)

msporny (approved):

msporny:

PR #143, opened 2023-07-31
Correcting previously mis-corrected sentence

TallTed:
**Important questions, where the answers may lead me to change this PR:**
* Are the `security domain` and/or `receiver-supplied challenge` both optional? 
* Can either be present without the other, or does including one mandate the other? 
* Are these the only optional `proof options` (i.e., while all the others are `MUST`, these are `MAY`)?

***
<a href="https://pr-preview.s3.amazonaws.com/TallTed/vc-data-integrity/pull/143.html" title="Last updated on Aug 1, 2023, 4:12 PM UTC (161bee2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/143/28a168f...TallTed:161bee2.html" title="Last updated on Aug 1, 2023, 4:12 PM UTC (161bee2)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1837,9 +1837,9 @@ the <a>cryptographic suite</a> type; an identifier for the
 <a>verification method</a> (<var>verificationMethod</var>) that can be used to
 verify the authenticity of the proof; an [[!XMLSCHEMA11-2]] combined date
 and time string (<var>created</var>) containing the current date and time,
-accurate to at least one second, in Universal Time Code format, a
-<a href="#dfn-domain">security domain</a> (<var>domain</var>), and a
-receiver-supplied challenge (<var>challenge</var>) might also be specified in
+accurate to at least one second, in Universal Time Code format. A
+<a href="#dfn-domain">security domain</a> (<var>domain</var>) and/or a
+receiver-supplied challenge (<var>challenge</var>) MAY also be specified in
 the <var>options</var>. A <dfn>secured data document</dfn> is produced as
 output. Whenever this algorithm encodes strings, it MUST use UTF-8 encoding.
         </p>


Wind4Greg (approved):

dmitrizagidulin (approved):

msporny:
@TallTed, I've made an attempt at answering your questions below:

> * Are the `security domain` and/or `receiver-supplied challenge` both optional?

Yes.

> * Can either be present without the other, or does including one mandate the other?

Either can be present without the other.

> * Are these the only optional `proof options` (i.e., while all the others are `MUST`, these are `MAY`)?

I'm going to answer your question in a round-about way. 

On a conforming proof object, the following properties are required: `type`, `proofPurpose`, `verificationMethod`, `proofValue`

On a conforming proof object, the following properties are optional: `id`, `created`, `expires`, `domain`, `challenge`, `previousProof`, `nonce`

Implementations are expected to provide mechanisms for setting all of the fields above, which might include "convenience options" that set multiple values at the same time, or use implementer-defined defaults for both required and optional values. We are trying to stay out of the implementers way so they can create the interfaces that they want to create while ensuring that the algorithms clearly articulate where each option is used in the algorithms (which can be implemented in any way that an implementer would like, as long as the output of the final algorithm matches what is specified).

I hope that gives you enough to go on for revising this PR. If you're done, let me know, as it's ready for merge if you are done.

msporny (approved):

TallTed:
@msporny Hmmm. The re-corrected sentences are good now. We do not seem to list the same required and optional properties for a conforming proof object, in the way you have here. Perhaps we should, maybe as bullet lists before or after the algorithm (as now written), and rephrase your large paragraph into something meant for implementers?

msporny:
@TallTed wrote:
> Perhaps we should, maybe as bullet lists before or after the algorithm (as now written), and rephrase your large paragraph into something meant for implementers?

Yeah, we might want to do that... it would be easier to read here... we might want to do that for all the other sections as well. 

I'm going to merge this in now, and then look at applying your suggestions more generally throughout the rest of the algorithms in the spec.

msporny:
Normative, multiple positive reviews, no objections, merging.

PR #148, opened 2023-08-04
Add vocab definitions for secretKeyJwk and secretKeyMultibase

dmitrizagidulin:
- Addresses issue https://github.com/w3c/vc-di-eddsa/issues/4
- Addresses @OR13's comment https://github.com/w3c/vc-data-integrity/pull/135/files#r1279821900

### DIFF
--- a/vocab/security/vocabulary.yml
+++ b/vocab/security/vocabulary.yml
@@ -194,7 +194,7 @@ property:
     domain: sec:Proof
     range: xsd:string
     defined_by: https://www.w3.org/TR/vc-data-integrity/#dfn-proofvalue
-    
+
   - id: expires
     label: Proof expiration time
     defined_by: https://www.w3.org/TR/vc-data-integrity/#dfn-expires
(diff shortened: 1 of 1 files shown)


OR13 (approved):

OR13:
See also the comments on https://github.com/w3c/vc-data-integrity/pull/135

Wind4Greg (approved):

dlongley (approved):

vocab/security/vocabulary.yml
```diff
@@ -235,6 +235,19 @@ property:
       - label: ed25519-2020
         url: https://w3c-ccg.github.io/lds-ed25519-2020/
 
+  - id: secretKeyMultibase
+    label: Secret key multibase
+    domain: sec:Multikey
+    range: xsd:string
```
msporny, line 257:
```suggestion
    range: sec:multibase
```

msporny (changes requested):
Minor range fix, otherwise LGTM.

vocab/security/vocabulary.yml
```diff
@@ -235,6 +235,19 @@ property:
       - label: ed25519-2020
         url: https://w3c-ccg.github.io/lds-ed25519-2020/
 
+  - id: secretKeyMultibase
+    label: Secret key multibase
+    domain: sec:Multikey
+    range: sec:multibase
```
iherman: The `sec:multibase` class does not exist in the current vocabulary (unless I miss it in the maze of PR-s). Note that, for a class, it should be `sec:Multibase` (ie, capital 'M')

iherman (changes requested):

vocab/security/vocabulary.yml
OR13: https://github.com/w3c/vc-data-model/blob/main/contexts/credentials/v2#L245

^ related (and in the vocabulary already).

I agree regarding the capitalization

dlongley: `sec:multibase` isn't a class, it's a datatype, like `xsd:string`. This needs the same consideration as `sec:cryptosuiteString`.

dmitrizagidulin: @iherman @dlongley so, do I need to do anything in this PR, with respect to `sec:multibase` data type?

OR13: > `sec:multibase` isn't a class, it's a datatype, like `xsd:string`. This needs the same consideration as `sec:cryptosuiteString`.

Then the capitalization makes sense as is, here is another reference to data types in the v2 context for comparison:

- https://github.com/w3c/vc-data-model/blob/main/contexts/credentials/v2#L62

iherman: > `sec:multibase` isn't a class, it's a datatype, like `xsd:string`. This needs the same consideration as `sec:cryptosuiteString`.

Ah! O.k., understood; the missing capitalization is fine then. However, just as for cryptosuite, the datatype must be properly defined in the spec and then the vocabulary yml file must be updated.

I would propose that this PR would include the datatype definitions in the spec and when all these types of PRs are merged I will take care of incorporating the datatype definitions into the vocabulary specification itself in a separate PR.

msporny:
Normative, multiple reviews, changes requested and made, objections clarified until there were no objections remaining, merging.

Issue #78, opened 2023-01-14
Expiration Date


David-Chadwick:
All cryptographic proofs should have an expiration date since no cryptography lasts forever. The current specification includes the `created` property but has no equivalent `expires` property.

dlongley:
The context does provide an `expires` term here: https://github.com/w3c/vc-data-integrity/blob/main/contexts/data-integrity/v1#L23-L26

However, it does look like we don't have examples / text in the spec around its usage.

David-Chadwick:
What is the difference between the created and expires properties of the proof and the validFrom and validTo properties of the VC?

Wind4Greg:
My take is that `created` and `expires` are more associated with the cryptographic properties of the proof and would take into account key material expiration. In addition a VC may contain multiple proof via proof set or proof chain and hence it could be useful to differentiate.

David-Chadwick:
So what does it mean if the proof is not valid (i.e. it is outside its validity period) but the verifiableCredential is valid (i.e. it is inside its validity period)? How can the verifier believe the validity period of the verifiableCredential?

msporny:
> So what does it mean if the proof is not valid (i.e. it is outside its validity period) but the verifiableCredential is valid (i.e. it is inside its validity period)? 

By default, the signature will fail to verify, which means the VC will fail to verify.

> How can the verifier believe the validity period of the verifiableCredential?

Answer above... proof expiration needs to be checked (if it's specified), then validity period for VC. These are business rules, of course, so the verifier could choose to ignore expired proofs.

msporny:
PR #124 has been merged to address this issue. Marking as pending close. @David-Chadwick let us know if you require additional changes.

David-Chadwick:
Where is the text that relates `expires` to validUntil? I could not find anything in the current text that does this.
Note also that `created `should also be related to validFrom.
It makes no sense to me to have different values for these two sets of validity times, given that the VC valid properties are for the VC and not for the unsigned credential.
OTOH if the validFrom and validUntil were properties of the underlying credential then it makes perfect sense for them to be different to the signature `expires` and `created` properties

msporny:
@David-Chadwick wrote:
> OTOH if the validFrom and validUntil were properties of the underlying credential then it makes perfect sense for them to be different to the signature expires and created properties

[...]

PR #155, opened 2023-08-06
Add text on difference between proof and VC validity periods.

msporny:
This PR attempts to address issue #78 by adding text to the "Relationship to VCs" section on the difference between proof and VC validity periods.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/155.html" title="Last updated on Aug 17, 2023, 6:09 PM UTC (0c4701f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/155/73bf77e...0c4701f.html" title="Last updated on Aug 17, 2023, 6:09 PM UTC (0c4701f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1543,10 +1543,10 @@ href="#dfn-expires">`expires`</a> properties, and the validity period of a
 <a data-cite="?VC-DATA-MODEL-2.0#defn-validUntil">`validUntil`</a> properties.
 While these properties might sometimes express the same validity periods, at
 other times they might not be aligned. When verifying a
-<a href="#proofs">proof</a>, it is important to ensure that the time is within
-the validity period for the proof. When
-<a data-cite="?VC-DATA-MODEL-2.0#validation">validating</a>
-a <a>verifiable credential</a>, it is important to ensure that the time is
+<a href="#proofs">proof</a>, it is important to ensure that the time of interest
+(the current time or another time) is within the validity period for the proof.
+When <a data-cite="?VC-DATA-MODEL-2.0#validation">validating</a> a
+<a>verifiable credential</a>, it is important to ensure that the time of interest is
 within the validity period for the <a>verifiable credential</a>. Note that a
 failure to validate either the validity period for the <a
 href="#proofs">proof</a>, or the validity period for the <a>verifiable
--- a/index.html
+++ b/index.html
@@ -1500,6 +1500,7 @@ this specification.
 
       <section>
         <h2>Relationship to Verifiable Credentials</h2>
+
         <p>
 Cryptographic suites that implement this specification can be used to secure
 <a data-cite="?VC-DATA-MODEL-2.0#dfn-verifiable-credential">
@@ -1507,8 +1508,11 @@ verifiable credentials</a> and
 <a data-cite="?VC-DATA-MODEL-2.0#dfn-verifiable-presentation">
 verifiable presentations</a>. Implementers
 that are addressing those use cases are cautioned that additional checks might
-be appropriate when processing those types of documents. For example, there are
-some use cases where it is important to ensure that the
+be appropriate when processing those types of documents.
+        </p>
+
+        <p>
+There are some use cases where it is important to ensure that the
 <a>verification method</a> used in a proof is associated with the
 <a data-cite="?VC-DATA-MODEL-2.0#issuer">`issuer`</a> in a
 <a data-cite="?VC-DATA-MODEL-2.0#dfn-verifiable-credential">
@@ -1527,6 +1531,26 @@ This particular association indicates that the
 <a data-cite="?VC-DATA-MODEL-2.0#dfn-holder">`holder`</a>, respectively,
 is the controller of the <a>verification method</a> used to verify
 the proof.
+        </p>
+
+        <p>
+Document authors and implementers are advised to understand the difference
+between the validity period of a <a href="#proofs">proof</a>, which is expressed
+using the <a href="#dfn-created">`created`</a> and <a
+href="#dfn-expires">`expires`</a> properties, and the validity period of a
+<a>verifiable credential</a>, which is expressed using the
+<a data-cite="?VC-DATA-MODEL-2.0#defn-validFrom">`validFrom`</a> and
+<a data-cite="?VC-DATA-MODEL-2.0#defn-validUntil">`validUntil`</a> properties.
+While these properties might sometimes express the same validity periods, at
+other times they might not be aligned. When verifying a
+<a href="#proofs">proof</a>, it is important to ensure that the time is within
+the validity period for the proof. When
+<a data-cite="?VC-DATA-MODEL-2.0#validation">validating</a>
+a <a>verifiable credential</a>, it is important to ensure that the time is
+within the validity period for the <a>verifiable credential</a>. Note that a
+failure to validate either the validity period for the <a
+href="#proofs">proof</a>, or the validity period for the <a>verifiable
+credential</a> might result in accepting data that ought to have been rejected.
         </p>
       </section>
 
--- a/index.html
+++ b/index.html
@@ -1550,7 +1550,7 @@ When <a data-cite="?VC-DATA-MODEL-2.0#validation">validating</a> a
 within the validity period for the <a>verifiable credential</a>. Note that a
 failure to validate either the validity period for the <a
 href="#proofs">proof</a>, or the validity period for the <a>verifiable
-credential</a> might result in accepting data that ought to have been rejected.
+credential</a>, might result in accepting data that ought to have been rejected.
         </p>
       </section>
 
--- a/index.html
+++ b/index.html
@@ -1538,19 +1538,27 @@ Document authors and implementers are advised to understand the difference
 between the validity period of a <a href="#proofs">proof</a>, which is expressed
 using the <a href="#dfn-created">`created`</a> and <a
 href="#dfn-expires">`expires`</a> properties, and the validity period of a
-<a>verifiable credential</a>, which is expressed using the
+<a data-cite="?VC-DATA-MODEL-2.0#dfn-credential">credential</a>,
+which is expressed using the
 <a data-cite="?VC-DATA-MODEL-2.0#defn-validFrom">`validFrom`</a> and
 <a data-cite="?VC-DATA-MODEL-2.0#defn-validUntil">`validUntil`</a> properties.
 While these properties might sometimes express the same validity periods, at
 other times they might not be aligned. When verifying a
 <a href="#proofs">proof</a>, it is important to ensure that the time of interest
-(the current time or another time) is within the validity period for the proof.
+(which might be the current time or any other time) is within the
+validity period for the proof (that is, between
+<a href="#dfn-created">`created`</a> and <a href="#dfn-expires">`expires`</a> ).
 When <a data-cite="?VC-DATA-MODEL-2.0#validation">validating</a> a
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1528,6 +1532,26 @@ <h2>Relationship to Verifiable Credentials</h2>
...
+href="#dfn-expires">`expires`</a> properties, and the validity period of a
+<a>verifiable credential</a>, which is expressed using the
+<a data-cite="?VC-DATA-MODEL-2.0#defn-validFrom">`validFrom`</a> and
+<a data-cite="?VC-DATA-MODEL-2.0#defn-validUntil">`validUntil`</a> properties.
+While these properties might sometimes express the same validity periods, at
+other times they might not be aligned. When verifying a
+<a href="#proofs">proof</a>, it is important to ensure that the time is within
+the validity period for the proof. When
+<a data-cite="?VC-DATA-MODEL-2.0#validation">validating</a>
+a <a>verifiable credential</a>, it is important to ensure that the time is
```
dlongley: Trying to indicate that "the time" one is whatever is of interest, which may commonly be the current time, but does not have to be.

```suggestion
<a href="#proofs">proof</a>, it is important to ensure that the time of interest
(the current time or another time) is within the validity period for the proof.
When <a data-cite="?VC-DATA-MODEL-2.0#validation">validating</a> a
<a>verifiable credential</a>, it is important to ensure that the time of interest is
```

dlongley (approved):

Wind4Greg (approved):

iherman (approved):

David-Chadwick (changes requested):
Whilst this text is beneficial, it is not sufficient on its own since it raises the following question. When, or under what circumstances, will the validity periods be the same, and when or why will they be different?

index.html
```diff
@@ -1528,6 +1532,26 @@ <h2>Relationship to Verifiable Credentials</h2>
...
+While these properties might sometimes express the same validity periods, at
+other times they might not be aligned. When verifying a
+<a href="#proofs">proof</a>, it is important to ensure that the time of interest
+(the current time or another time) is within the validity period for the proof.
+When <a data-cite="?VC-DATA-MODEL-2.0#validation">validating</a> a
+<a>verifiable credential</a>, it is important to ensure that the time of interest is
+within the validity period for the <a>verifiable credential</a>. Note that a
+failure to validate either the validity period for the <a
+href="#proofs">proof</a>, or the validity period for the <a>verifiable
+credential</a> might result in accepting data that ought to have been rejected.
```
```suggestion
credential</a>, might result in accepting data that ought to have been rejected.
```

TallTed (changes requested):

brentzundel (approved):

TallTed:
> [@David-Chadwick] Whilst this text is beneficial, it is not sufficient on its own since it raises the following question. When, or under what circumstances, will the validity periods be the same, and when or why will they be different.

Do you have any answers to those questions (which aren't really questions, since the sentence lacks the key syntactical element, `?`)?

Perhaps you might offer some text that gets us closer to what you consider to be sufficient?

David-Chadwick:
Here is my suggestion
"Whilst `created` and `expires` refer to the validity period of the proof i.e. of the `verifiableCredential`, 'validFrom' and 'validUntil' from the VC Data Model refer to the validity period of the credential object."

index.html
```diff
@@ -1528,6 +1532,26 @@ <h2>Relationship to Verifiable Credentials</h2>
 is the controller of the <a>verification method</a> used to verify
 the proof.
         </p>
+
+        <p>
+Document authors and implementers are advised to understand the difference
+between the validity period of a <a href="#proofs">proof</a>, which is expressed
+using the <a href="#dfn-created">`created`</a> and <a
+href="#dfn-expires">`expires`</a> properties, and the validity period of a
+<a>verifiable credential</a>, which is expressed using the
```
David-Chadwick: replace <a>verifiable credential</a> with
 credential object

We have already agreed that there are 3 entities of interest:
1. the underlying physical certificate/qualification/credential
2. the credential object which is the JSON representation of 1.
3. the verifiable credential.
Thus the validFrom and validUntil properties will have been taken from the validity period of 1.

David-Chadwick (changes requested):

dmitrizagidulin (approved):
:+1:

index.html
TallTed: > [@David-Chadwick] We have already agreed that there are 3 entities of interest:

*You* have agreed upon this.

You have a consistent pattern of stating that things you have asserted have been agreed upon by the rest of the group. Such blind and repetitive statements _can_ be a powerful rhetorical tool, but they are _not_ a good way to proceed when trying to develop a globally acceptable and ratified standard.

> 1. the underlying physical certificate/qualification/credential

There is absolutely no requirement that your first listed item must exist. Such a physical document *might* exist but it equally might *not*.

As an issuer, my assertions may be made and placed directly in your second listed item, for protection/encapsulation by/in your third listed item.

> 2. the credential object which is the JSON representation of 1.

Not JSON, JSON-LD. We *have* all agreed on JSON-LD representation (i.e., `application/vc+ld+json`), as of the Miami F2F.

And, as noted above, your item 1 may not exist, so item 2 is a JSON-LD representation of some number of assertions, prepared for protection/encapsulation in/by item 3.

> 3. the verifiable credential.

Being item 2 plus proof attachment/insertion.

> Thus the `validFrom` and `validUntil` properties will have been taken from the validity period of 1.

Except that item 1 may not exist, so may not provide any validity period details.

David-Chadwick: I do not see how your statements above negate any of my original ones. The fact that there often is a physical certificate, and that this is an entity of interest, does not state that there is a requirement for it to always exist. The fact that JSON-LD is a subclass of JSON does not negate that fact that the credential is also JSON. Finally PR#1211 from the data model states that the validity period is optional, and if present, it is the validity period of the credential and not of the verifiableCredential.

msporny: @David-Chadwick, I believe what @TallTed is saying is the following:

> the underlying physical certificate/qualification/credential

This thing might not exist, so asserting that the validity period of the proof, or the validity period of the verifiable credential lines up with it (the physical credential) is not always true, and might become increasingly not true because some physical credentials might cease to be issued because their digital counterparts are easier to issue/manage.

So, tying any validity period to the physical certificate is fraught.

> the credential object which is the JSON representation of 1.

While this thing exists, the group tends to not care as much about it as the verifiable credential. So, stating that the `validFrom` and `validUntil` refers to the credential object isn't quite right either... because both of those items are optional and what makes them enforceable (in a meaningful way) is their expression on a verifiable credential.

> the verifiable credential. Thus the validFrom and validUntil properties will have been taken from the validity period of 1.

Not necessarily... which is the point. I think you're wanting to say something that we don't have consensus on.

The only thing this PR is attempting to highlight is that the validity period of the proof might not always line up with the validity period of the verifiable credential... and that's it. The validity period of the "credential object" (as you've coined the term) is a bit murky because the thing isn't secured... so is it really the validity period, or just something some attacker injected into the "credential object"?

I'd like to merge this PR... can you think of any language we could add that would address your concern AND not raise objections from others that have reviewed this PR?

David-Chadwick: Please take another look at  PR#1211 from the data model (which I think you authored). This states that the validity period is optional, and if present, it is the validity period of the credential. (You can ignore my addition of the word object to this, since most (all?) of our data structures can be regarded as objects). It does not say that it is the validity period of the verifiable credential, which you are stating above. So clearly there is a disconnect somewhere that needs fixing. This is my concern.

David-Chadwick:
The current PR text should be changed to the following in order to align with the VC DM PR#1211

Document authors and implementers are advised to understand the difference between the validity period of a `proof`, which is expressed using the `created` and `expires` properties, and the validity period of a credential , which is expressed using the `validFrom `and `validUntil` properties. While these properties might sometimes express the same validity periods, at other times they might not be aligned. When verifying a `proof`, it is important to ensure that the time of interest (the current time or another time) is within the validity period for the proof. When `validating`  a  verifiable credential , it is important to ensure that the time of interest is within the validity period for the  credential . Note that a failure to validate either the validity period for the `proof`, or the validity period for the  credential , might result in accepting data that ought to have been rejected.

index.html
David-Chadwick: @msporny said "The validity period of the "credential object" (as you've coined the term)". I did not coin this term. It is already in the VCDM quote "For example, a credential  object with the type of ExampleDegreeCredential ," :-)

msporny: @David-Chadwick wrote:
>  I did not coin this term. It is already in the VCDM quote "For example, a credential object with the type of ExampleDegreeCredential ,"

Hmm... I'm not seeing the term "credential object" here: https://github.com/w3c/vc-data-model/pull/1211/files

Ah!, but I do now see it in the main spec:

https://w3c.github.io/vc-data-model/

... once... and I believe that to be an error. So, you're right, the term /does/ exist in the main spec (and you didn't coin it), but since it only appears once, it's probably just a hold over from previous iterations of the spec. 

And, as you say... saying "object" is probably neither here nor there because many of the things in the spec that we describe could be viewed as "objects".

I /thought/ that you were describing a NEW kind of thing, that you were calling a "credential object", which was purely JSON (and unsecured). I've re-read your statements in this thread and now am thinking that you are asserting that "credential" as described in the main spec and "credential object" are synonymous. They are exactly the same thing. Thus, we can just simplify your statements above to "credential", as described by the specification... and the thing that differentiates a "credential" from a "verifiable credential" is the fact that the latter is secured in some way, and the former is not?

> This states that the validity period is optional, and if present, it is the validity period of the credential. (You can ignore my addition of the word object to this, since most (all?) of our data structures can be regarded as objects). It does not say that it is the validity period of the verifiable credential, which you are stating above. So clearly there is a disconnect somewhere that needs fixing. This is my concern.

I had to go and look at the main spec again, and yes, you are correct. If we change this language to align with the main spec, would that address your concern on this PR, @David-Chadwick ? That is, we change the mention of "validFrom" and "validUntil" relationship from "verifiable credential" to "credential" in this PR. Would that address your concern?

David-Chadwick: Yes and I produced the revised text for you below (2 days ago)

TallTed:
Massaging [the latest](https://github.com/w3c/vc-data-integrity/pull/155#issuecomment-1676093163) from @David-Chadwick --
```
Document authors and implementers are advised to understand the difference between
the validity period of a `proof`, which is expressed using the `created` and `expires`
properties, and the validity period of a credential, which is expressed using the
`validFrom` and `validUntil` properties. While these properties might sometimes express
the same validity period, at other times they might not be aligned. When verifying a
`proof`, it is important to ensure that the time of interest (which may be the current
time or any other time) is within the validity period for the `proof` (i.e., between
`created` and `expires`). When `validating` a verifiable credential, it is important
to ensure that the time of interest is within the validity period for the credential
(i.e., between `validFrom` and `validUntil`). Note that a failure to validate either
the validity period for the `proof`, or the validity period for the credential, might
result in accepting data that ought to have been rejected.
```

msporny:
@David-Chadwick if you are ok w/ @TallTed's modification of your text ([above](https://github.com/w3c/vc-data-integrity/pull/155#issuecomment-1678015151)), I'm happy to update this PR w/ it and merge it. I believe it contains the changes that you were requesting?

index.html
msporny: I see it now, ok, thx @David-Chadwick.

David-Chadwick:
@msporny Yes thanks, please go ahead

msporny:

Issue #159, opened 2023-08-06
Clearly define conformance classes


msporny:
The conformance classes in the specification are currently undefined. We need to define clear conformance classes that document authors and implementers can say that their software conforms to.

msporny:
PR #160 has been raised to address this issue. This issue will be closed once that PR has been merged.

msporny:
PR #160 has been merged, closing.

PR #160, opened 2023-08-06
Add conformance classes to specification.

msporny:
This PR attempts to address issue #159 by defining the conformance classes that are relevant to the specification.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/160.html" title="Last updated on Aug 12, 2023, 8:10 PM UTC (e39eacb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/160/73bf77e...e39eacb.html" title="Last updated on Aug 12, 2023, 8:10 PM UTC (e39eacb)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -425,22 +425,21 @@ A <dfn>conforming secured document</dfn> is any concrete expression of the
 data model that follows the relevant normative requirements in Sections
 <a href="#proofs"></a>, <a href="#proof-purposes"></a>,
 <a href="#contexts-and-vocabularies"></a>, and
-<a href="#dataintegrityproof"></a>. Specifically,
-all relevant normative statements in those sections MUST be enforced.
+<a href="#dataintegrityproof"></a>.
         </p>
 
         <p>
 A <dfn>conforming controller document</dfn> is any concrete expression of the
-data model that follows the relevant normative requirements in Section
-<a href="#controller-documents"></a>. Specifically, all relevant normative
-statements in that section this document MUST be enforced.
+data model that follows the relevant normative requirements in Sections
+<a href="#controller-documents"></a> and
+<a href="#contexts-and-vocabularies"></a>.
         </p>
 
         <p>
 A <dfn>conforming verification method</dfn> is any concrete expression of the
 data model that follows the relevant normative requirements in Sections
-<a href="#verification-methods"></a>. Specifically, all relevant normative
-statements in that section this document MUST be enforced.
+<a href="#verification-methods"></a> and
+<a href="#contexts-and-vocabularies"></a>.
         </p>
 
         <p>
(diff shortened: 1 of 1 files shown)


Wind4Greg (approved):

dlongley (approved):

index.html
```diff
@@ -420,6 +420,48 @@ <h3>Design Goals and Rationale</h3>
...
+
+        <p>
+A <dfn class="lint-ignore">conforming cryptographic suite specification</dfn> is
+any document that follows the relevant normative requirements in Section
+<a href="#cryptographic-suites"></a>.
+        </p>
+
+        <p>
+A <dfn class="lint-ignore">conforming processor</dfn> is any algorithm realized
+as software and/or hardware that generates or consumes a
```
```suggestion
as software and/or hardware that generates and/or consumes a
```

TallTed (changes requested):
Minor but important

iherman (commented):
I am bothered by the terminology used around "data models", like:

> A conforming secured document is any concrete expression of the data model that follows the relevant normative requirements in Sections 2.1 Proofs, 2.2 Proof Purposes, 2.6 Contexts and Vocabularies, and 3.1 DataIntegrityProof.

Reading the term "data model" may set the expectation of some normative words about _what_ our data model is for this document. However, by going to, say, [§2.1 Proofs](https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/160.html#proofs), there is "just" a set of properties defined for proofs, without any reference to any well-defined data model. The generic introduction to [§2 Data Model](https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/160.html#data-model):

> This section specifies the data model that is used for expressing data integrity proofs, controller documents, and verification methods.

is also void of any definition.

I realize, and I agree, that the document purposefully avoids making a reference to VCs or VPs at this point (the relationships are detailed in [§2.5](https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/160.html#relationship-to-verifiable-credentials)) because the DI can be used more generally (e.g., for DID documents). But I wonder whether it is not appropriate to say that the Data Model of DI relies on what is described as "Claims" in the VCDM specification in [VCDM §3.1 Claims](https://www.w3.org/TR/vc-data-model-2.0/#claims) (which is a simple version of subject-property-values relationships). Whether the DI spec would just simply refer to VCDM §3.1, or take over the same text with diagrams, is an editorial choice. (DRY principle would dictate to use a reference, on the other hand that section of VCDM is set as non-normative, which may not be appropriate here.)

dmitrizagidulin (approved):
:+1: seems like a decent starting point. I agree with Ivan, though, that just saying 'the data model' might be somewhat ambiguous.

msporny:
@iherman and @dmitrizagidulin wrote:
> I am bothered by the terminology used around "data models"

I have raised issue #165 to track this concern. I understand the trepidation, but also didn't want to create a big debate around what the data model is for data integrity. One could argue that it's a version of INFRA that maps cleanly to both JSON-LD and JSON... or, one could argue that it's JSON-LD with special rules if you don't want to use a JSON-LD processor and just operate using JSON tooling. Both of those statements are true, and what we're probably missing is general guidance around "Gaining the benefits of JSON-LD without using JSON-LD Processors."

How useful it is to go into these depths is questionable from an implementer standpoint... do they need an answer to this question to write an implementation that is conformant? No, I don't think they do. :)

Sure, it might be helpful to some implementers to know that the mental model (data model) sticks together... but trying to convey that in the conformance section, or having to repeat it in every specification that uses JSON-LD is probably not helpful. This should probably become a section in the JSON-LD specification (the fact that you can gain many of the benefits of JSON-LD w/o doing full blown JSON-LD (or RDF) processing.

In any case, I don't know what to write/change and need some help doing so. Let's discuss in issue #165 and, if we have a flash of clarity, then we can raise another PR to modify this section. In the meantime, we don't have /any/ conformance classes, so I'm going to merge this PR (since no one is objecting, though there is uneasiness around the "data model"). We can continue the discussion in #165.

msporny:
Normative, multiple reviews, changes requested and made, #165 raised to track remaining concerns, no objections, merging.

Issue #161, opened 2023-08-06
Add sec:cryptosuiteString data type to vocabulary


msporny:
It seems like we have consensus on the naming pattern that will be used for cryptosuites: https://github.com/w3c/vc-di-ecdsa/pull/23

The one downside for this naming pattern is the number of characters that are used (around of 15 characters) and how they interact w/ semantic compression mechanisms like CBOR-LD (badly, not easy to compress non-repeating text strings). This is a problem for small payload sizes, such as VCs sent over NFC and QR Codes, where the byte payload has to stay under around 395 bytes for a QR Code payload to be read on an average retail optical scanner.

Adding a `sec:cryptosuite` datatype will allow us to compress the 15 characters down to single byte values for the cryptosuite type. This issue is to track adding the data type to the specification and the vocabulary.

msporny:
PR #162 has been raised to address this issue. This issue will be closed if PR #162 is merged.

msporny:
PR #162 has been merged, closing.

PR #162, opened 2023-08-07
Add `cryptosuiteString` subtype for the `cryptosuite` property.

msporny:
This PR is an attempt to address issue #161 by adding a `cryptosuiteString` subtype for the `cryptosuite` property to ensure that it is possible to compress cryptosuite strings when performing semantic compression (CBOR-LD).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/162.html" title="Last updated on Aug 17, 2023, 6:39 PM UTC (e1ed114)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/162/3b3e64a...e1ed114.html" title="Last updated on Aug 17, 2023, 6:39 PM UTC (e1ed114)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1905,11 +1905,11 @@ such as when an undefined term is detected in an input document.
           <h3 id="cryptosuiteString">The `cryptosuiteString` Datatype</h3>
 
           <p>
-This specification encodes cryptographic identifiers as enumerable strings,
-which is useful in processes that need to efficiently encode such strings,
-such as compression algorithms. In environments that support data types for
-string values, such as RDF [[?RDF-CONCEPTS]], cryptographic identifier content
-is indicated using a literal value whose datatype is set to
+This specification encodes cryptographic suite identifiers as enumerable
+strings, which is useful in processes that need to efficiently encode such
+strings, such as compression algorithms. In environments that support data types
+for string values, such as RDF [[?RDF-CONCEPTS]], cryptographic identifier
+content is indicated using a literal value whose datatype is set to
 `https://w3id.org/security#cryptosuiteString`.
           </p>
 
@@ -1920,13 +1920,13 @@ The <code>cryptosuiteString</code> datatype is defined as follows:
           <dl>
             <dt>The URL denoting this datatype</dt>
             <dd>
-`https://w3id.org/security#cryptosuiteString`.
+`https://w3id.org/security#cryptosuiteString`
             </dd>
             <dt>The lexical space</dt>
             <dd>
 The set of all cryptosuite types, expressed using American Standard Code
-for Information Interchange [[ASCII]] strings that are defined as the set of all
-Data Integrity cryptosuite specifications.
+for Information Interchange [[ASCII]] strings, that are defined by the
+collection of all Data Integrity cryptosuite specifications.
             </dd>
             <dt>The value space</dt>
             <dd>
@@ -1936,12 +1936,13 @@ property, as defined in Section <a href="#dataintegrityproof"></a>.
             <dt>The lexical-to-value mapping</dt>
             <dd>
 Any element of the lexical space is mapped to the result of parsing it into an
-internal representation that uniquely identifies the cryptosuite value from all
-other possible cryptosuite values.
+internal representation that uniquely identifies the cryptosuite type from all
+other possible cryptosuite types.
             </dd>
             <dt>The canonical mapping</dt>
             <dd>
-Any element of the value space is mapped to the corresponding string in the lexical space.
+Any element of the value space is mapped to the corresponding string in the
+lexical space.
             </dd>
           </dl>
         </section>
--- a/index.html
+++ b/index.html
@@ -1924,13 +1924,13 @@ The <code>cryptosuiteString</code> datatype is defined as follows:
             </dd>
             <dt>The lexical space</dt>
             <dd>
-The set of all cryptosuite types, expressed using American Standard Code
+The union of all cryptosuite strings, expressed using American Standard Code
 for Information Interchange [[ASCII]] strings, that are defined by the
 collection of all Data Integrity cryptosuite specifications.
             </dd>
             <dt>The value space</dt>
             <dd>
-The set of all cryptosuite types that are expressed using the `cryptosuite`
+The union of all cryptosuite types that are expressed using the `cryptosuite`
 property, as defined in Section <a href="#dataintegrityproof"></a>.
             </dd>
             <dt>The lexical-to-value mapping</dt>
--- a/index.html
+++ b/index.html
@@ -2006,8 +2006,8 @@ The `type` property MUST contain the string `DataIntegrityProof`.
           <dt>cryptosuite</dt>
           <dd>
 The `cryptosuite` value MUST specify a string that identifies the
-<a>cryptographic suite</a>. If the processing environment supports string
-subtypes, the string subtype MUST be
+<a>cryptographic suite</a>. If the processing environment supports subtypes
+of `string`, the type of the `cryptosuite` value MUST be the `string` subtype
 `https://w3id.org/security#cryptosuiteString`.
           </dd>
           <dt>proofValue</dt>
--- a/index.html
+++ b/index.html
@@ -2005,8 +2005,10 @@ The `type` property MUST contain the string `DataIntegrityProof`.
           </dd>
           <dt>cryptosuite</dt>
           <dd>
-The `cryptosuite` property MUST contain a string specifying the name of the
-cryptosuite.
+The `cryptosuite` value MUST specify a string that identifies the
+<a>cryptographic suite</a>. If the processing environment supports string
+subtypes, the string subtype MUST be
+`https://w3id.org/security#cryptosuiteString`.
           </dd>
           <dt>proofValue</dt>
           <dd>
--- a/index.html
+++ b/index.html
@@ -1906,8 +1906,8 @@ such as when an undefined term is detected in an input document.
 
           <p>
 This specification encodes cryptographic identifiers as enumerable strings,
-which is useful in processes, such as compression algorithms, that need to
-efficiently encode such strings. In environments that support data types for
+which is useful in processes that need to efficiently encode such strings,
+such as compression algorithms. In environments that support data types for
 string values, such as RDF [[?RDF-CONCEPTS]], cryptographic identifier content
 is indicated using a literal value whose datatype is set to
 `https://w3id.org/security#cryptosuiteString`.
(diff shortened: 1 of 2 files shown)


vocab/security/vocabulary.yml
```diff
@@ -235,7 +235,7 @@ property:
   - id: cryptosuite
     label: Cryptographic suite
     domain: sec:DataIntegrityProof
-    range: xsd:string
+    range: sec:cryptosuiteString
```
iherman, line 243: That is not correct in the vocabulary: the value of `range` must be a class or a new datatype. Do we really want to go down the (tedious!) process of formally define a new datatype for `cryptosuiteString`?

I would propose to leave the range a string, and add a comment (or part of the `cryptosuite` property definition in the spec) what the constraints on the value are. But do not treat it as a datatype with a separate URL.

iherman (changes requested):

vocab/security/vocabulary.yml
iherman: Just to add to my previous remark. If the WG _does_ decide to define a new datatype, what I would think we should do is as follows.

- I do not think we should bother extending the yml2vocab script to add a datatype definition feature in general; this is a very rare occurrence, and this tool is not aimed at being a universal RDF tool. Also, the datatype definition is _not_ part of the RDF Vocabulary Graph, in other words, it is not something we would add to the Turtle and JSON-LD files, only the HTML version. As a consequence, what I would propose to do is to add the formal definition to the integrity spec (see below), and add a separate section into the Template html file that refers back to the definition in the integrity spec, a bit like all the others. I am happy to do that, once the references in the integrity spec is available.
- The formal definition of a datatype requires the formal definition of the lexical space, value space, etc. Even for a type that is relatively simple it is, as I said, tedious. I think the formal definition of the [rdf:JSON Datatype](https://www.w3.org/TR/json-ld11/#the-rdf-json-datatype) is a good example for what should be added to the integrity spec (probably in a normative appendix, to avoid breaking the flow of the main text).

msporny: >  Do we really want to go down the (tedious!) process of formally define a new datatype for cryptosuiteString?

Yes, we do, because if we don't do that, the resolution of using long cryptosuite names falls apart due to the burden it places on low-bandwidth environments. In other words, the pain felt in defining a new type will be on the Editors. If we don't define this type, the pain felt will be on every implementer out there in a low bandwidth environment. I'll volunteer to define the data type, thanks for the pointer to rdf:JSON... we'll model it after that.

iherman: > Yes, we do, because if we don't do that, the resolution of using long cryptosuite names falls apart due to the burden it places on low-bandwidth environments. In other words, the pain felt in defining a new type will be on the Editors. If we don't define this type, the pain felt will be on every implementer out there in a low bandwidth environment. I'll volunteer to define the data type, thanks for the pointer to rdf:JSON... we'll model it after that.

Fine. Then let us do it, but we should follow something like https://github.com/w3c/vc-data-integrity/pull/162#discussion_r1285419464 to make it proper. It is tedious but not too much after all; the only important point is to define more formally the subset of strings that can be used as a datatype value. The rest is obvious (and the rdf:JSON stuff could indeed be used as a pattern).

I am o.k. to expand the formal vocabulary with the datatype stuff once the core spec has it, either in this PR or (probably better) in another one.

Wind4Greg (approved):

dlongley (approved):

index.html
```diff
@@ -1819,8 +1819,10 @@ <h3>DataIntegrityProof</h3>
           </dd>
           <dt>cryptosuite</dt>
           <dd>
-The `cryptosuite` property MUST contain a string specifying the name of the
-cryptosuite.
+The `cryptosuite` value MUST specify a string that identifies the
+<a>cryptographic suite</a>. If the processing environment supports string
+subtypes, the string subtype MUST be
```
```suggestion
<a>cryptographic suite</a>. If the processing environment supports subtypes
of `string`, the type of the `cryptosuite` value MUST be the `string` subtype
```

TallTed (changes requested):

dmitrizagidulin (approved):

vocab/security/vocabulary.yml
msporny: @iherman I have added the definition of cryptosuiteString, based on the link to provided for rdf:JSON, in this commit: https://github.com/w3c/vc-data-integrity/pull/162/commits/40be121985806f03bc07dbef7a253e91b587d737

If you are ok with that section (or a variation of it), I can do the same for the `multibase` datatype in another PR.

Requesting a re-review from you now.

msporny: Requesting a re-review from @TallTed as well because I added a bunch of text that's hard to read and I need his help to make it suck less. :)

index.html
```diff
@@ -1901,6 +1901,51 @@ <h3>Securing Data Losslessly</h3>
...
+
+          <dl>
+            <dt>The URL denoting this datatype</dt>
+            <dd>
+is `https://w3id.org/security#cryptosuiteString`.
+            </dd>
+            <dt>The lexical space</dt>
+            <dd>
+is the set of all cryptosuite types expressed using American Standard Code
+for Information Interchange [[ASCII]] strings that are defined by the sum of all
```
iherman: I wonder whether term "union" is not better than "sum" ("union" in the sense of a Union datatype in programming languages).

iherman (approved):
- Purely editorial: I have the strange feeling of a circular specification. The `cryptosuiteString` definition refers to the `dataIntegrityProof` section, which then refers to the `cryptosuiteString` section... And I did not see anywhere a clear specification (or reference to it) of the fact that cryptosuites have an identifier in terms of an ASCII string, and those are the ones we are talking about. At present, this datatype looks much more mysterious than what it is... But I may have missed some details.
- For the records and to other reviewers: the `cryptosuiteString` datatype definition will be added to the vocabulary in a separate PR once this one settles. The reason for the delay is that the ability to add a datatype specification required an extension of the underlying `yml2vocab` tool and, although that extension is already done, it has not yet uploaded to the npm repository.

index.html
```diff
@@ -1901,6 +1901,51 @@ <h3>Securing Data Losslessly</h3>
           </p>
         </section>
 
+        <section>
+          <h3 id="cryptosuiteString">The `cryptosuiteString` Datatype</h3>
+
+          <p>
+This specification encodes cryptographic identifiers as enumerable strings,
+which is useful in processes, such as compression algorithms, that need to
+efficiently encode such strings. In environments that support data types for
```
```suggestion
which is useful in processes that need to efficiently encode such strings,
such as compression algorithms. In environments that support data types for
```

index.html
```diff
@@ -1901,6 +1901,51 @@ <h3>Securing Data Losslessly</h3>
...
+          </p>
+
+          <p>
+The <code>cryptosuiteString</code> datatype is defined as follows:
+          </p>
+
+          <dl>
+            <dt>The URL denoting this datatype</dt>
+            <dd>
+is `https://w3id.org/security#cryptosuiteString`.
```
```suggestion
`https://w3id.org/security#cryptosuiteString`.
```

index.html
```diff
@@ -1901,6 +1901,51 @@ <h3>Securing Data Losslessly</h3>
...
+            </dd>
+            <dt>The lexical space</dt>
+            <dd>
+is the set of all cryptosuite types expressed using American Standard Code
+for Information Interchange [[ASCII]] strings that are defined by the sum of all
+Data Integrity cryptosuite specifications.
```
```suggestion
The set of all cryptosuite types, expressed using American Standard Code
for Information Interchange [[ASCII]] strings that are defined as the set of all
Data Integrity cryptosuite specifications.
```

index.html
```diff
@@ -1901,6 +1901,51 @@ <h3>Securing Data Losslessly</h3>
...
+            </dd>
+            <dt>The value space</dt>
+            <dd>
+is the set of all cryptosuite types that are expressed using the `cryptosuite`
+property as defined in Section <a href="#dataintegrityproof"></a>.
```
```suggestion
The set of all cryptosuite types that are expressed using the `cryptosuite`
property, as defined in Section <a href="#dataintegrityproof"></a>.
```

index.html
```diff
@@ -1901,6 +1901,51 @@ <h3>Securing Data Losslessly</h3>
...
+            </dd>
+            <dt>The lexical-to-value mapping</dt>
+            <dd>
+maps any element of the lexical space to the result of parsing it into an
+internal representation that uniquely identifies the cryptosuite value from all
+other possible cryptosuite values.
```
```suggestion
Any element of the lexical space is mapped to the result of parsing it into an
internal representation that uniquely identifies the cryptosuite value from all
other possible cryptosuite values.
```

index.html
```diff
@@ -1901,6 +1901,51 @@ <h3>Securing Data Losslessly</h3>
...
+            </dd>
+            <dt>The canonical mapping</dt>
+            <dd>
+maps any element of the value space to the corresponding string in the lexical space.
```
```suggestion
Any element of the value space is mapped to the corresponding string in the lexical space.
```

index.html
```diff
@@ -2005,8 +2050,10 @@ <h3>DataIntegrityProof</h3>
           </dd>
           <dt>cryptosuite</dt>
           <dd>
-The `cryptosuite` property MUST contain a string specifying the name of the
-cryptosuite.
+The `cryptosuite` value MUST specify a string that identifies the
+<a>cryptographic suite</a>. If the processing environment supports subtypes
+of `string`, the type of the `cryptosuite` value MUST be the `string` subtype
+`https://w3id.org/security#cryptosuiteString`.
```
```suggestion
The value of the `cryptosuite` property MUST be a string that identifies the
<a>cryptographic suite</a>. If the processing environment supports subtypes
of `string`, the type of the `cryptosuite` value MUST be the
`https://w3id.org/security#cryptosuiteString` subtype of `string`.
```

vocab/security/vocabulary.yml
TallTed: Re-reviewed. Not too horrible. Editorial changes requested.

TallTed (changes requested):

index.html
```suggestion
for Information Interchange [[ASCII]] strings that are defined by the union of all
```

msporny: Yes, "union" is clearly better, applied in a commit below.

msporny: Addressed @iherman's concern in e1ed114a6e41dbac6bd5c8c26b94acbe230cc241.

msporny:

PR #166, opened 2023-08-14
Fix basic requirements of `proof`

TallTed:
***
<a href="https://pr-preview.s3.amazonaws.com/TallTed/vc-data-integrity/pull/166.html" title="Last updated on Aug 14, 2023, 7:41 PM UTC (70813d2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/166/6c94d0b...TallTed:70813d2.html" title="Last updated on Aug 14, 2023, 7:41 PM UTC (70813d2)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -532,9 +532,9 @@ information is provided using Linked Data vocabularies such as the
 
         <p>
 When expressing a <a>data integrity proof</a> on an object, a
-<dfn class="lint-ignore">`proof`</dfn> property MUST be used where its value is a
-single object, or an unordered set of objects, expressed using the properties
-below:
+<dfn class="lint-ignore">`proof`</dfn> property MUST be used. If present, its
+value MUST be either a single object, or an unordered set of objects, expressed
+using the properties below:
         </p>
 
         <dl style="margin-left: 1em;">


Wind4Greg (approved):

index.html
```diff
@@ -532,9 +532,9 @@ <h3>Proofs</h3>
 
         <p>
 When expressing a <a>data integrity proof</a> on an object, a
-<dfn class="lint-ignore">`proof`</dfn> property MUST be used where its value is a
-single object, or an unordered set of objects, expressed using the properties
-below:
+<dfn class="lint-ignore">`proof`</dfn> property MUST be used. If present, its
+value MUST be either a single object, or an unordered set of objects, expressed
```
msporny, line 536:
```suggestion
<dfn class="lint-ignore">`proof`</dfn> property MUST be used and its
value MUST be either a single object, or an unordered set of objects, expressed
```

It's a bit strange to say "If present" for a property that MUST be present? Minor rewording suggestions.

dlongley (approved):
Approving w/@msporny's suggestion.

index.html
TallTed: @msporny @dlongley -- Reading the whole paragraph, it's only a MUST "When expressing a <a>data integrity proof</a> on an object", so "if present" seems OK.

I think the "If present" phrasing covers any time the `proof` property is present, while the "and" phrasing only covers the `proof` property "When expressing a <a>data integrity proof</a> on an object".

Some additional rephrasing, possibly including other segments of the doc, may be needed to make the restrictions on the `proof` property clearer.

msporny: If you feel strongly about this, I'm fine w/ leaving it as is. I don't think the new wording I suggested changes the meaning of the sentence.

TallTed: I prefer as-is, because this doesn't spark the need for further review/rewording.

msporny (approved):

msporny (approved):

PR #178, opened 2023-08-24
Align `datetimes` to `dateTimeStamp` values to ensure timezone.

msporny:
This PR attempts to align the way datetime values are used in this specification with the way they are used in the vc-data-model, namely, ensuring that a `dateTimeStamp` value is used (which ensures that a timezone is specified when expressing a datetime). 

VC v1.0 and v1.1 didn't do this and there were a few instances where systems were creating datetime values that did not include a timezone (and thus were ambiguous because interpreting the time was done using the local timezone... and if a datetime was set in one timezone, there could be a misinterpretation of the datetime value when interpreted in another timezone).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/178.html" title="Last updated on Aug 24, 2023, 8:58 PM UTC (8fd52e2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/178/4a82497...8fd52e2.html" title="Last updated on Aug 24, 2023, 8:58 PM UTC (8fd52e2)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -579,13 +579,13 @@ by a verifier during the verification process.
           <dt><dfn class="lint-ignore">created</dfn></dt>
           <dd>
 The date and time the proof was created is OPTIONAL and, if included, MUST be
-specified as an [[XMLSCHEMA11-2]] combined date and time string.
+specified as an [[XMLSCHEMA11-2]] `dateTimeStamp` string.
           </dd>
 
           <dt><dfn class="lint-ignore">expires</dfn></dt>
           <dd>
 An OPTIONAL property that conveys the date and time that a proof expires and that, if present, MUST be
-specified as an [[XMLSCHEMA11-2]] combined date and time string.
+specified as an [[XMLSCHEMA11-2]] `dateTimeStamp` string.
           </dd>
 
           <dt id="defn-domain">domain</dt>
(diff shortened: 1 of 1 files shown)


dlongley (approved):

brianorwhatever:
Didn't VCDM 1 also specify that the milliseconds couldn't be included as it adds a tracking vector? Is that still the case or did I make this up entirely?

iherman (approved):

Wind4Greg (approved):

msporny:
> Didn't VCDM 1 also specify that the milliseconds couldn't be included as it adds a tracking vector? Is that still the case or did I make this up entirely?

I vaguely remember something saying that somewhere :) -- so, either it's a shared delusion, or the text exists/existed somewhere. The signature (if it's not an unlinkable one) is more of a tracking vector, so I don't think this is a concern.

We did say that second precision was preferred because it allows good compression in CBOR-LD. The second you add fractions of a second, your byte storage requirements go up a bit... but again, we don't want to prevent people from being able to specify fractions of a second.

IOW, I think we're ok here.

msporny:
Normative, multiple reviews, no changes requested, no objections, merging.

PR #181, opened 2023-08-25
Add cryptosuiteString to v2 context, preserving older v1 context.

msporny:
This PR does the final work to address issue #161 -- since `cryptosuiteString` is a breaking change, and since we have organizations that have deployed the v1 context, we're updating to a v2 context in order to not break implementations in the field.

### DIFF
--- a/contexts/data-integrity/v2
+++ b/contexts/data-integrity/v2
@@ -1,0 +1,77 @@
+{
+  "@context": {
+    "id": "@id",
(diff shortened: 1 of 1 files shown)


dlongley (approved):

Wind4Greg (approved):

msporny:
Normative, multiple reviews, no changes requested, no objections, merging.

Issue #182, opened 2023-08-26
Adjustments to the `revoked` property


OR13:
Multikey defines `revoked` but `JsonWebKey` does not.

```
"revoked": {
  "@id": "https://w3id.org/security#revoked",
  "@type": "http://www.w3.org/2001/XMLSchema#dateTime"
},
```

On https://github.com/w3c/vc-jose-cose/pull/144#discussion_r1305826152 @TallTed asks if this field should include TZ.

If we make such a change, we would want to ensure it applies to all Verification Method classes, and was documented the same in both specs.

msporny:
> we would want to ensure it applies to all Verification Method classes, and was documented the same in both specs.

Agree.

For the time being, I'll just bring both VM classes inline w/ each other.

msporny:
PR #184 has been raised to address this issue. This issue will be closed once PR #184 is merged.

msporny:
PR #184 has been merged, closing.

PR #184, opened 2023-08-26
Add `revoked` and `expires` properties to JsonWebKey context.

msporny:
This PR attempts to address issue #182 by adds the `revoked` and `expires` properties to JsonWebKey context. It also differentiates the anchors for `expires` on `proof` and `verificationMethod`.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/184.html" title="Last updated on Sep 2, 2023, 2:57 PM UTC (c4aede2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/184/0a64834...c4aede2.html" title="Last updated on Sep 2, 2023, 2:57 PM UTC (c4aede2)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -996,13 +996,28 @@ Registries [TBD -- DIS-REGISTRIES].
 The value of the `controller` property MUST be a <a
 data-cite="INFRA#string">string</a> that conforms to the [[URL]] syntax.
                 </dd>
+                <dt><dfn class="lint-ignore">expires</dfn></dt>
+                <dd>
+The `expires` property is OPTIONAL. It is set, in advance, by the
+<a>controller</a> of a  <a>verification method</a> to signal when that method
+can no longer be used for verification purposes. If provided, it MUST be an
+[[XMLSCHEMA11-2]] `dateTimeStamp` string specifying when the
+<a>verification method</a> SHOULD cease to be used. Once the value is set, it is
+not expected to be updated, and systems depending on the value are expected to
+not verify any proofs associated with the <a>verification method</a> at or after
+the time of expiration.
+                </dd>
                 <dt><dfn class="lint-ignore">revoked</dfn></dt>
                 <dd>
-The `revoked` property is OPTIONAL. If provided, it MUST be an [[XMLSCHEMA11-2]]
+The `revoked` property is OPTIONAL. It is set by the <a>controller</a> of a
+<a>verification method</a> to signal when that method is to no longer to be used
+for verification purposes, such as after a security compromise of the
+<a>verification method</a>. If provided, it MUST be an [[XMLSCHEMA11-2]]
 `dateTimeStamp` string specifying when the <a>verification method</a>
-SHOULD cease to be used. Once the value is set, it is not expected to be updated, and
-systems depending on the value are expected to not verify any proofs associated
-with the <a>verification method</a> at or after the time of revocation.
+SHOULD cease to be used. Once the value is set, it is not expected to be
+updated, and systems depending on the value are expected to not verify any
+proofs associated with the <a>verification method</a> at or after the time of
+revocation.
                 </dd>
               </dl>
             </dd>
--- a/index.html
+++ b/index.html
@@ -1949,6 +1949,23 @@ href="#proofs">proof</a>, or the validity period for the
 <a data-cite="?VC-DATA-MODEL-2.0#dfn-credential">credential</a>, might result
 in accepting data that ought to have been rejected.
         </p>
+
+        <p>
+Finally, implementers are also urged to understand that there is a difference
+between the <a href="#dfn-revoked">revocation time</a> and
+<a href="#defn-vm-expires">expiration time</a> for a <a>verification method</a>,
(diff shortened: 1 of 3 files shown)


Wind4Greg (approved):

iherman (approved):

dlongley (approved):

OR13 (approved):

mkhraisha (approved):

index.html
```diff
@@ -995,14 +996,29 @@ <h2>Verification Methods</h2>
...
-The `revoked` property is OPTIONAL. If provided, it MUST be an [[XMLSCHEMA11-2]]
+The `revoked` property is OPTIONAL. It is set by the <a>controller</a> of a
+<a>verification method</a> to signal when that method is to no longer to be used
+for verification purposes, such as after a security compromise of the
+<a>verification method</a>. If provided, it MUST be an [[XMLSCHEMA11-2]]
 combined date and time string specifying when the <a>verification method</a>
-SHOULD cease to be used. Once the value is set, it is not expected to be updated, and
-systems depending on the value are expected to not verify any proofs associated
-with the <a>verification method</a> at or after the time of revocation.
+SHOULD cease to be used. Once the value is set, it is not expected to be
```
OR13, line 1017: @msporny this seems like a great places to remove confusion over "verification method revocation" and "credential revocation"... we get this question a LOT... and it would nice to be able to share a note to a comment regarding this. something like:

> verification method revocation time information can reveal information about a controller. credential status do not include timing information, because it can degrade the privacy characteristics of some status lists... there is not formal relationship between these concepts, etc....

msporny: Agreed, I'll add that text to that effect in a future update to this PR.

index.html
```diff
@@ -614,10 +614,11 @@ <h3>Proofs</h3>
...
           </dd>
 
-          <dt><dfn class="lint-ignore">expires</dfn></dt>
+          <dt id="defn-proof-expires">expires</dt>
           <dd>
-An OPTIONAL property that conveys the date and time that a proof expires and that, if present, MUST be
-specified as an [[XMLSCHEMA11-2]] combined date and time string.
+An OPTIONAL property that conveys the date and time that a proof expires and
+that, if present, MUST be specified as an [[XMLSCHEMA11-2]] combined date and
+time string.
```
```suggestion
The `expires` property is OPTIONAL. If present, it MUST be an [[XMLSCHEMA11-2]]
`dateTimeStamp` string specifying when the proof expires.
```

TallTed (changes requested):
The language we developed later is better, and should be applied throughout, as here.

index.html
OR13: @dlongley ahh perhaps this is related to your comment on https://github.com/w3c/vc-data-model/issues/1262#issuecomment-1698238618

... confusion intensifies : )

I think what we mean to say here is that "ANY proof with created after `revoked` time MUST be considered invalid at *validation* time... as opposed to verification time, where we expect it to succeed, even for "revoked" verification methods?

dlongley: I think it's possible to parameterize "verification time" (and not always presume it's "the current system time"). It's that time that should be used when looking at these properties to determine if the proof verifies or not. We don't have to have things fall to "validation time". In other words, one could ask some verification software to verify a proof at time `X`.

vocab/security/vocabulary.yml
```diff
@@ -207,9 +207,11 @@ property:
...
 
   - id: expires
-    label: Proof expiration time
-    defined_by: https://www.w3.org/TR/vc-data-integrity/#dfn-expires
-    domain: sec:Proof
+    label: Expiration time for a proof or verification method
+    defined_by: https://www.w3.org/TR/vc-data-integrity/#defn-proof-expires
+    domain:
+      - sec:Proof
+      - sec:VerificationMethod
```
TallTed, line 214: According to @iherman in https://github.com/w3c/vc-data-model/issues/1262#issuecomment-1695821941
> the `domain` `expired` is set to both (i.e., the union) of `Proof` and `VerificationMethod`.

Note that `expires rdfs:domain Proof, VerificationMethod` does not set `rdfs:domain` to the _union_ of `Proof` and `VerificationMethod`, but rather to their _intersection_; that is, any entity with the property `expires` must be *both* a `Proof` _and_ a `VerificationMethod`.

For the apparently intended _union_, where an entity with the property `expires` could be *either* a `Proof` _or_ a `VerificationMethod`, `rdfs:domain` must be changed to `schema:domainIncludes`.
```suggestion
    domainIncludes:
      - sec:Proof
      - sec:VerificationMethod
```

msporny: @iherman would have to add this to the yml2vocab tool, as I don't think it has a `domainIncludes` property? Thoughts, @iherman?

index.html
msporny: Added in c4aede2d1851454a91856b1ee8f17d22d683d353.

msporny:

Issue #180, opened 2023-08-25
When converting to RDF, ensure that BASE URL is null.


msporny:
Since Data Integrity secured documents can be transferred from one security domain to another, processors that process the secured document cannot assume any particular BASE URL. When de-serializing to RDF, implementations MUST ensure that the BASE URL is set to null.

msporny:
PR #185 has been raised to address this issue. Once PR #185 is merged, this issue will be closed.

msporny:
PR #185 has been merged, closing.

PR #185, opened 2023-08-26
Ensure that base URL is set to null when processing as RDF.

msporny:
This PR attempts to address issue #180 by ensuring that the base URL is set to null when processing as RDF.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/185.html" title="Last updated on Sep 2, 2023, 3:15 PM UTC (917c4dc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/185/00bd146...917c4dc.html" title="Last updated on Sep 2, 2023, 3:15 PM UTC (917c4dc)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2147,7 +2147,7 @@ such as when an undefined term is detected in an input document.
 Similarly, since <a>conforming secured documents</a> can be transferred from one
 security domain to another, <a>conforming processors</a> that process the
 <a>conforming secured document</a> cannot assume any particular base URL for
-the document. When serializing to RDF, implementations MUST ensure that the
+the document. When deserializing to RDF, implementations MUST ensure that the
 base URL is set to null.
           </p>
         </section>
--- a/index.html
+++ b/index.html
@@ -2142,6 +2142,14 @@ Canonicalization [[?RDF-CANON]], MUST throw an error, which SHOULD be
 `DATA_LOSS_DETECTION_ERROR`, when data is dropped by a JSON-LD processor,
 such as when an undefined term is detected in an input document.
           </p>
+
+          <p>
+Similarly, since <a>conforming secured documents</a> can be transferred from one
+security domain to another, <a>conforming processors</a> that process the
+<a>conforming secured document</a> cannot assume any particular base URL for
+the document. When serializing to RDF, implementations MUST ensure that the
+base URL is set to null.
+          </p>
         </section>
 
         <section>


Wind4Greg (approved):

iherman (approved):
(This does not affect the approval of this particular PR)

Do we make it invalid for someone to do this:

```json
"@context" : [
    "https://....",
    "@base": "my url"
]
```

the `@base` JSON-LD directive is ignored in an external context file, but it is valid this way...

msporny:
> Do we make it invalid for someone to do this:
> the `@base` JSON-LD directive is ignored in an external context file, but it is valid this way...

Hrm, what we don't want to do is allow the BASE to be set by the processor based on where the document lives, IIUC. So, setting `@base` in the way that you showed in your example is fine -- because the BASE won't be mis-interpreted based on location, which is what this PR is attempting to fix. That said, processors that run in non-JSON-LD mode won't process that `@base` definition, leading to potential interop issues, so perhaps the guidance in this PR is best in order to increase interop.

iherman:
> > Do we make it invalid for someone to do this:
> > the `@base` JSON-LD directive is ignored in an external context file, but it is valid this way...
> 
> Hrm, what we don't want to do is allow the BASE to be set by the processor based on where the document lives, IIUC. So, setting `@base` in the way that you showed in your example is fine -- because the BASE won't be mis-interpreted based on location, which is what this PR is attempting to fix. 

Ok, I understand

> That said, processors that run in non-JSON-LD mode won't process that `@base` definition, leading to potential interop issues, so perhaps the guidance in this PR is best in order to increase interop.

Well, this PR does not cover this specific case. A non-JSON-LD processor will have no idea what to do with `@base`, and it might simply ignore it. But then there is a discrepancy. My proposal would be that, beyond the text you propose in the PR, we should disallow the `@base` statement for VC-s (I do not think it is frequently used, so it is not a major loss).

index.html
```diff
@@ -2142,6 +2142,14 @@ <h3>Securing Data Losslessly</h3>
 `DATA_LOSS_DETECTION_ERROR`, when data is dropped by a JSON-LD processor,
 such as when an undefined term is detected in an input document.
           </p>
+
+          <p>
+Similarly, since <a>conforming secured documents</a> can be transferred from one
+security domain to another, <a>conforming processors</a> that process the
+<a>conforming secured document</a> cannot assume any particular base URL for
+the document. When serializing to RDF, implementations MUST ensure that the
+base URL is set to null.
```
```suggestion
When serializing to RDF, implementations MUST ensure that the
base URL is set to null.
```

This can be a bit misread as if the base URL was a particular term in RDF that has to be set to null. This isn't the case, RDF does not have this notion (only serializations like JSON-LD or Turtle have).  I wonder whether this sentence serves any useful purpose, actually: the previous sentence does say that "cannot assume any particular base URL for the document", which sounds enough to me.

index.html
```diff
@@ -2142,6 +2142,14 @@ <h3>Securing Data Losslessly</h3>
 `DATA_LOSS_DETECTION_ERROR`, when data is dropped by a JSON-LD processor,
 such as when an undefined term is detected in an input document.
           </p>
+
+          <p>
+Similarly, since <a>conforming secured documents</a> can be transferred from one
+security domain to another, <a>conforming processors</a> that process the
+<a>conforming secured document</a> cannot assume any particular base URL for
+the document. When serializing to RDF, implementations MUST ensure that the
```
```suggestion
the document. When deserializing to RDF, implementations MUST ensure that the
```

dlongley (approved):

msporny:
> Well, this PR does not cover this specific case. A non-JSON-LD processor will have no idea what to do with @base, and it might simply ignore it. 

Yes, exactly, which is why we should probably tell authors that they "SHOULD NOT use `@base`".

> But then there is a discrepancy. My proposal would be that, beyond the text you propose in the PR, we should disallow the @base statement for VC-s (I do not think it is frequently used, so it is not a major loss).

I have raised https://github.com/w3c/vc-data-model/issues/1266 to track this issue.

index.html
msporny: A normative MUST ensures that we can test this functionality in the test suite. We want to test to make sure that relative URLs aren't used and/or are detected to increase interoperability. If we don't have the normative MUST statement, we won't have a test for this, and interop will suffer, IMHO.

msporny:
Normative, multiple reviews, changes requested and made, issues raised for remaining concern, no objections, merging.

Issue #191, opened 2023-09-14
Remove references to multibase


selfissued:
`proofValue` is defined as:
> A string value that contains the data necessary to verify the digital proof using the verificationMethod specified. The contents of the value MUST be a [[MULTIBASE](https://w3c.github.io/vc-data-integrity/#bib-multibase)]-encoded binary value.

Multibase is not a standard and may never become one.  Please change the definition of `proofValue` to utilize a standard encoding.  My preference would be for it to use the base64url encoding.

OR13:
previously discussed in https://github.com/w3c/vc-data-integrity/pull/92

Which later led to:

- https://github.com/w3c/vc-data-model/issues/1261
- https://github.com/w3c/vc-data-model/issues/1175

BigBlueHat:
Multibase is already normatively referenced in [DID-CORE](https://www.w3.org/TR/did-core/#bib-multibase), fwiw.

OR13:
Can someone answer this gentleman's question here https://github.com/w3c/did-core/issues/405#issuecomment-1719869201

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-09-14-vcwg#section3-1) on 2023-09-14

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 3.1. Remove normative dependencies on multibase (issue vc-data-integrity#191)

_See github issue [vc-data-integrity#191](https://github.com/w3c/vc-data-integrity/issues/191)._




**Michael Jones:** in doing due diligence on IETF standardization on mutlibase I discovered a normative dependence on multibase in DI, even though it's not a standard. we shouldn't be doing that. we shouldn't take normative deps on non-standards. requesting in #191 the definition of proof be changed to not use multibase.  

> *Dave Longley:* [https://w3c.github.io/vc-data-integrity/#resource-integrity](https://w3c.github.io/vc-data-integrity/#resource-integrity) <-- see "At risk" on what selfissued just said.

**Sebastian Crane:** been discussing a lot on how to proceed. one option is to include the definitions inside the specification, since multibase is not a complex format and it is stable. can be in DI itself.  
� in the future can add an IETF normative dependency.  

**Dave Longley:** as seabass said, we have a feature at risk in the spec for exactly this reason. don't think it's a pre-CR issue. we can change the current CR to fix this language as necessary.  

[...]

PR #196, opened 2023-09-15
Remove normative dependency on Multibase and Multihash.

msporny:
This PR has been raised to address issue #191, which requested the removal of Multibase from the list of normative dependencies for the vc-data-integrity specification. I went ahead and removed Multihash as well from the list of normative dependencies as I expect the request applies to that specification as well.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/196.html" title="Last updated on Oct 6, 2023, 8:36 PM UTC (1de6e43)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/196/9c62804...1de6e43.html" title="Last updated on Oct 6, 2023, 8:36 PM UTC (1de6e43)">Diff</a>


Wind4Greg (approved):

dlongley (approved):

selfissued (changes requested):
This PR doesn't do the job. Putting a question mark in the references doesn't remove the references. Please remove them entirety, as agreed to at TPAC.

msporny:
> This PR doesn't do the job. Putting a question mark in the references doesn't remove the references. Please remove them entirety, as agreed to at TPAC.

That is not what was agreed to at TPAC. The agreement was to make the references non-normative, which this PR does. Please review the minutes:

https://github.com/w3c/vc-data-integrity/issues/191#issuecomment-1720718132

Specifically:

> Brent Zundel: appreciate that there's a feature at risk tag. is the intention, should multibase not be standardized by the time this moves from CR to REC, would the current link be moved to a non normative reference?
> Sebastian Crane: if that satisfies Mike I am happy to make that compromise myself and make it an informative reference. we are waiting for the spec to catch up.
> Sebastian Crane: (to summarize) - we are going to, in the cryptosuites, to have a specific encoding, for the spec text we are using a format compatible with multibase, so that if it becomes a standard we will use it normatively. we won't publish a normative ref if it's not ready by REC.

There was no opposition to that summary during the meeting. I do expect there to be opposition to removing all non-normative references to multiformats from the specification. If you would like to make that suggestion, please raise a PR (though I expect it to have a large number of downvotes since it weakens the specification by not pointing to useful, stable documentation that exists today in a non-normative fashion).

David-Chadwick (commented):
Sorry but I do not understand how adding a ? removes the normative dependency

msporny:
> Sorry but I do not understand how adding a ? removes the normative dependency

It is a ReSpec shortcut: https://respec.org/docs/#normative-vs-informative-references

selfissued (changes requested):
It was my understanding from the discussion at TPAC was that the resolution was going to be that the references were going to be removed entirely until such time as there is a multibase standard to reference.  And that it would be replaced by in-spec statements that these values are strings beginning with either "u" or "z" and followed respectively by either a base64url-encoded value or a base58btc-encoded value.

Note that there's already text in the spec at https://w3c.github.io/vc-data-integrity/#resource-integrity defining these uses of the "u" and "z" prefixes.  Promote these to be normative and ditch the reference to the individual Internet Draft.

msporny:
> It was my understanding from the discussion at TPAC was that the resolution was going to be that the references were going to be removed entirely until such time as there is a multibase standard to reference.

As you can see [from the quoted text above](https://github.com/w3c/vc-data-integrity/pull/196#issuecomment-1722438799), that understanding was incorrect. The agreement was to make the reference non-normative, which this PR does.

> And that it would be replaced by in-spec statements that these values are strings beginning with either "u" or "z" and followed respectively by either a base64url-encoded value or a base58btc-encoded value.

That spec text has existed in the DI spec for some time now:

https://w3c.github.io/vc-data-integrity/#resource-integrity

> Promote these to be normative and ditch the reference to the individual Internet Draft.

The references to the I-Ds, which are non-normative now, are at the same level as linking to a Wikipedia article, or some content on the web, that is, they are informative in nature and helpful because they explain Multihash and Multibase in more depth than we do in the current specification. Removing access to documentation that is helpful to developers that want to learn more about how each technology works would be harmful to developers trying to understand more about each technology.

selfissued:
Replacing the normative reference with an informative reference effectively leaves `proofValue` undefined.  That's not OK, since it's a required field.  This PR needs to be updated to normatively define the contents of the `proofValue` field.

OR13:
I think the `proofValue` issue can be addressed without a normative reference to specific text encodings.

Something like:

"Based on the `@type` and `cryptoSuite` used by the proof, the specification defining the crypto suite can define the semantics of `proofValue`, the JSON value MUST be a string."

Then site base58btc normatively as the string encoding, but just copy the alphabet used in to the spec.

TallTed:
> [@selfissued] This PR needs to be updated to normatively define the contents of the `proofValue` field

I'm sure an explicit [change suggestion](https://github.com/w3c/vc-data-integrity/pull/196/files) would be welcome, and more likely to satisfy you than most if not all alternatives, though it may not be 100% accepted by the rest of the WG, who may yet make comments or suggest additional changes thereon (e.g., if your change request includes fully removing the informative references to `[[?MULTIBASE]]` and/or `[[?MULTIHASH]]` which have already replaced the normative references to `[[MULTIBASE]]` and `[[MULTIHASH]]`, I expect the informative references will be restored).

index.html
```diff
@@ -647,7 +647,7 @@ <h3>Proofs</h3>
           <dd>
 A string value that contains the data necessary to verify the digital proof
 using the `verificationMethod` specified. The contents of the value MUST be a
```
TallTed: Because it's relying on an informative reference, I think the strongest this can be is SHOULD, not MUST -- unless we replicate sufficient portion of the `[[?MULTIBASE]]` spec that this replication may be relied on for the MUST. Similar applies to other MUSTs citing informative references.
```suggestion
using the `verificationMethod` specified. The contents of the value SHOULD be a
```

TallTed (changes requested):
I can be explicit about all the problematic points, if needed.

index.html
msporny: We do specify a subset of the multibase values in https://w3c.github.io/vc-data-integrity/#resource-integrity (but might have to tweak the language to refer to that section now). I'm also fine w downgrading to a SHOULD, but am concerned that if we do that, we open the door up to non-multibase values in a field whose datatype is multibase.

OR13: If we are saying that `proofValue` is multibase... thats a normative dependency.

I don't see a way to remove the normative dependency, without opening the door to `proofValue`s that are not multibase... I also don't think thats a bad thing... its the crypto suite that should define the proof encoding, not the data integrity spec.

selfissued: Indeed, it should be perfectly legal for a cryptosuite to define its `proofValue` output as being pure base64url with no prefix.

dlongley: Using multibase enables compression when using other tools like CBOR-LD. There should always be a header and a single literal datatype (such that more than one context is not needed with potentially different term mapping definitions) to encapsulate the data. This enables flexibility in the face of future evolution whilst reducing overhead and preserving features in other layers.

So with a `DataIntegrityProof` in this version of the spec, the `proofValue` should always be encoded using multibase, i.e., there should be a header and a specific base to be used, as defined by the cryptosuite.

thereisnogabe (approved):

BigBlueHat (approved):

msporny:
@selfissued wrote:
> Replacing the normative reference with an informative reference effectively leaves proofValue undefined. That's not OK, since it's a required field.  This PR needs to be updated to normatively define the contents of the proofValue field.

Fixed in dd44d6b1513b96ae72c96ac245f65507290a301e.

msporny:
> "Based on the @type and cryptoSuite used by the proof, the specification defining the crypto suite can define the semantics of proofValue, the JSON value MUST be a string."

If we use a "string" datatype, it creates the following two issues:

1. CBOR-LD compression of multibase values breaks because there would be no way to tell if the first character in the `proofValue` is a multibase header or just so happens to be a value that starts with something that looks like a multibase value.
2. It just kicks the definition of the encoding of the `proofValue` to each cryptosuite, so we'd end up effectively repeating the same text here in every cryptosuite.

> Then site base58btc normatively as the string encoding, but just copy the alphabet used in to the spec.

That alphabet exists in the spec today: https://w3c.github.io/vc-data-integrity/#example-an-integrity-protected-image-that-is-associated-with-an-object

... but I moved it into its own section to make it more clear about the normative definition based on what @selfissued was asking for. That should address @TallTed's concern about SHOULD vs. MUST (it's a MUST now because the spec provides the normative values).

I believe the latest changes should address each of your concerns, requesting a re-review from each of you.

index.html
msporny: @TallTed wrote:
> unless we replicate sufficient portion of the [[?MULTIBASE]] spec that this replication may be relied on for the MUST.

This is what the newest changes to this PR does, based on the request from @selfissued.

index.html
```diff
@@ -658,11 +658,13 @@ <h3>Proofs</h3>
...
-value. The contents of this value are determined by a cryptosuite and set to the
-<em>proof value</em> generated by the
-<a href="#add-proof">Add Proof Algorithm</a>.
+[[?MULTIBASE]]-encoded binary value with a header and encoding described in
+<a href="#multibase-0"></a>. The contents of this value are determined by a
+specific cryptosuite and set to the <em>proof value</em> generated by the
+<a href="#add-proof">Add Proof Algorithm</a> for that cryptosuite. Alternative
+properties with different encodings specified by the cryptosuite MAY be used,
+instead of this property, to encode the data necessary to verify the digital
+proof.
```
```suggestion
[[?MULTIBASE]]-encoded binary value with a header and encoding as described in
<a href="#multibase-0"></a>. The contents of this value are determined by a
specific cryptosuite and set to the <em>proof value</em> generated by the
<a href="#add-proof">Add Proof Algorithm</a> for that cryptosuite. Alternative
properties with different encodings specified by the cryptosuite MAY be used,
instead of this property, to encode the data necessary to verify the digital
proof.
```

index.html
```diff
@@ -1719,51 +1721,21 @@ <h2>Capability Delegation</h2>
...
+        <h2>Multibase</h2>
 
-        <p class="issue atrisk" title="Unification of cryptographic hash expression formats are under discussion">
-The Working Group is currently attempting to determine whether cryptographic hash
-expression formats can be unified across all of the VCWG core specifications.
-Candidates for this mechanism include `digestSRI` and `digestMultibase`. There
-are arguments for and against unification that the WG is currently debating.
+        <p class="issue atrisk" title="Multibase and Multihash may be standardized at IETF">
+The  [[?MULTIBASE]] specification has been dispatched at IETF and may be
+standardized. There is active discussion on this initiative in
```
```suggestion
The [[?MULTIBASE]] specification has been dispatched at IETF and may be
standardized there. There is active discussion on this initiative in
```

index.html
selfissued: It sounds like @dlongley is saying that there needs to be a normative dependency on multibase here:
> So with a DataIntegrityProof in this version of the spec, the proofValue should always be encoded using multibase, i.e., there should be a header and a specific base to be used, as defined by the cryptosuite.

Removing that dependency is exactly the problem that this PR needs to be addressing, which it hasn't yet done.

index.html
```diff
@@ -1719,51 +1721,21 @@ <h2>Capability Delegation</h2>
...
-  },
-  ...
-}
-        </pre>
-
-        <p>
-Common [[MULTIBASE]] header values and their encoding alphabets are provided
-below:
+The following common [[?MULTIBASE]] header values and their encoding alphabets
+are normative and provided below:
```
```suggestion
The common [[?MULTIBASE]] header values and their encoding alphabets
as provided below are normative:

```

index.html
```diff
@@ -1795,13 +1767,27 @@ <h2>Resource Integrity</h2>
         </table>
 
         <p>
-Other [[MULTIBASE]] encoding values MAY be used, but interoperability is not
+Other [[?MULTIBASE]] encoding values MAY be used, but interoperability is not
 guaranteed between implementations.
```
```suggestion
guaranteed between implementations using such.
```

index.html
```diff
@@ -1795,13 +1767,27 @@ <h2>Resource Integrity</h2>
...
+before this specification goes to the Proposed Recommendation phase, the table
+below will be replaced with normative references to the Multihash
+specification.
+        </p>
 
         <p>
-Common [[MULTIHASH]] header values and their encodings are provided
-below:
+The following [[?MULTIHASH]] header values and their encodings are normative
+and provided below:
```
```suggestion
The [[?MULTIHASH]] header values and their encodings
as provided below are normative:
```

index.html
```diff
@@ -1850,16 +1836,51 @@ <h2>Resource Integrity</h2>
...
+
+        <p class="issue atrisk" title="Unification of cryptographic hash expression formats are under discussion">
+The Working Group is currently attempting to determine whether cryptographic hash
+expression formats can be unified across all of the VCWG core specifications.
+Candidates for this mechanism include `digestSRI` and `digestMultibase`. There
+are arguments for and against unification that the WG is currently debating.
+        </p>
+
+        <p>
+When including a link to an external resource in a
```
```suggestion
When a link to an external resource is included in a
```

TallTed (changes requested):
Editorial language tweaks

index.html
msporny: > It sounds like @dlongley is saying that there needs to be a normative dependency on multibase here

No, I don't believe that is what he is saying.

You said this before:
> And that it would be replaced by in-spec statements that these values are strings beginning with either "u" or "z" and followed respectively by either a base64url-encoded value or a base58btc-encoded value.... Note that there's already text in the spec at https://w3c.github.io/vc-data-integrity/#resource-integrity defining these uses of the "u" and "z" prefixes. Promote these to be normative and ditch the reference to the individual Internet Draft.

Those references are now normative (which is what you asked for), and the references to the Multibase/Multihash I-Ds at IETF are now non-normative (which is what the group agreed to at TPAC).

> Removing that dependency is exactly the problem that this PR needs to be addressing, which it hasn't yet done.

Where is the normative dependency that you are alluding to?

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-09-26-vcwg-special#section1-1) on 2023-09-26

- no resolutions were taken

#### 1.1. Remove normative dependency on Multibase and Multihash. (pr vc-data-integrity#196)

_See github pull request [vc-data-integrity#196](https://github.com/w3c/vc-data-integrity/pull/196)._




**Manu Sporny:** If you could take another look at that PR, Mike -- and if Orie can look at that PR then that's the last thing to be addressed before we go forward with CR.  
… I guess we can talk about that more tomorrow, this is just a request for a re-review.  

**Kristina Yasuda:** Cool, ok. Put that on the agenda for tomorrow, Orie can join tomorrow. The controller document conversations in DI and vc-jose-cose -- if you could take a look at my last comment and respond that would help.  

**Manu Sporny:** Kristina, if you made that comment an hour or two ago I probably thumbs up'd it.  

**Kristina Yasuda:** After that.  

**Manu Sporny:** I'll look.  

</details>

iherman (changes requested):
With the current setup, [§2.3.1.2](https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/196.html#multikey) is a normative section that does not have normative references. I do not think that is o.k. I believe that this section must be flagged as informative or marked as "at risk". The latter may force us to _remove_ the section if Multibase does not receive a normative stamp, which may harm already deployed software. I would therefore avoid that, and suggest marking the section as informative for now, making it clear (in a note) that if the IETF stamp is there, then this become normative (provided we have enough test results). Because `publicKeyJwk` is also a (normative) possibility to express the verification method, we do not really have a normative problem expressing that.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-09-27-vcwg#section2-2) on 2023-09-27

- no resolutions were taken

#### 2.2. Remove normative dependency on Multibase and Multihash. (pr vc-data-integrity#196)

_See github pull request [vc-data-integrity#196](https://github.com/w3c/vc-data-integrity/pull/196)._




**Manu Sporny:** selfissued suggested we removed normative dependencies on any multiformats specifications, because they aren't standards.  
… I heard consensus during f2f where we agree to make those references non-normative.  
… This is what the PRs I raised are doing.  
… I think what selfissued is asking for is the complete removal to any reference. Even if they are non-normative.  
… As for the other issues raised with other cryptosuites, they say what the normative values should be.  

**Michael Jones:** The core of the discussion wasn't about normative vs non-normative references. It was about required features needing to be normatively defined.  
… They have to be defined before going to CR.  

> *Manu Sporny:* [https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/196.html#dfn-proofvalue](https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/196.html#dfn-proofvalue).

> *Manu Sporny:* ^ normatively defined right there.

**Michael Jones:** I'm still seeing that the value is defined, which cannot be defined since multiformats isn't a standard.  
… I think we should be able to normatively defined the contents of every spec without referring to the multiformats spec.  
… Adding a note stating that it's compatible is ok.  

**Kristina Yasuda:** To clarify, you are asking the DI document to redefine multibase/multicodec instead of referring to the internet draft?  

**Michael Jones:** Yes, with the subset that actually matters.  

**Manu Sporny:** That's exactly what it does.  
… `<`manu reads the pr preview`>`.  

> *Michael Jones:* "The contents of the value MUST be a [MULTIBASE]-encoded binary value with a header and encoding as described in 2.4 Multibase." retains normative use.

**Kristina Yasuda:** Why do you need a reference to multibase?  

**Manu Sporny:** Because it's useful for implementers. Explains the background.  
… Useful as pointing to any non-normative document.  

> *Manu Sporny:* "[MULTIBASE]-encoded".

**Manu Sporny:** I'm confused what's being objected to.  

**Michael Jones:** quoted the sentence that's objected to.  

**Manu Sporny:** I'll make the changes to remove the multibase bit.  
… The other PRs point to the section that will be fixed.  
… That's how they refer to it normatively.  

> *Michael Jones:* See [https://github.com/w3c/vc-di-eddsa/pull/63/files](https://github.com/w3c/vc-di-eddsa/pull/63/files).

**Manu Sporny:** I'll make sure that all similarly worded phrases refer to the normative spec (i.e. DI).  

> *Kristina Yasuda:* same for [https://github.com/w3c/vc-di-eddsa/issues/62](https://github.com/w3c/vc-di-eddsa/issues/62) and [https://github.com/w3c/vc-di-bbs/issues/92.](https://github.com/w3c/vc-di-bbs/issues/92.).

**Michael Jones:** That also has a sentence.  

**Manu Sporny:** I can remove that as well.  
… we have a path forward.  

**Sebastian Crane:** Pointing out the multibase isn't a protected trademark.  
… We should be specific and define everything that we need.  
… We can still call it multibase.  

> *Orie Steele:* [https://twitter.com/multibase_co](https://twitter.com/multibase_co) is a YC company, doing web3 analytics.... which i get ads from constantly.

**Kristina Yasuda:** might be confusing, but that's fine.  
… got 3 minutes, any issue/PR?  

---

</details>

iherman:
> With the current setup, [§2.3.1.2](https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/196.html#multikey) is a normative section that does not have normative references. I do not think that is o.k. I believe that this section must be flagged as informative or marked as "at risk". The latter may force us to _remove_ the section if Multibase does not receive a normative stamp, which may harm already deployed software. I would therefore avoid that, and suggest marking the section as informative for now, making it clear (in a note) that if the IETF stamp is there, then this become normative (provided we have enough test results). Because `publicKeyJwk` is also a (normative) possibility to express the verification method, we do not really have a normative problem expressing that.

Having read the meeting minutes of the last call (see also https://github.com/w3c/vc-data-integrity/pull/196#issuecomment-1738307954) I realize I misread the text. The intention of the spec is to normatively define the portion of MULTIBASE that is used by us. That is fine. However, the normative definition should then _not refer_ to the MULTIBASE document; the text should stand by itself. It may (should) refer to it informatively to show the background for reasons of compatibility, and it is also o.k. to refer to do it in the initial note. At the moment, it is not clear how the 'z' and 'u' characters should be used, for example (without going back to the MULTIBASE document). 

My proposal is to 

- put there a _complete, comprehensive_ definition for the features we use
- give background information in the initial note, with a non-normative reference to Multibase
- the feature should then ***NOT*** be at risk! The exact text may change, but the feature, as used in DI is not at risk at all...
- the text in [Datatype specification](https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/196.html#multibase) must also be changed: the datatype is defined via §2.3.1.2 and _not_ via the Multibase spec (but that can also change later)

(The same comments apply to Multihash, b.t.w.)

OR13:
- https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/196.html#multibase
- https://github.com/w3c/vc-data-model/blob/main/contexts/credentials/v2#L313

@iherman are you saying this URL needs to change in the v2 context ^

msporny:
@selfissued @OR13 I have made the specific change requested on the call yesterday in this commit: https://github.com/w3c/vc-data-integrity/pull/196/commits/7bb2e688e4491a06f316e089877bee1799f84df5

I have also gone through and removed any other reference to [[?MULTIBASE]] or [[?MULTICODEC]] in normative statements as well to address your concerns in case they exist anywhere else in the specification:

b9d5d96dfb27d86b43063f56c49cdd181d3b212f

I also took @selfissued's advice and modified the issue marker to note that compatibility between what we're defining in the specification, the Multiformats community, and the I-Ds at IETF (which might go standards track in a future Multiformats WG) are a desired outcome.

@iherman I have also attempted to address your concerns in b9d5d96dfb27d86b43063f56c49cdd181d3b212f:

> put there a complete, comprehensive definition for the features we use

I think this is done as of the latest commit?

> give background information in the initial note, with a non-normative reference to Multibase

I think this is done as of the latest commit.

> the feature should then NOT be at risk! The exact text may change, but the feature, as used in DI is not at risk at all...

Yes, agree.

> the text in [Datatype specification](https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/196.html#multibase) must also be changed: the datatype is defined via §2.3.1.2 and not via the Multibase spec (but that can also change later)

I have updated that text in b9d5d96dfb27d86b43063f56c49cdd181d3b212f

I think this resolves all issues raised on the call yesterday. Requesting re-review from @OR13, @selfissued, and @iherman .

iherman:
> * https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/196.html#multibase
> * https://github.com/w3c/vc-data-model/blob/main/contexts/credentials/v2#L313
> 
> @iherman are you saying this URL needs to change in the v2 context ^

I do not understand the question, @OR13.

iherman:
@msporny,

> 
> > put there a complete, comprehensive definition for the features we use
> 
> I think this is done as of the latest commit?

I tried to look at this, and I do not see any change. What I see (I downloaded your branch to my local machine) is:


Screenshot 2023-09-28 at 16 15 06

I try to understand the table _without_ looking at the IETF document, and I do not understand. What does 'u' and 'z' mean, how they are used, how does the multibase value reall looks like? It is not said.

> 
> > give background information in the initial note, with a non-normative reference to Multibase
> 
> I think this is done as of the latest commit.

You mean in the introductory issue box? +1 to that.

> 
> > the feature should then NOT be at risk! The exact text may change, but the feature, as used in DI is not at risk at all...
> 
> Yes, agree.

But the issue box still says "Feature at risk". This should be removed (also in the multihash section)

> 
> > the text in [Datatype specification](https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/196.html#multibase) must also be changed: the datatype is defined via §2.3.1.2 and not via the Multibase spec (but that can also change later)
> 
> I have updated that text in [b9d5d96](https://github.com/w3c/vc-data-integrity/commit/b9d5d96dfb27d86b43063f56c49cdd181d3b212f)

Yes, that looks good to me.

> 
> I think this resolves all issues raised on the call yesterday. Requesting re-review from @OR13, @selfissued, and @iherman .

index.html
```diff
@@ -1719,51 +1721,25 @@ <h2>Capability Delegation</h2>
...
-  },
-  ...
-}
-        </pre>
-
         <p>
-Common [[MULTIBASE]] header values and their encoding alphabets are provided
-below:
+The common Multibase header values and their encoding alphabets
+as provided below are normative:
```
dlongley: To address Ivan's comment:

```suggestion
A Multibase string includes a single character header which identifies the
base and encoding alphabet used to encode a binary value, followed the
encoded binary value (using that base and alphabet). The common
Multibase header values and their associated base encoding alphabets as
provided below are normative:
```

iherman: Thanks, @dlongley, this is the kind of specification text I was looking for.

Although not part of the original PR, the same kind of text should also be added to the section on Multihash (which, in its current form, is no less cryptic)

OR13:
> > * https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/196.html#multibase
> > * https://github.com/w3c/vc-data-model/blob/main/contexts/credentials/v2#L313
> > 
> > @iherman are you saying this URL needs to change in the v2 context ^
> 
> I do not understand the question, @OR13.

@iherman `proofValue` is normatively defined via the JSON-LD context to be:

```
"@type": "https://w3id.org/security#multibase"
```

W3C is not defining multibase as an RDF type, W3ID.org (a separate legal entity with some overlap in governance, but not equivalent to this working group or W3C)  is defining multibase as an RDF type... @iherman I am asking you if this is a problem or not based on the other changes being made in this PR, from the perspective of W3C.

iherman:
> > > * https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/196.html#multibase
> > > * https://github.com/w3c/vc-data-model/blob/main/contexts/credentials/v2#L313
> > > 
> > > @iherman are you saying this URL needs to change in the v2 context ^
> > 
> > 
> > I do not understand the question, @OR13.
> 
> @iherman `proofValue` is normatively defined via the JSON-LD context to be:
> 
> ```
> "@type": "https://w3id.org/security#multibase"
> ```
> 
> W3C is not defining multibase as an RDF type, W3ID.org (a separate legal entity with some overlap in governance, but not equivalent to this working group or W3C) is defining multibase as an RDF type... @iherman I am asking you if this is a problem or not based on the other changes being made in this PR, from the perspective of W3C.

The definition of the multibase RDF Type is defined in the DI spec of W3C. It is an entry in the security vocabulary and, so far, the WG agreement is that this vocabulary's URI is `https://w3id.org/security`. There is nothing special about multibase in this respect. Put it another way, there is no difference between that datatype in the vocabulary and, say, the `proof` property. This PR has no effect on these whatsoever.

iherman:
> the definition of the multibase RDF Type is defined in the DI spec of W3C.

@OR13 , I forgot to make a reference to where this happens: look at §2.9.3 in the branch of this PR.

msporny:
@iherman wrote:
> I try to understand the table without looking at the IETF document, and I do not understand. What does 'u' and 'z' mean, how they are used, how does the multibase value reall looks like? It is not said.

This has been fixed in 2651487fc6fd4d29fa7d805d43f1b873ecadc42a and 02121e37066827d60bcb902c1dffa115484d4383.

> You mean in the introductory issue box? +1 to that.

Yes.

> But the issue box still says "Feature at risk". This should be removed (also in the multihash section)

Fixed in 55638820b0aace3b8c8fc4ef29ea4339c886c47f.

Does that address all of your concerns, @iherman ?

iherman (approved):
The latest changes answered all my concerns on this PR. To avoid the hassle of yet another PR, I would prefer to update the specification text for multihash in this PR, too, but I leave this to the editors.

msporny:
> The latest changes answered all my concerns on this PR. To avoid the hassle of yet another PR, I would prefer to update the specification text for multihash in this PR, too, but I leave this to the editors.

That was done in https://github.com/w3c/vc-data-integrity/commit/02121e37066827d60bcb902c1dffa115484d4383

index.html
```diff
@@ -1719,51 +1721,28 @@ <h2>Capability Delegation</h2>
...
-  },
-  ...
-}
-        </pre>
-
         <p>
-Common [[MULTIBASE]] header values and their encoding alphabets are provided
-below:
+A Multibase string includes a single character header which identifies the
+base and encoding alphabet used to encode a binary value, followed the
```
David-Chadwick: followed by....

index.html
```diff
@@ -1795,13 +1774,33 @@ <h2>Resource Integrity</h2>
         </table>
 
         <p>
-Other [[MULTIBASE]] encoding values MAY be used, but interoperability is not
-guaranteed between implementations.
+Other Multibase encoding values MAY be used, but interoperability is not
+guaranteed between implementations using such.
```
David-Chadwick: using such values.

David-Chadwick (approved):
Approved subject to the minor editorials

index.html
msporny: Fixed in 3a076dd89efb196b94cea106c6720b7aa632b6e5. Thank you!

index.html
msporny: Also fixed in 3a076dd89efb196b94cea106c6720b7aa632b6e5. Thank you!

OR13 (approved):

index.html
```diff
@@ -1188,10 +1190,10 @@ <h3>Multikey</h3>
 
             <p>
 In the example above, the `publicKeyMultibase` value starts with the letter `z`,
-which is the [[?MULTIBASE]] header that conveys that the binary data is
-base58-encoded using the Bitcoin base-encoding alphabet. The decoded binary data
-[[?MULTICODEC]] header is `0xed`, which specifies that the remaining data
-is a 32-byte raw Ed25519 <a>public key</a>.
+which is the <a href="#multibase-0">Multibase</a> header that conveys that the
+binary data is base-58-btc-encoded using the Bitcoin base-encoding alphabet. The
```
selfissued, line 1194: base-58-btc-encoded is not defined in the specification.  Given there's a normative dependency on it, it needs to be defined.

index.html
```diff
@@ -1188,10 +1190,10 @@ <h3>Multikey</h3>
...
+decoded binary data header is `0xed01`, which specifies that the
+remaining data is a 32-byte raw Ed25519 <a>public key</a>.
```
selfissued, line 1196: How does 0xed01 specify the format of the data?  This needs to be defined so implementers can do the same thing.

selfissued, line 1196: How does 0xed01 specify the format of the data? This needs to be defined so implementers can do the same thing.

index.html
```diff
@@ -1212,10 +1214,10 @@ <h3>Multikey</h3>
...
             <p>
 In the example above, the `secretKeyMultibase` value starts with the letter `z`,
-which is the [[?MULTIBASE]] header that conveys that the binary data is
-base58-encoded using the Bitcoin base-encoding alphabet. The decoded binary data
-[[?MULTICODEC]] header is `0x1300`, which specifies that the remaining data
-is a 32-byte raw Ed25519 private key.
+which is the <a href="#multibase-0">Multibase</a> header that conveys that the
+binary data is base-58-btc-encoded using the Bitcoin base-encoding alphabet. The
+decoded binary data header is `0x8026`, which specifies that the
+remaining data is a 32-byte raw Ed25519 private key.
```
selfissued, line 1220: How does 0x8026 specify the format of the data?  This needs to be defined so that implementers can do the same thing.

index.html
```diff
@@ -1719,51 +1721,28 @@ <h2>Capability Delegation</h2>
...
-    "digestMultibase": "zQmdfTbBqBPQ7VNxZEYEj14VmRuZBkqFbiwReogJgS1zR1n"
-  },
-  ...
-}
-        </pre>
-
         <p>
-Common [[MULTIBASE]] header values and their encoding alphabets are provided
-below:
+A Multibase string includes a single character header which identifies the
```
selfissued, line 1741: The meanings of these single characters needs to be defined, since there's a normative dependency on them.  Implementers need clear specifications of these data structures.

index.html
```diff
@@ -1795,13 +1774,33 @@ <h2>Resource Integrity</h2>
...
+alignment between the Multihash values used in this specification and the
+Multihash values defined by the current Multihash community and any potential
+future IETF Multiformats Working Group.
         </p>
 
         <p>
-Common [[MULTIHASH]] header values and their encodings are provided
-below:
+A Multihash value starts with a binary header, which identifies the specific
+cryptographic hash algorithm and parameters used to generate the digest,
```
selfissued, line 1816: The binary values identifying the particular cryptographic hash algorithms needs to be defined in this specification for people to be able to interoperably implement this.

index.html
```diff
@@ -1850,16 +1849,52 @@ <h2>Resource Integrity</h2>
...
+<a>conforming secured document</a>, it is desirable to know whether the resource
+that is identified has changed since the proof was created. This applies to
+cases where there is an external resource that is remotely retrieved as well as
+to cases where the <a>verifier</a> might have a locally cached copy of the
+resource.
+        </p>
+        <p>
+To enable confirmation that a resource referenced by a <a>conforming secured document</a>
+has not changed since the document was secured, an implementer MAY include a
+property named <dfn class="lint-ignore">`digestMultibase`</dfn> in any object
```
selfissued, line 1903: I don't see where in the spec the digestMultibase property is defined.

index.html
```diff
@@ -1850,16 +1849,52 @@ <h2>Resource Integrity</h2>
...
-specifications.
+      </section>
+
+      <section>
+        <h2>Resource Integrity</h2>
+
+        <p class="issue atrisk" title="Unification of cryptographic hash expression formats are under discussion">
+The Working Group is currently attempting to determine whether cryptographic hash
+expression formats can be unified across all of the VCWG core specifications.
+Candidates for this mechanism include `digestSRI` and `digestMultibase`. There
```
selfissued, line 1888: I don't see where in this specification the digestSRI property is defined.

index.html
```diff
@@ -2259,12 +2294,13 @@ <h4 id="cryptosuiteString">The `cryptosuiteString` Datatype</h3>
...
-formats, which are useful in environments that cannot directly represent binary
-values. This specification makes use of this encoding. In environments that
-support data types for string values, such as RDF [[?RDF-CONCEPTS]],
-[[?MULTIBASE]]-encoded content is indicated using a literal value whose datatype
-is set to `https://w3id.org/security#multibase`.
+<a href="#multibase">Multibase</a>-encoded strings are used to encode binary
+data into ASCII-only formats, which are useful in environments that cannot
+directly represent binary values. This specification makes use of this encoding.
+In environments that support data types for string values, such as RDF
+[[?RDF-CONCEPTS]], <a href="#multibase">Multibase</a>-encoded content is
```
selfissued, line 2329: If we're citing RDF-CONCEPTS, it would be useful to have a sentence with the citation describing how it informs implementations in this case.

index.html
```diff
@@ -2278,9 +2314,9 @@ <h4 id="multibase">The `multibase` Datatype</h3>
             </dd>
             <dt>The lexical space</dt>
             <dd>
-Any string that starts with a [[?MULTIBASE]] character and the rest of
-the characters consist of allowable characters in the respective base-encoding
-alphabet.
+Any string that starts with a <a href="#multibase">Multibase</a> character and
```
selfissued, line 2345: Please define what the multibase characters are.

index.html
```diff
@@ -2290,7 +2326,7 @@ <h4 id="multibase">The `multibase` Datatype</h3>
             <dd>
 Any element of the lexical space is mapped to the value space by base-decoding
 the value based on the base-decoding alphabet associated with the
```
selfissued, line 2356: Define the base-decoding alphabet and the meanings of its values.  The definitions needs to be specific enough to enable interoperable implementations.

That, or use of the base-decoding alphabet should be removed from the specification.

index.html
```diff
@@ -2290,7 +2326,7 @@ <h4 id="multibase">The `multibase` Datatype</h3>
...
-first  [[?MULTIBASE]] character in the lexical string.
+first <a href="#multibase">Multibase</a> character in the lexical string.
```
selfissued, line 2357: Again, "multibase character" is currently undefined.

selfissued (changes requested):
As described in my comments, the uses of multibase in the specification are not currently normatively defined to a sufficient degree that interoperable implementations are enabled.  This needs to be corrected or the use of multibase needs to be removed.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-10-04-vcwg#section1-3) on 2023-10-04

- no resolutions were taken

#### 1.3. Remove normative dependency on Multibase and Multihash. (pr vc-data-integrity#196)

_See github pull request [vc-data-integrity#196](https://github.com/w3c/vc-data-integrity/pull/196)._




**Manu Sporny:** this is a base PR for many other things.  
… selfissued has made many requests.  
… and I want to get through them on the call.  

> *Manu Sporny:* See [https://github.com/w3c/vc-data-integrity/pull/196#pullrequestreview-1657894864](https://github.com/w3c/vc-data-integrity/pull/196#pullrequestreview-1657894864).

**Manu Sporny:** 'cause this is dragging out.  
… selfissued, you are asking for more normative definitions where there are just examples.  
… so the normative references you want are in other specs.  

> *Manu Sporny:* [https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/196.html#multibase-0](https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/196.html#multibase-0).

**Manu Sporny:** this preview ^ shows these definitions have been in there for awhile.  
… it's normatively defined in the specification.  
… the other thing is asking about the header format and how they're used.  
… they're headers, they're doing what they've been defined to do.  
… these are examples, so they are not normative.  

> *Manu Sporny:* [https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/196.html#resource-integrity](https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/196.html#resource-integrity).

**Manu Sporny:** same thing is true for the others also.  
… this stuff is defined in there.  
… I can go and respond to each one, but they are all already normatively defined.  

**Michael Jones:** I am reading the specification.  
… the text that you're talking about does not normatively define how to translate the alphabet to string or binary.  
… in all cases where we're using data structures, we need to define in sufficient detail, so implementers can build implementations.  
… I'm still doing a review, but there's similar incomplete definitions in EDSA as well.  
… counting on tribal knowledge isn't acceptable when we're normatively referencing something.  

**Manu Sporny:** base encoding something is something no ones had a problem with since forever.  
… no other spec has to define this.  

**Michael Jones:** no. you must define it in the spec.  

> *Orie Steele:* `@manu`, you can paste in this example: [https://github.com/cryptocoinjs/base-x/blob/master/src/index.js.](https://github.com/cryptocoinjs/base-x/blob/master/src/index.js.) lol.

**Manu Sporny:** then if I define base encoding, will you let us merge this?  

**Michael Jones:** yes.  

**Manu Sporny:** so you want all these things redefined directly in this spec?  

**Michael Jones:** yes.  
… there's a term, multibase prefix...and that's not defined.  
… when I review stuff, I like to pretend I'm an implementer.  
… and imagine whether I can build code with that of others.  
… and if I can't, then I raise issues.  

> *Manu Sporny:* Here is the normative definition: [https://pr-preview.s3.amazonaws.com/w3c/vc-di-eddsa/pull/63.html#multikey](https://pr-preview.s3.amazonaws.com/w3c/vc-di-eddsa/pull/63.html#multikey).

> *Manu Sporny:* Here is the other normative definition: [https://pr-preview.s3.amazonaws.com/w3c/vc-di-ecdsa/pull/42.html#multikey](https://pr-preview.s3.amazonaws.com/w3c/vc-di-ecdsa/pull/42.html#multikey).

**Manu Sporny:** I have pointed directly to the normatively to the definitions.  
… and they tell you how to build the format front to back.  
… I can make these changes, selfissued.  
… but every time I make a change, you request new ones.  

**Michael Jones:** I'm sorry that's the way it feels.  
… but the intent is to have each of these specs completely define everything implementers need to do to implement everything.  

> *Dave Longley:* would referencing RFC 4648 and say that the bases used are `58`, `64`, and the alphabets are `base58-btc` and `base64-no-pad`?

> *Dave Longley:* would that help?

> *Dave Longley:* [https://datatracker.ietf.org/doc/html/rfc4648](https://datatracker.ietf.org/doc/html/rfc4648).

**Manu Sporny:** can you read what dlongley just shared.  

**Michael Jones:** maybe. I'd have to look.  

> *Orie Steele:* ref'ing that RFC seems like a good move.

**Manu Sporny:** I can address all the comments.  

**Kristina Yasuda:** sorry we can't pick a date.  
… we'll continue to believe there's good intent here.  
… but please address these concerns selfissued.  

**Sebastian Crane:** it seems selfissued's main concern is anything multibase/multikey.  
… it seems our charter includes defining these things.  

> *Orie Steele:* +1 to potentially publishing multibase as a w3c document.

**Sebastian Crane:** so, if selfissued, you take issue with these not being defined at the IETF, then a probably solution is to take these multibase/multikey stuff through the W3C process within this group.  

**Kristina Yasuda:** we can talk about multibase.  
… but I do not think it is realistic within the lifetime of the WG.  
… in the meantime if selfissued and manu can work through the PR that would help.  
… next week is IIW, but we will cancel the special topic call.  
… but keep the main call.  
… I think the special topic call will be about accessibility.  
… so seabass if you could send stuff out about it, that'd be great.  

**Sebastian Crane:** I will be absent for part of the coming weeks.  

**Ivan Herman:** next week, WG is at a EU unfriendly time.  
… and I want to be part of the timing discussions.  
… so you're welcome to meet and provide a time...but I may have to disapprove it after I see it.  

**Kristina Yasuda:** yeah, lets discuss offline.  
… thx all.  
… thx for scribing bigbluehat.  

---

</details>

index.html
msporny: There is a normative section on Multibase and base-58-btc here:

https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/196.html#multibase-0

There is now a normative section on how to do arbitrary base-encoding here (with normative algorithms and example code on how to implement those algorithms):

https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/196.html#base-encode

There is now also a normative section on how to do arbitrary base-decoding here (with normative algorithms and example code on how to implement those algorithms):

https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/196.html#base-decode

base-58-btc-encoding is now normatively defined in the specification. If you disagree, please provide concrete text that would address your concern.

index.html
msporny: You are referring to an example. Examples don't normatively define things in the specification. The normative definition of that value is in the EdDSA Cryptosuite specification:

https://pr-preview.s3.amazonaws.com/w3c/vc-di-eddsa/pull/63.html#multikey

`0xed01` is the Multicodec header for a Ed25519 public key (which is what the text above says). This is normatively defined in the EdDSA Cryptosuite specification and it's done in a way where it's clear what the entire field should be. If you disagree, please provide concrete text to address your concern.

index.html
msporny: You are referring to an example. Examples don't normatively define things in the specification. The normative definition of that value is in the EdDSA Cryptosuite specification:

https://pr-preview.s3.amazonaws.com/w3c/vc-di-eddsa/pull/63.html#multikey

`0x8026` is the Multicodec header for a Ed25519 private key (which is what the text above says). This is normatively defined in the EdDSA Cryptosuite specification and it's done in a way where it's clear what the entire field should be. If you disagree, please provide concrete text to address your concern.

index.html
msporny: The meanings of these single characters are defined here:

https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/196.html#multibase-0

index.html
msporny: The binary values are defined here:

https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/196.html#multihash

index.html
msporny: `digestSRI` is defined in VCDM -- https://w3c.github.io/vc-data-model/#integrity-of-related-resources

index.html
msporny: `digestMultibase` is defined here: https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/196.html#resource-integrity

index.html
msporny: It is not clear what sort of sentence you want written here. Please provide concrete specification text, or at least more prose suggesting what you want to see here.

index.html
msporny: That is done here:

https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/196.html#multibase-0

If you disagree, please provide concrete text regarding what you'd like to see.

index.html
msporny: It's defined here:

https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/196.html#multibase-0

What, specifically, are you looking for?

index.html
msporny: That is done here:

https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/196.html#multibase-0

and here:

https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/196.html#base-encode
https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/196.html#base-decode

When coupled with the Multibase section linked to above, this provides more than enough information to provide interoperable implementations. Producing interoperable implementations around multibase has not been a challenge for implementers to date. That said, we have a lot of normative text in the specification that implementers can follow now.

If you disagree, please provide some concrete text suggestions.

index.html
iherman: > digestSRI is defined in VCDM -- https://w3c.github.io/vc-data-model/#integrity-of-related-resources

See also https://github.com/w3c/vc-data-model/pull/1296 which formalizes that.

dmitrizagidulin (approved):
👍

seabass-labrax (approved):
LGTM!

msporny:
@selfissued re-ping to re-review... ideally, you approve before we merge this. It's all green on reviews so far.

selfissued (approved):
Thanks for doing this.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-10-11-vcwg#section1-1) on 2023-10-11

- no resolutions were taken

#### 1.1. Remove normative dependency on Multibase and Multihash. (pr vc-data-integrity#196)

_See github pull request [vc-data-integrity#196](https://github.com/w3c/vc-data-integrity/pull/196)._




**Manu Sporny:** Mike Jones had requested that we normatively define a number of things in multibase/multihash. The changes have been to define how base encoding works and provide algorithms for base encoding and decoding and specifically provide the alphabets for the bases used.  
… Those are the updates and changes and that has gone into PR #196.  
… I believe I have addressed every single one of Mike Jones' requests and questions and AFAICT everything is normatively defined. The hope is to get his review in and get this PR in before Nov 7.  

**Brent Zundel:** Merge it.  
… Mike has approved.  

</details>

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-10-11-vcwg#section1-2) on 2023-10-11

- no resolutions were taken

#### 1.2. Remove references to Multiformats specifications. (pr vc-di-eddsa#63)

_See github pull request [vc-di-eddsa#63](https://github.com/w3c/vc-di-eddsa/pull/63)._




**Manu Sporny:** That is the basis for the other two PRs.  

_See github pull request [vc-data-integrity#196](https://github.com/w3c/vc-data-integrity/pull/196)._




**Manu Sporny:** PR #63 builds on DI PR #196. My presumption there is if Mike is good with 196, he's good with the others that refer to it.  
… He has +1'd that as well.  

</details>

msporny:

PR #201, opened 2023-09-18
Fix context media types served by github pages.

dlongley:
This adds symlinks to each context to ensure GitHub pages serves the proper media type. The symlinks added here can be removed in the future once hosting is provided by another service that uses the proper media types without them.


msporny (approved):

msporny:
It's my understanding that we're getting complaints from implementers (from JFF Plugfest 3) now that they are using the `https://w3id.org/security/data-integrity/v1` redirect (and Github pages is serving the wrong media type for these files without the extensions). 

Given this is a bug fix that doesn't modify any content and ensures that Github pages is operating properly, merging.

An upstream fix will be required in w3id.org to point to the files w/ the file extensions (so Github pages serves them properly).

PR #212, opened 2023-10-16
Updated the URL-s in the vocabulary

iherman:
This is an ***editorial*** cleanup for the URL-s in the voabulary, following the separate discussion in https://github.com/w3c/vc-data-model/pull/1296#discussion_r1357635748 but also https://github.com/w3c/vc-data-model/pull/1296#discussion_r1357668261.

The generated files are not on `/TR`, so this PR does not affect the current CR review.

### DIFF
--- a/vocab/security/vocabulary.drawio
+++ b/vocab/security/vocabulary.drawio
@@ -94,17 +94,17 @@
         <mxCell id="Uf8WLKuzS3drS_BCJ-BJ-112" value="&lt;span style=&quot;color: rgb(0, 0, 0); font-family: Helvetica; font-size: 12px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: center; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(251, 251, 251); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial; float: none; display: inline !important;&quot;&gt;Datatype&lt;/span&gt;" style="text;whiteSpace=wrap;html=1;fontSize=16;" parent="Uf8WLKuzS3drS_BCJ-BJ-113" vertex="1">
           <mxGeometry width="90" height="40" as="geometry" />
         </mxCell>
-        <UserObject label="&lt;i&gt;VerificationMethod&lt;/i&gt;" link="https://w3id.org/security/#VerificationMethod" id="Uf8WLKuzS3drS_BCJ-BJ-37">
+        <UserObject label="&lt;i&gt;VerificationMethod&lt;/i&gt;" link="https://w3id.org/security#VerificationMethod" id="Uf8WLKuzS3drS_BCJ-BJ-37">
           <mxCell style="ellipse;whiteSpace=wrap;html=1;fontSize=16;fillColor=none;strokeWidth=2;strokeColor=#336600;" parent="1" vertex="1">
             <mxGeometry x="1104.668136020151" y="30" width="208.7336272040302" height="46.89083823529412" as="geometry" />
           </mxCell>
         </UserObject>
-        <UserObject label="&lt;i&gt;controller&lt;/i&gt;" link="https://w3id.org/security/#controller" id="Uf8WLKuzS3drS_BCJ-BJ-44">
+        <UserObject label="&lt;i&gt;controller&lt;/i&gt;" link="https://w3id.org/security#controller" id="Uf8WLKuzS3drS_BCJ-BJ-44">
           <mxCell style="rounded=1;whiteSpace=wrap;html=1;fontSize=16;fillColor=none;strokeWidth=2;" parent="1" vertex="1">
             <mxGeometry x="868.003765743073" y="298.9969117647059" width="163.7314231738035" height="32.536500000000004" as="geometry" />
           </mxCell>
         </UserObject>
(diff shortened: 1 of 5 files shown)


Wind4Greg (approved):

vocab/security/vocabulary.drawio
```diff
@@ -237,42 +237,42 @@
...
           </mxCell>
         </UserObject>
-        <UserObject label="&lt;i&gt;proofValue&lt;/i&gt;" link="https://w3.org/2018/credentials/#proofValue" id="Uf8WLKuzS3drS_BCJ-BJ-11">
+        <UserObject label="&lt;i&gt;proofValue&lt;/i&gt;" link="https://w3id.org/security#proofValue" id="Uf8WLKuzS3drS_BCJ-BJ-11">
           <mxCell style="rounded=1;whiteSpace=wrap;html=1;fontSize=16;fillColor=none;strokeWidth=2;" parent="1" vertex="1">
             <mxGeometry x="555.3489294710328" y="440.29808823529413" width="163.7314231738035" height="32.536500000000004" as="geometry" />
           </mxCell>
         </UserObject>
-        <UserObject label="&lt;i&gt;expires&lt;/i&gt;" link="https://w3.org/2018/credentials/#expires" id="Uf8WLKuzS3drS_BCJ-BJ-12">
+        <UserObject label="&lt;i&gt;expires&lt;/i&gt;" link="https://w3id.org/security#expires" id="Uf8WLKuzS3drS_BCJ-BJ-12">
```
dlongley: This should be "expiration" ... but perhaps not for this PR. The context term is "expires" but the URL uses "#expiration". It's a unicorn, but has been in use for years that way.

```suggestion
        <UserObject label="&lt;i&gt;expires&lt;/i&gt;" link="https://w3id.org/security#expiration" id="Uf8WLKuzS3drS_BCJ-BJ-12">
```

dlongley (approved):

vocab/security/vocabulary.drawio
iherman: @dlongley Ah. So the term is `expires` in the spec, and, indeed, the `@context` uses `expiration` in the URL. Ouch.

There seems to be two ways to change this:

- Set the URL in the vocabulary to `expiration` rather than `expires` (the fact that the diagram uses `expires` is only an artifact). That means the vocabulary will be at odds with what the spec says, which is ugly.
- Update the `@context` file to the right ID

The second option seems to be the correct way of doing this. Any reason why we should not?

(But yes, I believe we should leave this for a separate PR.)

dlongley: > The second option seems to be the correct way of doing this. Any reason why we should not?

Yes, the reason not to do it is that many things have shipped / escaped into the wild using `#expiration` in the URL. The contexts have been that way for a long while and we should probably just embrace it. The vocab was erroneously changed to use `#expires` in some recent update that no one picked up on because it is a special unicorn -- and it should be changed back to `#expiration`, IMO. It's not pretty, but it is the way it's been done for a long time now and I don't think it buys us much to try and change to `#expires` -- with any number of possible issues coming up in the ecosystem if we do.

iherman: I am holding my nose but... I will do it tomorrow. It is more than what you put there as a change request: the change must also be made in the svg file and, more importantly, in the vocabulary.yml file as well.

iherman: @dlongley @msporny I have made the `expires -> expiration` change in https://github.com/w3c/vc-data-integrity/pull/212/commits/8afd6b9ff3a2cdfb7141843ad27511d8d5f1675d.

I was wondering whether we should put a small text as a comment into the vocabulary file explaining why, if one dereferences the definition link, one gets to the term `expires` in the VCDM spec. Because it seems to have a historical background, I would prefer one of you two to propose a text. Alternatively, we may remain silent...

(If you decide that a text is better, you can directly edit the yml file by adding a `comment:` field to the definition of `expiration`)

vocab/security/vocabulary.yml
```diff
@@ -206,7 +206,7 @@ property:
     domain: sec:Proof
     range: xsd:dateTime
 
-  - id: expires
+  - id: expiration
     label: Expiration time for a proof or verification method
     defined_by: https://www.w3.org/TR/vc-data-integrity/#defn-proof-expires
     domain:
```
dlongley, line 213:
```suggestion
    comment: This property is often expressed using `expires` as a shortened term in JSON-LD, for historical purposes; since this shortened term and its mapping to this property are in significant use in the ecosystem, this inconsistency between the short term name and the property identifier is expected.
    domain:
```

vocab/security/vocabulary.yml
```diff
@@ -209,6 +209,7 @@ property:
   - id: expiration
     label: Expiration time for a proof or verification method
     defined_by: https://www.w3.org/TR/vc-data-integrity/#defn-proof-expires
+    comment: This property is often expressed using `expires` as a shortened term in JSON-LD, for historical purposes; since this shortened term and its mapping to this property are in significant use in the ecosystem, this inconsistency between the short term name and the property identifier is expected.
```
```suggestion
    comment: Historically, this property has often been expressed using `expires` as a shortened term in JSON-LD. Since this shortened term and its mapping to this property are in significant use in the ecosystem, the inconsistency between the short term name (`expires`) and the property identifier (`...#expiration`) is expected and should not trigger an error.
```

TallTed (changes requested):
An attempt to clarify through some language tweaks...

msporny (approved):

msporny:

Issue #211, opened 2023-10-13
Show i18n best practice in example


msporny:
A quick i18n review by @aphillips requested that we demonstrate i18n best practices:

https://github.com/w3c/i18n-activity/issues/1771

This issue tracks that concern such that it's addressed before the DI specs go to CR.

msporny:
We should add an example that demonstrates how to secure an internationalized string.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-10-17-vcwg-special#section1-1) on 2023-10-17

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 1.1. Show i18n best practice in example (issue vc-data-integrity#211)

_See github issue [vc-data-integrity#211](https://github.com/w3c/vc-data-integrity/issues/211)._




**Manu Sporny:** We have started the CR transition for the DI specs and JSON schema, I made a mistake and didn't properly request i18n review, but fortunately Addison took care of this for us. He would like to see an example of an i18n string.  
… We have a hello world example that I'd like to keep as this is helpful in all specs generally, but I want to add an i18n example and that needs to go in before CR.  
… I make a PR to do that and use the mechanism that we all agreed to for that.  
… I would like to merge it before 7 days, it's informative, no normative text.  

**Brent Zundel:** The plan, as I understand it, is to raise a PR to add a non-normative i18n example to data integrity and then merge it at some point before the 7 day window.  
… Any objections to that plan of action?  
… No objections here, thank you, you're good to go.  

</details>

msporny:
PR #213 has been raised to address this issue. This issue will be closed once PR #213 has been merged.

aphillips:
Pasting in the I18N comment:

---
Example 1
https://www.w3.org/TR/vc-data-integrity/#example-a-simple-json-data-document

>```
>{
>  "title": "Hello world!"
> };
>```

Also found in the other docs in this review, e.g.

- https://www.w3.org/TR/vc-di-eddsa/#example-an-ed25519-digital-signature-expressed-as-a-ed25519signature2020
- https://www.w3.org/TR/vc-di-eddsa/#example-credential-without-proof 
- https://www.w3.org/TR/vc-di-eddsa/#example-credential-without-proof-1
- https://www.w3.org/TR/vc-di-ecdsa/#example-an-ecdsa-p-256-digital-signature-expressed-as-a-dataintegrityproof

[...]

PR #213, opened 2023-10-18
Add Internationalization section

msporny:
This PR is an attempt to address issue #211 by adding an Internationalization Considerations section that covers:

1. How to express i18n metadata in both JSON and JSON-LD by pointing to other NOTEs and RECs that define how to do so, and 
2. Examples on how to express i18n information in both JSON and JSON-LD.

As covered [during the call today](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-10-17-vcwg-special#section1-1), we will merge this PR faster than the typical 7 day window after we have multiple reviews. I plan to merge it by Thursday if there are no objections and at least two reviews.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/213.html" title="Last updated on Oct 21, 2023, 12:01 AM UTC (57f0e78)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/213/a6377ee...57f0e78.html" title="Last updated on Oct 21, 2023, 12:01 AM UTC (57f0e78)">Diff</a>


iherman (approved):
I am fine with merging this PR. However, on longer term, I think the issue I raised for the VCDM (https://github.com/w3c/vc-data-model/issues/1313) is relevant here as well. But possible changes on the examples are purely editorial, not affecting any normative section, so no reason to hold up the merge.

msporny:
> I think the issue I raised for the VCDM ([w3c/vc-data-model#1313](https://github.com/w3c/vc-data-model/issues/1313)) is relevant here as well.

Yes, agreed, we'll need a better example that makes using `direction` mandatory for proper display of the text.

dlongley (approved):

longpd (approved):

Wind4Greg (approved):

index.html
```diff
@@ -4152,6 +4152,105 @@ <h2>Presenting Time Values</h2>
 
     </section>
 
+    <section>
```
aphillips: I think the change you made above (to replace the string "hello world" with a data representation) addresses our (pending) comment. 

I'm not sure I would add an I18N Considerations section. (Generally I18N recommends writing a section like this only when it goes beyond "eat your vegetables".) While there isn't really anything wrong with the text here, it's somewhat orthogonal to this specific spec's task. Implementers will need to read the other VC specs anyway, where they will find similar guidance in a more directly-relevant context. 

So... unless you disagree, I would just drop this?

aphillips (changes requested):

index.html
```diff
@@ -4152,6 +4152,105 @@ <h2>Presenting Time Values</h2>
...
+specification is used by system software to secure digital information and
+the internationalization support of that digital information is a function of
+the data that is being secured, and not this specification.
+      </p>
+
+      <section>
+        <h2>Internationalization in JSON</h2>
+
+        <p>
+There is currently no standard way to express language and text direction
```
```suggestion
There is currently no standard way to express language and/or text direction
```

TallTed (changes requested):
Small tweak

index.html
msporny: Ok, I'll drop the section before merging this. I was attempting to give [[STRING-META]] some airtime along w/ pointing folks to relevant i18n sections elsewhere. That said, seems like there is support to drop it, so I will.

msporny: Done in https://github.com/w3c/vc-data-integrity/pull/213/commits/57f0e78a2a8b5713e53c4af5307a3c60a823350f

index.html
msporny: This section was removed as requested in https://github.com/w3c/vc-data-integrity/pull/213#discussion_r1364183506, so the change suggestion wasn't applied.

msporny:

PR #224, opened 2023-12-03
Binding Proof Chains Together

Wind4Greg:
The current algorithm in section 4.4 Add Proof Set/Chain doesn't bind proofs in a chain with each other. This slight modification of the algorithm achieves this result. This PR:

1. Modifies section 4.4 Add Proof Set/Chain to provide this binding. Thanks to @dlongley for suggesting the mechanism.
2. Modifies section 4.6 Verify Proof Sets and Chains to work with the improved algorithm
3. Modifies section 4.8 Processing Errors to formally define the `PROOF_CHAIN_CYCLE_ERROR` that was in the text but not defined.

Files for generating test vectors for these updated procedures have been created at https://github.com/Wind4Greg/EdDSA-Test-Vectors/tree/proof-chains.

***
<a href="https://pr-preview.s3.amazonaws.com/Wind4Greg/vc-data-integrity/pull/224.html" title="Last updated on Dec 12, 2023, 11:14 PM UTC (ba5482e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/224/8859155...Wind4Greg:ba5482e.html" title="Last updated on Dec 12, 2023, 11:14 PM UTC (ba5482e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3118,10 +3118,11 @@ Set <var>allProofs</var> to </var><var>securedDocument</var>.<var>proof</var>.
           <li>
 For each <var>proof</var> in <var>allProofs</var>, do the following steps:
           </li>
-          <li>
+          <ol class="algorithm">
+            <li>
 Let <var>matchingProofs</var> be an empty list.
-          </li>
-          <li>
+            </li>
+            <li>
 If <var>proof</var> contains a <code>previousProof</code> attribute and that
 attribute is a string, add the element from <var>allProofs</var> with an
 <code>id</code> attribute matching <code>previousProof</code> to matchingProofs.
@@ -3133,25 +3134,26 @@ If the <code>previousProof</code> attribute is an array, add each element from
 of that array. If any matches are missing, an error MUST be raised and SHOULD
 convey an error type of
 <a href="#MALFORMED_PROOF_ERROR">MALFORMED_PROOF_ERROR</a>.
-          </li>
-          <li>
+            </li>
+            <li>
 Let <var>unsecuredDocument</var> be a copy of <var>securedDocument</var> with
 the proof value removed and then set <var>unsecuredDocument.proof</var>
 to <var>matchingProofs</var>.
-          <li>
+            </li>
+            <li>
 Run steps 2 through 4 to validate the contents of <var>proof</var>.
-          </li>
-          </li>
-          <li>
+            </li>
+            <li>
 Run steps 6 through 11 of the algorithm in
 section <a href="#verify-proof"></a>; if no exceptions are raised, associate
 the <var>isProofVerified</var> value with this <var>proof</var>.
-            <p class="note">
+              <p class="note">
 We specifically use the value of <var>unsecuredDocument</var> as set above,
 rather than using the value from step 5 of the algorithm in section
 <a href="#verify-proof"></a>.
-            </p>
-          </li>
+              </p>
+            </li>
+          </ol>
           <li>
 Return the <var>allProofs</var> along with each proof's associated
 <var>isProofVerified</var> information.
--- a/index.html
+++ b/index.html
@@ -2990,9 +2990,9 @@ convey an error type of
           <li>
 Set <var>unsecuredDocument.proof</var> to <var>matchingProofs</var>.
             <p class="note">
-This step is critical to <q>bind</q> the previous proofs to the document prior
-to signing. The <var>proof</var> value for the document will be updated in a
-later step of this algorithm.
+This step is critical, as it <q>binds</q> the previous proofs to the document
+prior to signing. The <var>proof</var> value for the document will be updated
+in a later step of this algorithm.
             </p>
           </li>
           <li>
--- a/index.html
+++ b/index.html
@@ -3107,11 +3107,14 @@ Return <var>isProofVerified</var> as the <a>verification result</a>.
 In a proof set or proof chain, a <a>secured data document</a> has a
 <code>proof</code> attribute which contains a list of proofs
 (<var>allProofs</var>).
-The following algorithm specifies how to check the authenticity and
-integrity of a <a>secured data document</a> by verifying each
-proof in the <var>allProofs</var>. Required inputs are a
+The following algorithm gives one method to check the authenticity and
+integrity of a <a>secured data document</a>. It does this by verifying every
+proof in <var>allProofs</var>. Other approaches are possible, particularly if
+it is desired to only verify a subset of the proofs contained in
+<var>allProfs</var>. Required inputs are a
 <a>secured data document</a> (<var>securedDocument</var>). A list of
-verification results is produced as output.
+verification results corresponding to each proof in <var>allProofs</var> is
+produced as output.
         </p>
         <ol class="algorithm">
           <li>
--- a/index.html
+++ b/index.html
@@ -2983,7 +2983,9 @@ attribute is a string, add the element from <var>allProofs</var> with an
 <a href="#PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</a>.
 If the <code>previousProof</code> attribute is an array, add each element from
 <var>allProofs</var> with an <code>id</code> attribute that matches an element
-of that array. If any matches are missing, an error MUST be raised and SHOULD
+of that array. If any elements of <code>previousProof</code> array have an
+<code>id</code> attribute that does not match the <code>id</code> attribute
+of any element of <var>allProofs</var>, an error MUST be raised and SHOULD
 convey an error type of
 <a href="#PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</a>.
           </li>
--- a/index.html
+++ b/index.html
@@ -2971,15 +2971,35 @@ with the <var>proof</var> attribute removed. Let <var>output</var> be a copy of
 the <var>unsecuredDocument</var>.
           </li>
           <li>
+Let <var>matchingProofs</var> be an empty set.
+          </li>
+          <li>
 If <var>options</var> contains a <code>previousProof</code> attribute and that
-attribute is a string, check whether
-an element of the <var>allProofs</var> has a matching <code>id</code> attribute.
-If not, an error MUST be raised and SHOULD convey an error type of
-<a href="#PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</a>. If the
-<code>previousProof</code> attribute is an array, check that each element of that
-array is a string which matches the <code>id</code> attribute of an element
-of the <var>allProofs</var>. If not, an error MUST be raised and SHOULD convey
-an error type of <a href="#PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</a>.
+attribute is a string, add the element from <var>allProofs</var> with a matching
+<code>id</code> attribute to matchingProofs. If it does not exist, an error MUST
+be raised and SHOULD convey an error type of
+<a href="#PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</a>.
+If the <code>previousProof</code> attribute is an array, add an element from
+<var>allProofs</var> with an <code>id</code> attribute that matches each element
+of that array. If any matches are missing, an error MUST be raised and SHOULD
+convey an error type of
+<a href="#PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</a>.
+          </li>
+          <li>
+For each element <var>mp</var> of <var>matchingProofs</var>, if there are any,
+recursively run the previous step on <var>mp.previousProof</var> if it exist.
+            <p class="note">
+This steps insures that all proof dependencies are added regardless of the depth
+of the proof chain.
+            </p>
+          </li>
+          <li>
+Set <var>unsecuredDocument.proof</var> to <var>matchingProofs</var>.
+            <p class="note">
+This step is critical to <q>bind</q> the previous proofs to the document prior
+to signing. The <var>proof</var> value for the document will be updated in a
+later step of this algorithm.
+            </p>
           </li>
           <li>
 Run steps 2 through 8 of the algorithm in section <a href="#add-proof"></a>. If
--- a/index.html
+++ b/index.html
@@ -3122,21 +3122,48 @@ verification results is produced as output.
 Set <var>allProofs</var> to </var><var>securedDocument</var>.<var>proof</var>.
           </li>
           <li>
-For each <var>proof</var> in <var>allProofs</var>, run steps 2 through 11 of the algorithm in
-section <a href="#verify-proof"></a>; if no exceptions are raised, associate
-the <var>isProofVerified</var> value with this <var>proof</var>.
+For each <var>proof</var> in <var>allProofs</var> do the following steps:
           </li>
           <li>
-For each <var>proof</var> in <var>allProofs</var> that contains a
-<code>previousProof</code> attribute, modify the associated
-<var>isProofVerified</var> by iteratively checking that the proof identified by
-the <code>id</code> value in <var>previousProof</var>, along with every
-<code>previousProof</code> up the chain, is valid. If <code>previousProof</code> is an
-array, then all previous proofs identified by the array elements
-MUST be valid for the current proof to be considered valid. Cycle detection,
-such as keeping a list of previously visited proofs, MUST be used to guard
-against proof cycles. If a cycle is detected in a proof chain,
-then a `PROOF_CHAIN_CYCLE_ERROR` MUST be raised.
+Let <var>matchingProofs</var> be an empty set.
+          </li>
+          <li>
+If <var>proof</var> contains a <code>previousProof</code> attribute and that
+attribute is a string, add the element from <var>allProofs</var> with a matching
+<code>id</code> attribute to matchingProofs. If it does not exist, an error MUST
+be raised and SHOULD convey an error type of
+<a href="#MALFORMED_PROOF_ERROR">MALFORMED_PROOF_ERROR</a>.
+If the <code>previousProof</code> attribute is an array, add an element from
+<var>allProofs</var> with an <code>id</code> attribute that matches each element
+of that array. If any matches are missing, an error MUST be raised and SHOULD
+convey an error type of
+<a href="#MALFORMED_PROOF_ERROR">MALFORMED_PROOF_ERROR</a>.
+          </li>
+          <li>
+For each element <var>mp</var> of <var>matchingProofs</var>, if there are any,
+recursively run the previous step on <var>mp.previousProof</var> if it exist.
+Cycle detection, such as keeping a list of previously visited proofs, MUST be
+used to guard against proof cycles. If a cycle is detected in a proof chain,
+then a <a href="#PROOF_CHAIN_CYCLE_ERROR">PROOF_CHAIN_CYCLE_ERROR</a> MUST be
+raised.
+          </li>
+          <li>
+Let <var>unsecuredDocument</var> be a copy of <var>securedDocument</var> with
+the proof value removed and then set <var>unsecuredDocument.proof</var>
+to <var>matchingProofs</var>.
+          <li>
+Run steps 2 through 4 to validate the contents of <var>proof</var>.
+          </li>
+          </li>
+          <li>
+Run steps 6 through 11 of the algorithm in
+section <a href="#verify-proof"></a>; if no exceptions are raised, associate
+the <var>isProofVerified</var> value with this <var>proof</var>.
+            <p class="note">
+We specifically use the value of <var>unsecuredDocument</var> as set above
+rather than using value from step 5 of the algorithm in section
+<a href="#verify-proof"></a>
+            </p>
           </li>
           <li>
 Return the <var>allProofs</var> along with each proof's associated
@@ -3329,6 +3356,11 @@ The <a>verification method</a> in a <a>controller document</a> was not
 associated using the expected <a>verification relationship</a> as expressed
 in the `proofPurpose` property in the proof. See Section
 <a href="#retrieve-verification-method"></a>.
+          </dd>
+          <dt id="PROOF_CHAIN_CYCLE_ERROR">PROOF_CHAIN_CYCLE_ERROR (-26)</dt>
+          <dd>
+A proof cycle was detected during the verification of a proof set or proof
+chain. See Section <a href="#verify-proof-sets-and-chains"></a>.
           </dd>
         </dl>
       </section>
--- a/index.html
+++ b/index.html
@@ -3107,11 +3107,11 @@ Return <var>isProofVerified</var> as the <a>verification result</a>.
 In a proof set or proof chain, a <a>secured data document</a> has a
 <code>proof</code> attribute which contains a list of proofs
 (<var>allProofs</var>).
-The following algorithm gives one method to check the authenticity and
-integrity of a <a>secured data document</a>. It does this by verifying every
+The following algorithm provides one method of checking the authenticity and
+integrity of a <a>secured data document</a>, achieved by verifying every
 proof in <var>allProofs</var>. Other approaches are possible, particularly if
-it is desired to only verify a subset of the proofs contained in
-<var>allProfs</var>. Required inputs are a
+it is only desired to verify a subset of the proofs contained in
+<var>allProfs</var>. Required input is a
 <a>secured data document</a> (<var>securedDocument</var>). A list of
 verification results corresponding to each proof in <var>allProofs</var> is
 produced as output.
--- a/index.html
+++ b/index.html
@@ -3125,7 +3125,8 @@ Let <var>matchingProofs</var> be an empty list.
             <li>
 If <var>proof</var> contains a <code>previousProof</code> attribute and that
 attribute is a string, add the element from <var>allProofs</var> with an
-<code>id</code> attribute matching <code>previousProof</code> to matchingProofs.
+<code>id</code> attribute matching <code>previousProof</code> to
+<code>matchingProofs</code>.
 If a proof with <code>id</code> does not exist in <var>allProofs</var>, an error
 MUST be raised and SHOULD convey an error type of
 <a href="#MALFORMED_PROOF_ERROR">MALFORMED_PROOF_ERROR</a>.
--- a/index.html
+++ b/index.html
@@ -3111,7 +3111,13 @@ The following algorithm provides one method of checking the authenticity and
 integrity of a <a>secured data document</a>, achieved by verifying every
 proof in <var>allProofs</var>. Other approaches are possible, particularly if
 it is only desired to verify a subset of the proofs contained in
-<var>allProfs</var>. Required input is a
+<var>allProfs</var>. If another approach is taken to verify only a subset of the
+proofs, then it is important to note any proof with <code>previousProof</code>
+in that subset can only be considered verified if the proofs it references are
+also considered verified.
+        </p>
+        <p>
+Required input is a
 <a>secured data document</a> (<var>securedDocument</var>). A list of
 verification results corresponding to each proof in <var>allProofs</var> is
 produced as output.
--- a/index.html
+++ b/index.html
@@ -3134,7 +3134,9 @@ MUST be raised and SHOULD convey an error type of
 <a href="#MALFORMED_PROOF_ERROR">MALFORMED_PROOF_ERROR</a>.
 If the <code>previousProof</code> attribute is an array, add each element from
 <var>allProofs</var> with an <code>id</code> attribute that matches an element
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2971,15 +2971,35 @@ <h3>Add Proof Set/Chain</h3>
...
-attribute is a string, check whether
-an element of the <var>allProofs</var> has a matching <code>id</code> attribute.
-If not, an error MUST be raised and SHOULD convey an error type of
-<a href="#PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</a>. If the
-<code>previousProof</code> attribute is an array, check that each element of that
-array is a string which matches the <code>id</code> attribute of an element
-of the <var>allProofs</var>. If not, an error MUST be raised and SHOULD convey
-an error type of <a href="#PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</a>.
+attribute is a string, add the element from <var>allProofs</var> with a matching
+<code>id</code> attribute to matchingProofs. If it does not exist, an error MUST
```
TallTed: What is the "it" that might not exist?

index.html
```diff
@@ -2971,15 +2971,35 @@ <h3>Add Proof Set/Chain</h3>
...
+<var>allProofs</var> with an <code>id</code> attribute that matches each element
+of that array. If any matches are missing, an error MUST be raised and SHOULD
+convey an error type of
+<a href="#PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</a>.
+          </li>
+          <li>
+For each element <var>mp</var> of <var>matchingProofs</var>, if there are any,
+recursively run the previous step on <var>mp.previousProof</var> if it exist.
+            <p class="note">
+This steps insures that all proof dependencies are added regardless of the depth
```
```suggestion
recursively run the previous step on <var>mp.previousProof</var> if it exists.
            <p class="note">
This step insures that all proof dependencies are added, regardless of the depth
```

index.html
```diff
@@ -3102,21 +3122,48 @@ <h3>Verify Proof Sets and Chains</h3>
 Set <var>allProofs</var> to </var><var>securedDocument</var>.<var>proof</var>.
           </li>
           <li>
-For each <var>proof</var> in <var>allProofs</var>, run steps 2 through 11 of the algorithm in
-section <a href="#verify-proof"></a>; if no exceptions are raised, associate
-the <var>isProofVerified</var> value with this <var>proof</var>.
+For each <var>proof</var> in <var>allProofs</var> do the following steps:
```
```suggestion
For each <var>proof</var> in <var>allProofs</var>, do the following steps:
```

index.html
```diff
@@ -3102,21 +3122,48 @@ <h3>Verify Proof Sets and Chains</h3>
...
-MUST be valid for the current proof to be considered valid. Cycle detection,
-such as keeping a list of previously visited proofs, MUST be used to guard
-against proof cycles. If a cycle is detected in a proof chain,
-then a `PROOF_CHAIN_CYCLE_ERROR` MUST be raised.
+Let <var>matchingProofs</var> be an empty set.
+          </li>
+          <li>
+If <var>proof</var> contains a <code>previousProof</code> attribute and that
+attribute is a string, add the element from <var>allProofs</var> with a matching
+<code>id</code> attribute to matchingProofs. If it does not exist, an error MUST
```
TallTed: Again, what is the "it" that might not exist?
```suggestion
<code>id</code> attribute to matchingProofs. If it does not exist, an error MUST
```

index.html
```diff
@@ -3102,21 +3122,48 @@ <h3>Verify Proof Sets and Chains</h3>
...
+<a href="#MALFORMED_PROOF_ERROR">MALFORMED_PROOF_ERROR</a>.
+If the <code>previousProof</code> attribute is an array, add an element from
+<var>allProofs</var> with an <code>id</code> attribute that matches each element
+of that array. If any matches are missing, an error MUST be raised and SHOULD
+convey an error type of
+<a href="#MALFORMED_PROOF_ERROR">MALFORMED_PROOF_ERROR</a>.
+          </li>
+          <li>
+For each element <var>mp</var> of <var>matchingProofs</var>, if there are any,
+recursively run the previous step on <var>mp.previousProof</var> if it exist.
```
```suggestion
recursively run the previous step on <var>mp.previousProof</var> if it exists.
```

index.html
```diff
@@ -3102,21 +3122,48 @@ <h3>Verify Proof Sets and Chains</h3>
...
+          </li>
+          </li>
+          <li>
+Run steps 6 through 11 of the algorithm in
+section <a href="#verify-proof"></a>; if no exceptions are raised, associate
+the <var>isProofVerified</var> value with this <var>proof</var>.
+            <p class="note">
+We specifically use the value of <var>unsecuredDocument</var> as set above
+rather than using value from step 5 of the algorithm in section
+<a href="#verify-proof"></a>
```
```suggestion
We specifically use the value of <var>unsecuredDocument</var> as set above,
rather than using the value from step 5 of the algorithm in section
<a href="#verify-proof"></a>.
```

TallTed (changes requested):
small typos, punctuation, grammar...

index.html
```diff
@@ -2971,15 +2971,35 @@ <h3>Add Proof Set/Chain</h3>
...
-<code>previousProof</code> attribute is an array, check that each element of that
-array is a string which matches the <code>id</code> attribute of an element
-of the <var>allProofs</var>. If not, an error MUST be raised and SHOULD convey
-an error type of <a href="#PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</a>.
+attribute is a string, add the element from <var>allProofs</var> with a matching
+<code>id</code> attribute to matchingProofs. If it does not exist, an error MUST
+be raised and SHOULD convey an error type of
+<a href="#PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</a>.
+If the <code>previousProof</code> attribute is an array, add an element from
+<var>allProofs</var> with an <code>id</code> attribute that matches each element
```
dlongley: I think this reads slightly better:

```suggestion
If the <code>previousProof</code> attribute is an array, add each element from
<var>allProofs</var> with an <code>id</code> attribute that matches an element
```

index.html
```diff
@@ -2971,15 +2971,35 @@ <h3>Add Proof Set/Chain</h3>
...
+of that array. If any matches are missing, an error MUST be raised and SHOULD
+convey an error type of
+<a href="#PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</a>.
+          </li>
+          <li>
+For each element <var>mp</var> of <var>matchingProofs</var>, if there are any,
+recursively run the previous step on <var>mp.previousProof</var> if it exist.
+            <p class="note">
+This steps insures that all proof dependencies are added regardless of the depth
+of the proof chain.
```
dlongley: So I don't think we need to run this step recursively. This is because any proof dependency will have already signed over the contents of its own dependencies. For example, if we have a proof chain of A => B => C, then B will have signed over A such that when C is created, it only needs to sign over B, not A and B.

Wind4Greg: Let me write up the security intent and considerations (attack model) for this first with some examples to be sure. If we don't it will be a nice simplification.

Wind4Greg: Hi @dlongley, here is my example analysis and agrees with your notion that the recursive part is unneeded. Please take a look for holes or other attack scenarios.

## Proof Chains and Sets

Instead of a single *proof* (proof options with *proofValue*) included in the signed VC we can have an array of them. We call such a thing a *proof set*. Each one of which should verify with the appropriate keying material.

The *holder* (or intermediary) receiving a signed VC containing a *proof set* is able to remove any of the *proofs* within the set prior to passing it on to another entity.

We can introduce a "dependency relation" between *proofs* in a *proof set* by assigning the proof that another proof relies on, i.e., the *dependent proof*, an *id* and the *relying proof* will have a list (or single item) denoting *previousProofs*. Such *dependency chains* can have arbitrary depth. The **intent** of such a *proof chain* is to model an approval chain in a business process or a notary witnessing analog signatures. In such cases we CANNOT allow the *proofs* that are relied upon to be **substituted** with other proofs or removed.

Example: The CEO signed off on a PO (document), based on the reasonableness of the item being requested and the fact that the CFO and the VP of Engineering signed off on the PO. Suppose we have a proof set:

```javascript
proofs = [
  {id: 0, pubKey: pubVPEng,  previusProof: null, proofValue: signature(VP_Eng, unsignedDocument)},
  {id: 1, pubKey: pubCFO, previousProof: 0, proofValue: signature(privCFO, unsignedDocument)},
  {id: 2, pubKey: pubCEO, previousProof: 1, proofValue: signature(privCEO, unsignedDocument)}
]
```

Where *pubX* denotes a public key and *privX* denotes the corresponding private key, and `signature(privKey, doc)` denotes a signature algorithm.

A proof chain such as this provides **no** protection from the recipient swapping out *dependent* signatures as in the following example:

```javascript
[
  {id: 1, pubKey: pubHacker, previousProof: null, proofValue: signature(privHacker, unsignedDocument)},
  {id: 2, pubKey: pubCEO, previousProof: 2, proofValue: signature(privCEO, unsignedDocument)}
]
```

Hence in the chain case it is very important to "bind" the previous proofs in the chain such as in the following.

```javascript
[
  {id: 1, pubKey: pubVPEng, previousProof: null, proofValue: signature(privVPEng, unsignedDocument)},
  {id: 2, pubKey: pubCFO, previousProof: 1, proofValue: signature(privCFO, unsignedDocument + proofs[0])},
  {id: 3, pubKey: pubCEO, previousProof: 2, proofValue: signature(privCEO, unsignedDocument + proofs[1])}
]
```

Now when verifying we see that proof[0] depends on the proper signing by VPEng and integrity of the unsigned document. We see that proofs[1] depends on the proper signing by the CFO and the integrity of the unsigned document and proofs[0], i.e., binds `{id: 1, pubKey: pubVPEng, previousProof: null, proofValue: signature(VP_Eng, unsignedDocument)}` with the document so it cannot be removed or swapped out. For proofs[2] we see that it depends on the proper signing by the CEO and the integrity of the unsigned document and proofs[1] (which in turn depends on document and proofs[0]). Hence the above scheme with the additional binding guards the integrity of the proof chain.

index.html
Wind4Greg: Will be doing a bunch of other changes and will try to incorporate all your suggestions then ping you to re-review. Cheers Greg

index.html
TallTed: This analysis should probably be put somewhere it will be (easily) visible after this conversation is resolved.

index.html
TallTed: It is MUCH more challenging and time consuming to review and check for all suggested changes if you don't use the GitHub "commit" or "add to batch" tools on each suggestion. I now need to search out the location of each of my comments, to see what the changes there were, as well as re-reviewing the whole document. Your [single giant commit](https://github.com/w3c/vc-data-integrity/pull/224/commits/d65d2b303aa15b837a6090c1c881caeee7e608f7) is almost a new PR, but without the same degree of reviewer accessibility.

index.html
```diff
@@ -3122,32 +3116,25 @@ <h3>Verify Proof Sets and Chains</h3>
 Set <var>allProofs</var> to </var><var>securedDocument</var>.<var>proof</var>.
           </li>
           <li>
-For each <var>proof</var> in <var>allProofs</var> do the following steps:
+For each <var>proof</var> in <var>allProofs</var>, do the following steps:
           </li>
           <li>
```
TallTed: The steps in the loop should be in a nested `<ol>` below the "do the following steps:", not at the same level.

index.html
Wind4Greg: Where? With example near definition? Security considerations section?

index.html
Wind4Greg: I need to use a real editing environment (code editor) to make significant changes to raw HTML. Yes, it is like a new PR that is why requested a re-review. Note that Git commit is used and the audit trail is intact.

index.html
Wind4Greg: Implemented and asked for re-review.

index.html
```diff
@@ -3102,22 +3116,44 @@ <h3>Verify Proof Sets and Chains</h3>
...
-For each <var>proof</var> in <var>allProofs</var> that contains a
-<code>previousProof</code> attribute, modify the associated
-<var>isProofVerified</var> by iteratively checking that the proof identified by
-the <code>id</code> value in <var>previousProof</var>, along with every
-<code>previousProof</code> up the chain, is valid. If <code>previousProof</code> is an
-array, then all previous proofs identified by the array elements
-MUST be valid for the current proof to be considered valid. Cycle detection,
-such as keeping a list of previously visited proofs, MUST be used to guard
-against proof cycles. If a cycle is detected in a proof chain,
-then a `PROOF_CHAIN_CYCLE_ERROR` MUST be raised.
```
dlongley, line 3119: EDIT: Please ignore and look at my other review...

<strike>So I believe we still need some of this text. In short, each of the proofs referenced (by `id`) in `previousProof` need to be verified prior to checking the current proof. And if a cycle is detected when doing this checking, the error mentioned must be raised (I see we have text that adds a definition for it below, but without this text there would nothing that would throw it either).

So I expect we can simplify the above text considerably, but we can't remove it entirely. We can probably insert some text above (I'll try to add a suggestion or comment above for that).

</strike>

index.html
```diff
@@ -3102,22 +3116,44 @@ <h3>Verify Proof Sets and Chains</h3>
...
+If a proof with <code>id</code> does not exist in <var>allProofs</var>, an error
+MUST be raised and SHOULD convey an error type of
+<a href="#MALFORMED_PROOF_ERROR">MALFORMED_PROOF_ERROR</a>.
+If the <code>previousProof</code> attribute is an array, add each element from
+<var>allProofs</var> with an <code>id</code> attribute that matches an element
+of that array. If any matches are missing, an error MUST be raised and SHOULD
+convey an error type of
+<a href="#MALFORMED_PROOF_ERROR">MALFORMED_PROOF_ERROR</a>.
+            </li>
+            <li>
```
dlongley, line 3152: EDIT: Please ignore and look at my other review...

<strike>See my other comment about needing to keep some text about verifying previous proofs before reading more here.

Now, once we have the `matchingProofs` here, we need to check that each of them has been verified already. I recommend we express this somewhat simply as we did before (see text currently marked to be removed below near my other comment) and allow implementers room to decide how to best ensure each previous proof was already verified (e.g., recursively, with a non-recursive scanning algorithm, etc.), but state that if any cycles are detected the proof cycle error must be thrown as before.

If we wanted to be prescriptive, one way to do this would be to use a map to populate with keys that are matching proof IDs and values that keep track of whether the proof is currently being verified or has been verified (absence in that map means the proof hasn't been seen at all yet). This map would be passed along / shared during verification of all related proofs, being initialized only at the outset (if it wasn't passed in as an option). We would then iterate through each matching proof and, if the proof ID was already in the map and not verified, we would throw the cycle error, otherwise we would add it and recurse into this algorithm, passing the map. Each proof, upon verification, would have its entry updated to indicate as such -- and a verification failure would terminate the whole algorithm.

But, again, I think it's better to allow implementations to do the above part however they want to.
</strike>

dlongley (changes requested):

index.html
dlongley: Actually, ignore this comment -- I see that a different, simpler change is possible, I'll do another review.

index.html
dlongley: Actually, ignore this comment -- I see that a different, simpler change is possible, I'll do another review.

index.html
```diff
@@ -3310,6 +3346,11 @@ <h3>Processing Errors</h3>
 in the `proofPurpose` property in the proof. See Section
 <a href="#retrieve-verification-method"></a>.
           </dd>
+          <dt id="PROOF_CHAIN_CYCLE_ERROR">PROOF_CHAIN_CYCLE_ERROR (-26)</dt>
+          <dd>
+A proof cycle was detected during the verification of a proof set or proof
+chain. See Section <a href="#verify-proof-sets-and-chains"></a>.
+          </dd>
```
dlongley: I didn't see that we're already iterating over `allProofs` when doing proof chain checking. I was expecting that we would be picking out the proof we wanted to check as the starting point instead. This eliminates the need for my other comments above -- and, I believe, eliminates the need for this error as well. I don't think the cycles can happen without a proof failing to verify given the other changes we've made.

However, I do think we should modify this so that we don't have to process `allProofs`, as it places the burden on the presenter to avoid including any proofs that might not be accepted by the verifier, but that would not stop the verifier from being able to verify a particular proof (and its dependencies) that it does accept.

I think we need to consider how to do this, perhaps by keeping track of the dependencies as we go and allowing for the proof status to be verified / not verified / not considered (a third value). The "not considered" value would be perfectly fine if the proof was unrelated to the target proof that is being considered for verification.

dlongley (changes requested):

index.html
TallTed: I meant somewhere in GitHub, probably on this PR, but not attached to a [change request](https://github.com/w3c/vc-data-integrity/pull/224/files) where comments disappear once the request is marked as "resolved"; rather, in the [main "conversation" thread](https://github.com/w3c/vc-data-integrity/pull/224).

If others think it's worth including this analysis in the spec, I'll let them suggest where (perhaps an appendix, perhaps near the text on which you've commented, perhaps elsewhere).

index.html
TallTed: @dlongley -- Could you `<strike>` the comment(s) that should be ignored? This will help save us the time of reading those comments before we get to the "ignore" flag. :-)

index.html
TallTed: @dlongley -- Could you `<strike>` the comment(s) that should be ignored? This will help save us the time of reading those comments before we get to the "ignore" flag. :-)

index.html
```diff
@@ -2971,15 +2971,29 @@ <h3>Add Proof Set/Chain</h3>
...
-an error type of <a href="#PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</a>.
+attribute is a string, add the element from <var>allProofs</var> with an
+<code>id</code> attribute matching <code>previousProof</code> to
+<var>matchingProofs</var>. If a proof with
+<code>id</code> equal to <code>previousProof</code>does not exist in
+<var>allProofs</var>, an error MUST be raised and SHOULD convey an error type of
+<a href="#PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</a>.
+If the <code>previousProof</code> attribute is an array, add each element from
+<var>allProofs</var> with an <code>id</code> attribute that matches an element
+of that array. If any matches are missing, an error MUST be raised and SHOULD
```
TallTed: `If any matches are missing,` is unclear to me. Maybe something like this --
```suggestion
of that array. If any elements of <code>previousProof</code> array have an
<code>id</code> attribute that does not match the <code>id</code> attribute
of any element of <var>allProofs</var>, an error MUST be raised and SHOULD
```

index.html
```diff
@@ -2971,15 +2971,29 @@ <h3>Add Proof Set/Chain</h3>
...
+convey an error type of
+<a href="#PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</a>.
+          </li>
+          <li>
+Set <var>unsecuredDocument.proof</var> to <var>matchingProofs</var>.
+            <p class="note">
+This step is critical to <q>bind</q> the previous proofs to the document prior
+to signing. The <var>proof</var> value for the document will be updated in a
+later step of this algorithm.
```
```suggestion
This step is critical, as it <q>binds</q> the previous proofs to the document
prior to signing. The <var>proof</var> value for the document will be updated
in a later step of this algorithm.
```

index.html
```diff
@@ -3102,22 +3116,44 @@ <h3>Verify Proof Sets and Chains</h3>
...
+For each <var>proof</var> in <var>allProofs</var>, do the following steps:
+          </li>
+          <ol class="algorithm">
+            <li>
+Let <var>matchingProofs</var> be an empty list.
+            </li>
+            <li>
+If <var>proof</var> contains a <code>previousProof</code> attribute and that
+attribute is a string, add the element from <var>allProofs</var> with an
+<code>id</code> attribute matching <code>previousProof</code> to matchingProofs.
```
```suggestion
<code>id</code> attribute matching <code>previousProof</code> to
<code>matchingProofs</code>.
```

index.html
```diff
@@ -3102,22 +3116,44 @@ <h3>Verify Proof Sets and Chains</h3>
...
+If a proof with <code>id</code> does not exist in <var>allProofs</var>, an error
+MUST be raised and SHOULD convey an error type of
+<a href="#MALFORMED_PROOF_ERROR">MALFORMED_PROOF_ERROR</a>.
+If the <code>previousProof</code> attribute is an array, add each element from
+<var>allProofs</var> with an <code>id</code> attribute that matches an element
+of that array. If any matches are missing, an error MUST be raised and SHOULD
```
TallTed: As [above](https://github.com/w3c/vc-data-integrity/pull/224/files#r1417842986).

index.html
Wind4Greg: I'm inclined towards the security considerations section. That way its easy for cryptographic reviewers to find.

index.html
Wind4Greg: What do you want changed? This has been incorporated in a recent commit.

index.html
dlongley: @TallTed, +1, struck here and in the other comment.

dlongley (approved):
I'm happy for us to merge this (once other reviewers are also satisfied) and come up with a separate tweak that covers my comments here: https://github.com/w3c/vc-data-integrity/pull/224#discussion_r1416148731

index.html
Wind4Greg: Hi Dave (@dlongley), removed the PROOF_CHAIN_CYCLE  error.

As for your other comment. Currently we say: "For each proof in allProofs, do the following steps: ". Couldn't we just modify this one line to something like

"For each proof in the secured document to be verified, do the following steps:"

That way no one is forced to verify all the proofs and the algorithm still works as it stands.

index.html
```diff
@@ -2971,15 +2971,31 @@ <h3>Add Proof Set/Chain</h3>
...
-an error type of <a href="#PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</a>.
+attribute is a string, add the element from <var>allProofs</var> with an
+<code>id</code> attribute matching <code>previousProof</code> to
+<var>matchingProofs</var>. If a proof with
+<code>id</code> equal to <code>previousProof</code>does not exist in
+<var>allProofs</var>, an error MUST be raised and SHOULD convey an error type of
+<a href="#PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</a>.
+If the <code>previousProof</code> attribute is an array, add each element from
+<var>allProofs</var> with an <code>id</code> attribute that matches an element
+of that array. If any elements of <code>previousProof</code> array have an
```
TallTed: Fixing my earlier mistake --
```suggestion
of that array. If any element of <code>previousProof</code> array has an
```

index.html
dlongley: @Wind4Greg,

> "For each proof in the secured document to be verified, do the following steps:"...

Well, I think this implies that the verifier knows those proofs to begin with when I would expect them to only know that there's a particular proof they want to verify ... which itself allows the verifier to discover what others need verification (through `previousProof`).

We might need to have a previous step that will walk through `previousProof` recursively and build a list of proofs to verify -- stopping recursion whenever we see a proof we've already seen. We don't need to throw a cycle error this way, and we won't get stuck in a loop, and we will only return that verification passed if each of those proofs verifies.

So I think we need some (ideally short) text that covers this simply enough that can be put as a starting step to gather all the proofs that need verification (vs. just assuming the verifier knows what they will be) and then we can make the statement you're saying without issue, I'd think. Then we'd be good here, IMO.

index.html
TallTed: "A recent commit" is difficult to review, at best. Please provide a direct link, when referencing such.

Here's the complete change request. It's very similar to [this change request](https://github.com/w3c/vc-data-integrity/pull/224#discussion_r1417842986), but impacts line 3137 instead of line 2986. GitHub is not properly highlighting most of the insert. Note that it replaces `If any matches are missing`, which remains unclear to me, with `If any element of previousProof array has an id attribute that does not match the id attribute of any element of allProofs`, modulo tags. --
```suggestion
of that array. If any element of <code>previousProof</code> array has an
<code>id</code> attribute that does not match the <code>id</code> attribute
of any element of <var>allProofs</var>, an error MUST be raised and SHOULD
```

index.html
Wind4Greg: Oops your right @dlongley. Let me try out a couple things to see if we can avoid using recursion to create the list of proofs to be check up front.

Wind4Greg: @dlongley This is starting to seem like an optimization. I don't have any examples of elaborate proof sets/chains with chains that don't depend on some of the proofs to test against. If you have a suggested algorithm change let me know. I'm leaning on putting some text at the beginning to let them know that they can optimize of just produce the recursively generated list at the beginning.

dlongley: @Wind4Greg,

> I'm leaning on putting some text at the beginning to let them know that they can optimize of just produce the recursively generated list at the beginning.

I'd say go ahead and add some text to this effect and we'll see what reviewers think. It's not core to the algorithm and the `previousProof` indicates how you can find out which proofs are related -- implementers can process this however they like to obtain the proofs that need to be verified, but they must do so to ensure that the whole chain is properly verified.

dlongley: A test that will fail verification if one of the (deep) dependency proofs is left out would be a good way to ensure implementers are doing this this check.

Wind4Greg: @dlongley I just added intro text to the chain verification algorithm stating that the algorithm given is one approach and that others are possible, particularly if only subset of the `allProofs` needs to be verified.

index.html
```diff
@@ -3091,33 +3107,61 @@ <h3>Verify Proof Sets and Chains</h3>
...
 <code>proof</code> attribute which contains a list of proofs
 (<var>allProofs</var>).
-The following algorithm specifies how to check the authenticity and
-integrity of a <a>secured data document</a> by verifying each
-proof in the <var>allProofs</var>. Required inputs are a
+The following algorithm gives one method to check the authenticity and
+integrity of a <a>secured data document</a>. It does this by verifying every
+proof in <var>allProofs</var>. Other approaches are possible, particularly if
+it is desired to only verify a subset of the proofs contained in
+<var>allProfs</var>. Required inputs are a
```
```suggestion
The following algorithm provides one method of checking the authenticity and
integrity of a <a>secured data document</a>, achieved by verifying every
proof in <var>allProofs</var>. Other approaches are possible, particularly if
it is only desired to verify a subset of the proofs contained in
<var>allProfs</var>. Required input is a
```

dlongley: `allProfs` => `allProofs` typo fix could be added to this suggestion at the end.

index.html
```diff
@@ -3091,33 +3107,61 @@ <h3>Verify Proof Sets and Chains</h3>
...
 (<var>allProofs</var>).
-The following algorithm specifies how to check the authenticity and
-integrity of a <a>secured data document</a> by verifying each
-proof in the <var>allProofs</var>. Required inputs are a
+The following algorithm provides one method of checking the authenticity and
+integrity of a <a>secured data document</a>, achieved by verifying every
+proof in <var>allProofs</var>. Other approaches are possible, particularly if
+it is only desired to verify a subset of the proofs contained in
+<var>allProfs</var>. Required input is a
 <a>secured data document</a> (<var>securedDocument</var>). A list of
```
dlongley, line 3121: I do think we should add another sentence here that says that "If another approach is taken to verify only a subset of the proofs, then it is important to note any proof with `previousProof` in that subset can only be considered verified if the proofs it references are also considered verified." Then we should be good.

dlongley (approved):
Approving with the caveat comment I mentioned.

index.html
Wind4Greg: Good point @dlongley . Added your text. See update.

index.html
```diff
@@ -3111,7 +3111,13 @@ <h3>Verify Proof Sets and Chains</h3>
 integrity of a <a>secured data document</a>, achieved by verifying every
 proof in <var>allProofs</var>. Other approaches are possible, particularly if
 it is only desired to verify a subset of the proofs contained in
-<var>allProfs</var>. Required input is a
+<var>allProfs</var>. If another approach is taken to verify only a subset of the
+proofs, then it is important to note any proof with <code>previousProof</code>
+in that subset can only be considered verified if the proofs it references are
+also considered verified.
```
```suggestion
<var>allProofs</var>. If another approach is taken to verify only a subset of the
proofs, then it is important to note that any proof in that subset with a
<code>previousProof</code> can only be considered verified if the proofs it
references are also considered verified.
```

TallTed (changes requested):
a little rephrasing for clarity

msporny (approved):

msporny:

PR #226, opened 2023-12-08
Define algorithm input and output types.

jyasskin:
Both the transform, hash, and proof algorithms that cryptographic suites have to define, and this
spec's Verify Proof interface.

This removes the underspecified "proof options" field in favor of saying they're members of the `proof` object for verification, or defined by the input cryptographic suite for signing.

I started this on my way to trying to have "verify" also return the verified document", but I decided this was a big enough change to start with.

Apologies for any obvious mistakes I've made here.

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/vc-data-integrity/pull/226.html" title="Last updated on Jan 2, 2024, 7:58 PM UTC (e7a445f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/226/3140804...jyasskin:e7a445f.html" title="Last updated on Jan 2, 2024, 7:58 PM UTC (e7a445f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2456,8 +2456,15 @@ verification of the data.
             <dd>A [=hashing algorithm=] that takes a [=JSON object=] as input and returns a [=byte
             sequence=].</dd>
             <dt><dfn>proof creation options</dfn></dt>
-            <dd>A [=struct=] type that needs to be supplied to the [=cryptographic suite/serialize
-            proof=] operation.</dd>
+            <dd>
+              <p>A [=struct=] type that needs to be supplied to the [=cryptographic suite/serialize
+              proof=] operation.</p>
+              <p class="note">To sign data, cryptographic suites need to be able to use a private
+              key. This private key will often be implied by a field in the [=proof creation
+              options=], but it is often a better design to avoid passing the key directly. Passing
+              a key identifier or an asynchronous signing algorithm allows implementations to take
+              advantage of hardware security modules (HSMs) to get better security.</p>
+            </dd>
             <dt><dfn>serialize proof</dfn></dt>
             <dd>A [=proof serialization algorithm=] that takes a [=byte sequence=] and an instance
             of this suite's [=proof creation options=] type as input and returns a [=JSON object=]
--- a/index.html
+++ b/index.html
@@ -2442,6 +2442,33 @@ parameters, and other necessary details used to perform cryptographic
 verification of the data.
         </li>
         <li>
+          The specification MUST define a <dfn>suite selection algorithm</dfn> that accepts a [=JSON
+          object=] |proof| and uses the properties described in [[[#proofs]]] to return either
+          failure or a [=VC cryptographic suite=] [[Infra]] [=struct=]. This algorithm SHOULD be
+          registered in [[[VC-SPECS]]]. A
+          <dfn class="export">VC cryptographic suite</dfn> [=struct=] has the following
+          [=struct/items=].
+          <dl data-dfn-for="cryptographic suite">
+            <dt><dfn>transform</dfn></dt>
+            <dd>A [=transformation algorithm=] that takes a [=JSON object=] as input and returns a
+            [=JSON object=].</dd>
+            <dt><dfn>hash</dfn></dt>
+            <dd>A [=hashing algorithm=] that takes a [=JSON object=] as input and returns a [=byte
+            sequence=].</dd>
+            <dt><dfn>proof creation options</dfn></dt>
+            <dd>A [=struct=] type that needs to be supplied to the [=cryptographic suite/serialize
+            proof=] operation.</dd>
+            <dt><dfn>serialize proof</dfn></dt>
+            <dd>A [=proof serialization algorithm=] that takes a [=byte sequence=] and an instance
+            of this suite's [=proof creation options=] type as input and returns a [=JSON object=]
+            or failure.</dd>
+            <dt><dfn>verify proof</dfn></dt>
+            <dd>A [=proof verification algorithm=] that takes an [=unsecured data document=] and a
+            [=JSON object=] with the properties described in [[[#proofs]]], and returns "valid" or
+            "invalid".</dd>
+          </dl>
+        </li>
+        <li>
 The specification MUST detail any known resource starvation attack that can
 occur in an algorithm and provide testable mitigations against each attack.
         </li>
@@ -2543,11 +2570,11 @@ the Verifiable Credentials Specifications Directory</a>.
       <h2>Algorithms</h2>
 
       <p>
-The algorithms defined below are generalized in that they require a specific
-<a>transformation algorithm</a>, <a>hashing algorithm</a>,
-<a>proof serialization algorithm</a>, and <a>proof verification algorithm</a> to be
-specified by a particular cryptographic suite (see Section
-<a href="#cryptographic-suites"></a>).
+The algorithms defined below operate on documents represented as <dfn
+data-cite="RFC8259#section-4">JSON objects</dfn>. This specification follows [[JSON-LD-API]] in
+representing a JSON object as an [[Infra]] [=map=]. An
+<dfn>unsecured data document</dfn> is a [=JSON object=] with no "`proof`" [=map/key=]. A
+<dfn>secured data document</dfn> is a [=JSON object=] with a "`proof`" [=map/key=].
       </p>
 
       <p class="issue" title="Verification hash algorithm definition">
@@ -2865,20 +2892,11 @@ function baseDecode(sourceEncoding, sourceBase, baseAlphabet) {
         <p>
 The following algorithm specifies how to add a digital proof to a document,
 which  can be used to verify the authenticity and integrity of an
-<dfn>unsecured data document</dfn>. Required inputs are an
-<a>unsecured data document</a> (<var>unsecuredDocument</var>) and
-<a>proof options</a> (<var>options</var>). The
-<a>proof options</a> MUST contain a type identifier for the
-<a>cryptographic suite</a> (<var>type</var>) and any other properties needed by
-the <a>cryptographic suite</a> type; an identifier for the
-<a>verification method</a> (<var>verificationMethod</var>) that can be used to
-verify the authenticity of the proof; an [[!XMLSCHEMA11-2]] `dateTimeStamp`
-string (<var>created</var>) containing the current date and time,
-accurate to at least one second, in Universal Time Code format. A
-<a href="#dfn-domain">security domain</a> (<var>domain</var>) and/or a
-receiver-supplied challenge (<var>challenge</var>) MAY also be specified in
-the <var>options</var>. A <dfn>secured data document</dfn> is produced as
-output. Whenever this algorithm encodes strings, it MUST use UTF-8 encoding.
+<a>unsecured data document</a>. Required inputs are an
+<a>unsecured data document</a> (<var>unsecuredDocument</var>), a <a>VC cryptographic suite</a>
+(|suite:VC cryptographic suite|), and an instance of |suite|'s [=cryptographic suite/proof creation
+options=] (|options|). A <a>secured data document</a> is produced as output. Whenever this algorithm
+encodes strings, it MUST use UTF-8 encoding.
         </p>
 
         <ol class="algorithm">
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -2440,6 +2440,33 @@ <h2>Cryptographic Suites</h2>
 verification of the data.
         </li>
         <li>
+          The specification MUST define a <dfn>suite selection algorithm</dfn> that accepts a [=JSON
+          object=] |proof| and uses the properties described in [[[#proofs]]] to return either
+          failure or a [=VC cryptographic suite=] [[Infra]] [=struct=]. This algorithm SHOULD be
```
jyasskin, line 2447: This might be able to take just the `type` and `cryptosuite`, but I wasn't sure if y'all wanted to support some suites that also need other properties to figure out the algorithms.

index.html
```diff
@@ -2863,45 +2890,28 @@ <h3>Add Proof</h3>
...
-<dfn>unsecured data document</dfn>. Required inputs are an
-<a>unsecured data document</a> (<var>unsecuredDocument</var>) and
-<a>proof options</a> (<var>options</var>). The
-<a>proof options</a> MUST contain a type identifier for the
-<a>cryptographic suite</a> (<var>type</var>) and any other properties needed by
-the <a>cryptographic suite</a> type; an identifier for the
-<a>verification method</a> (<var>verificationMethod</var>) that can be used to
-verify the authenticity of the proof; an [[!XMLSCHEMA11-2]] `dateTimeStamp`
-string (<var>created</var>) containing the current date and time,
-accurate to at least one second, in Universal Time Code format. A
```
jyasskin, line 2877: I dropped the timestamp because the signing algorithm can retrieve the current date and time just as well as the caller.

index.html
```diff
@@ -2863,45 +2890,28 @@ <h3>Add Proof</h3>
...
 The following algorithm specifies how to add a digital proof to a document,
 which  can be used to verify the authenticity and integrity of an
-<dfn>unsecured data document</dfn>. Required inputs are an
-<a>unsecured data document</a> (<var>unsecuredDocument</var>) and
-<a>proof options</a> (<var>options</var>). The
-<a>proof options</a> MUST contain a type identifier for the
-<a>cryptographic suite</a> (<var>type</var>) and any other properties needed by
-the <a>cryptographic suite</a> type; an identifier for the
-<a>verification method</a> (<var>verificationMethod</var>) that can be used to
-verify the authenticity of the proof; an [[!XMLSCHEMA11-2]] `dateTimeStamp`
```
jyasskin, line 2875: The verification method is being used to smuggle in a private key, but that's far from obvious in the old interface. The new interface also doesn't say it explicitly, but the private key should go into the _proof creation options_ that the suite specifies.

index.html
```diff
@@ -2440,6 +2440,33 @@ <h2>Cryptographic Suites</h2>
...
+          registered in [[[VC-SPECS]]]. A
+          <dfn class="export">VC cryptographic suite</dfn> [=struct=] has the following
+          [=struct/items=].
+          <dl data-dfn-for="cryptographic suite">
+            <dt><dfn>transform</dfn></dt>
+            <dd>A [=transformation algorithm=] that takes a [=JSON object=] as input and returns a
+            [=JSON object=].</dd>
+            <dt><dfn>hash</dfn></dt>
+            <dd>A [=hashing algorithm=] that takes a [=JSON object=] as input and returns a [=byte
+            sequence=].</dd>
```
jyasskin, line 2457: There's a note that the hashing algorithm can actually return things that aren't just single byte sequences. The suite could specify a type for this, or we could just say it returns an instance of a type that can be passed to the **serialize proof** algorithm, or we could merge hashing with serializing.

index.html
```diff
@@ -3000,21 +3013,31 @@ <h3>Verify Proof</h3>
...
-<var>challenge</var>. The <var>expectedProofPurpose</var> value is used to
-ensure that the <var>proof</var> was generated by the proof creator for the
-expected reason by the verifier, such as "authentication". The <var>domain</var>
-value is used by the proof creator to lock a proof to a particular security
-domain, and used by the verifier to ensure that a proof is not being used across
-different security domains. The <var>challenge</var> value is used by the
-verifier to ensure that an attacker is not replaying previously created proofs.
-Additional options are expected to be provided by libraries to ensure that
-digital proof time stamps do not deviate more than an acceptable time frame
-for a given use case. A <dfn>verification result</dfn> is produced as output.
```
jyasskin, line 3035: I dropped this for now, but from some of the discussion elsewhere, it sounded like the WG wants these time-checking operations to be part of "validation" rather than "verification". If that's the case, this algorithm should *return* some extra data for the validator to act on.

index.html
```diff
@@ -3000,21 +3013,31 @@ <h3>Verify Proof</h3>
...
+<var>proof</var> was generated by the proof creator for the expected reason by the verifier, such as
+"authentication"
+          </li>
+          <li>
+|domain|, an optional [=string=], used by the proof creator to lock a proof to a particular security
+domain, and used by the verifier to ensure that a proof is not being used across different security
+domains
+          </li>
+          <li>
+|challenge|, an optional [=string=] [=challenge=], used by the verifier to ensure that an attacker
```
jyasskin, line 3059: Are these actually strings?

index.html
```diff
@@ -3046,41 +3074,30 @@ <h3>Verify Proof</h3>
...
-Let <var>isProofVerified</var> be the result of running the
-<a>proof verification algorithm</a> associated with the
-<a>cryptographic suite</a> with the <var>hashData</var> and <var>options</var>
-parameters provided as inputs to the algorithm.
+Let |isProofVerified:valid or invalid| be the result of running |suite|'s
+[=cryptographic suite/verify proof=] algorithm on <var>hashData</var> and <var>proof</var>.
           </li>
           <li>
-If <var>options</var>.<var>domain</var> is set and it does not match
+If <var>domain</var> was given, and it does not match
```
jyasskin, line 3114: "match" is a yellow flag in my spec reviews. Usually it should be "equal", but sometimes people mean a different comparison algorithm that needs to be specified. That fix should go in a different PR.

index.html
dlongley: An important thing we want to avoid is saying that any algorithms require private key material to be passed in opposed to an interface for performing signing. It's often the case that signals are sent to implementers that they ought to require people to pass in private key material when requiring a signer interface instead enables much better security practices (such as HSM use, etc.). These signals also sometimes create headaches where implementers choose to provide synchronous interfaces (as opposed to asynchronous ones) which do not even allow people to "squeeze in" HSM-backed signing mechanisms.

jyasskin: Good point. I've tried to incorporate this in 3d6aa172a485ade9d5a5167d2a7b7ff4221064ad.

msporny (approved):
Approving so we can get these changes into the specification and rework the language. 

There are now merge conflicts with @Wind4Greg's previous PR that got merged. I tried for an hour or so to resolve the differences and was unsure of the result.

@jyasskin, can you please rebase and fix the conflicts so we can do a clean merge?

jyasskin:
@msporny Thanks for trying. I've rebased, and I think it was just a textual conflict and not a problem with our meanings. It should be good to go.

index.html
msporny: The intention here was "equal" ... as in string equality. Can fix in a future iteration as an editorial fix.

msporny:
@jyasskin wrote:
> I've rebased, and I think it was just a textual conflict and not a problem with our meanings. It should be good to go.

Great, thank you! ... and thank you for this PR, it's a solid improvement to the interfaces between VCDM, the VC Specs Directory, and the various Data Integrity specs... I'll try to follow it up w/ interface PRs to each of the Data Integrity specs over the coming weeks.

My focus over the next month will be getting VCDM v2.0 and vc-bitstring-status-list into CR... I'll then circle back around to the Data Integrity specs.

index.html
```diff
@@ -3064,41 +3099,30 @@ <h3>Verify Proof</h3>
...
           <li>
-Let <var>transformedData</var> be the result of
-<a href="#dfn-transformation">transforming</a> the <var>unsecuredDocument</var>
-according to a <a>transformation algorithm</a> associated with the
-<a>cryptographic suite</a> specified in <var>proof</var> and the
-<var>options</var> parameters provided as inputs to the algorithm.
-The type of <a>cryptographic suite</a> is specified by the
-<var>proof.type</var> value and MAY be further described by
-<a>cryptographic suite</a>-specific properties expressed in <var>proof</var>.
+Let <var>transformedData</var> be the result of running |suite|'s'
```
TallTed, line 3102:
```suggestion
Let <var>transformedData</var> be the result of running |suite|'s
```

index.html
```diff
@@ -3064,41 +3099,30 @@ <h3>Verify Proof</h3>
...
+[=cryptographic suite/transform=] algorithm on the <var>unsecuredDocument</var>.
         </li>
           <li>
-Let <var>hashData</var> be the result of
-<a href="#dfn-hashing">hashing</a> the <var>transformedData</var>
-according to a <a>hashing algorithm</a> associated with the
-<a>cryptographic suite</a> specified in the <var>proof</var> and
-<var>options</var> parameters provided as inputs to the algorithm.
+Let <var>hashData</var> be the result of running |suite|'s'
```
TallTed, line 3106:
```suggestion
Let <var>hashData</var> be the result of running |suite|'s
```

index.html
```diff
@@ -3064,41 +3099,30 @@ <h3>Verify Proof</h3>
...
+[=cryptographic suite/hash=] algorithm on the <var>transformedData</var>.
           </li>
           <li>
-Let <var>isProofVerified</var> be the result of running the
-<a>proof verification algorithm</a> associated with the
-<a>cryptographic suite</a> with the <var>hashData</var> and <var>options</var>
-parameters provided as inputs to the algorithm.
+Let |isProofVerified:valid or invalid| be the result of running |suite|'s
```
TallTed, line 3110:
```suggestion
Let |isProofVerified:valid| or |isProofVerified:invalid| be the result of running |suite|'s
```

index.html
msporny: We wanted the ability to specify the time for which a proof was being created, or verified. There are use cases where you want to verify against a time in the past, or sign for a time in the future. I believe this line was there to ensure that one could have that flexibility. We can add this back in a future PR.

TallTed (changes requested):

index.html
msporny: Yes, we wanted to give implementations the latitude to accept other parameters that they might need. The algorithms here are supposed to define the bare minimum necessary, but not restrict implementations for doing more than the bare minimum.

index.html
msporny: Yes, they are.

msporny:

index.html
msporny: Ah, crap... I merged w/o seeing these... raise another PR to fix these, please, @TallTed.... note that I'm getting ready to merge the other PR from @jyasskin in a bit, so maybe wait a day for this stuff to settle before doing your PR.

jyasskin: @TallTed's suggestion isn't the right fix here, although I agree that what I wound up with is awkward. This is the `|variable:type|` syntax from https://respec.org/docs/#giving-variables-data-types, and the type of the variable is actually the enum type of "valid" or "invalid". Respec doesn't accept `"` inside its |var| syntax, and it doesn't appear to have another way of attaching types, so I dropped the quotes. One could argue that I should have instead given up on assigning a type to this variable.

TallTed: @jyasskin — I think you're in a better position than I to log a ReSpec issue, about how to type such enum-typed vars ... whether that leads to some complex manual process, an improvement of ReSpec, or otherwise.

jyasskin: Indeed, I am. And in figuring out what issue to file, I found the markup I should have used: https://github.com/w3c/respec-web-services/pull/399 and https://github.com/w3c/respec/wiki/data-type.

TallTed: @jyasskin — As you've noted that _Respec doesn't accept `"` inside its `|var|` syntax_, it seems that some addition detail should be in [that wiki](https://github.com/w3c/respec/wiki/data-type) (and hence the proper ReSpec documentation) to cover such scenarios as this enum typing.

jyasskin: Sure, done: https://github.com/w3c/respec/wiki/data-type. Along with the bug to make it support quotes.

PR #244, opened 2024-02-12
Fix algorithm misalignments using new cryptosuite interface.

msporny:
This PR updates the "Cryptographic Suite Selection Algorithm" [interface that was defined](https://w3c.github.io/vc-data-integrity/#dfn-suite-selection-algorithm) by @jyasskin that MUST be defined by all DI cryptographic suite specifications. PR https://github.com/w3c/vc-di-ecdsa/pull/57 implements the interface on the ECDSA cryptosuite and this PR attempts to align that PR with the DI specification. This PR also fixes some misalignments between the base algorithms and the usage of the suite.

This PR doesn't attempt to align the selective disclosure algorithms, that might happen in a separate PR.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/244.html" title="Last updated on Feb 24, 2024, 6:46 PM UTC (1f7ee96)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/244/26857e9...1f7ee96.html" title="Last updated on Feb 24, 2024, 6:46 PM UTC (1f7ee96)">Diff</a>


index.html
```diff
@@ -2435,37 +2435,58 @@ <h2>Cryptographic Suites</h2>
...
-          The specification MUST define a <dfn>suite selection algorithm</dfn> that accepts a [=JSON
-          object=] |proof| and uses the properties described in [[[#proofs]]] to return either
-          failure or a [=VC cryptographic suite=] [[Infra]] [=struct=]. This algorithm SHOULD be
-          registered in [[[VC-SPECS]]]. A
-          <dfn class="export">VC cryptographic suite</dfn> [=struct=] has the following
-          [=struct/items=].
+The specification MUST define a <dfn>suite selection algorithm</dfn> that
+accepts a set of options ([=map=] |options|) and returns a
+[=cryptographic suite=] ([=struct=] |cryptosuite|). This algorithm SHOULD be
+registered in [[[VC-SPECS]]]. A cryptographic suite [=struct=] has the
```
iherman: Just knit-picking. We may want to use a different term for "registered" to avoid misunderstandings. The VC-SPECS document explicitly says that this is not a registry, and using the term "registered" may lead the reader to believe that it is. Just to be careful...

```suggestion
[=cryptographic suite=] ([=struct=] |cryptosuite|). This algorithm SHOULD be
listed in [[[VC-SPECS]]]. A cryptographic suite [=struct=] has the
```

index.html
```diff
@@ -2435,37 +2435,58 @@ <h2>Cryptographic Suites</h2>
...
+([=struct=] |canonicalProofConfig|) as input and returns a [=byte sequence=],
+or an error, as output.
+            </dd>
+            <dt><dfn>createProofConfig</dfn></dt>
+            <dd>
+              <p>
+An algorithm that takes proof options ([=struct=] |options|) as input and
+produces a proof configuration ([=struct=] proofConfig) as output, which is
+then supplied to the [=cryptographic suite/serializeProof=] algorithm. If a
+failure occurs, and error is produced.
```
```suggestion
failure occurs, an error is produced.
```

index.html
```diff
@@ -2435,37 +2435,58 @@ <h2>Cryptographic Suites</h2>
...
+input and produces a signal digital proof value ([=JSON object=] proof), or
+an error, as output.
+            </dd>
+            <dt><dfn>verifyProof</dfn></dt>
+            <dd>
+A [=proof verification algorithm=] that takes an  are cryptographic hash data
+([=byte sequence=] |hashData|), a digital proof
+([=byte sequence=] |proofBytes|) and proof options ([=struct=] options) as
+input and produces a verification result represented as a boolean value, or
+an error, as output.
```
```suggestion
([=byte sequence=] |proofBytes|), and proof options ([=struct=] options) as
input, and produces as output either a verification result represented as a
boolean value or an error.
```

index.html
```diff
@@ -2435,37 +2435,58 @@ <h2>Cryptographic Suites</h2>
...
+identifier or an asynchronous signing algorithm allows implementations to take
+advantage of hardware security modules (HSMs) to get better security.
+              </p>
+            </dd>
+            <dt><dfn>serializeProof</dfn></dt>
+            <dd>
+An [=proof serialization algorithm=] that takes cryptographic hash data
+([=byte sequence=] |hashData|) and proof options ([=struct=] options) as
+input and produces a signal digital proof value ([=JSON object=] proof), or
+an error, as output.
```
```suggestion
input, and produces as output either a single digital proof value
([=JSON object=] proof) or an error.
```

index.html
```diff
@@ -2891,41 +2912,56 @@ <h3>Add Proof</h3>
...
 Let |output| be a copy of |unsecuredDocument|.
           </li>
           <li>
-Let |transformedData| be the result of running |suite|'s [=cryptographic suite/transform=]
-algorithm on |unsecuredDocument|.
+Set |suite:cryptographic suite| to the first non-empty result from running the
+[=suite selection algorithm=] associated with each item in |cryptosuites|,
+passing |options| as a parameter. If no suite is found, an error
+MUST be raised and SHOULD convey an error type of
+<a href="#PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</a>.
```
```suggestion
<a href="#PROOF_GENERATION_ERROR">`PROOF_GENERATION_ERROR`</a>.
```

index.html
```diff
@@ -2891,41 +2912,56 @@ <h3>Add Proof</h3>
...
-or |proof|.|proofPurpose| values are not set, an error
-MUST be raised and SHOULD convey an error type of
+Let |proof| be the result of running |suite|'s
+[=cryptographic suite/serializeProof=] algorithm with |hashData| and |options|
+passed as parameters.
+          </li>
+          <li>
+If the |proof|.|type|, |proof|.|verificationMethod|, or |proof|.|proofPurpose|
+values are not set, an error MUST be raised and SHOULD convey an error type of
 <a href="#PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</a>.
```
TallTed, line 2953:
```suggestion
<a href="#PROOF_GENERATION_ERROR">`PROOF_GENERATION_ERROR`</a>.
```

TallTed (changes requested):

Wind4Greg (approved):
Looks good with only one word usage issue: "signal" should be "single"? Right?

index.html
```diff
@@ -2435,37 +2435,58 @@ <h2>Cryptographic Suites</h2>
...
-              proof=] operation.</p>
-              <p class="note">To sign data, cryptographic suites need to be able to use a private
-              key. This private key will often be implied by a field in the [=proof creation
-              options=], but it is often a better design to avoid passing the key directly. Passing
-              a key identifier or an asynchronous signing algorithm allows implementations to take
-              advantage of hardware security modules (HSMs) to get better security.</p>
+A [=hashing algorithm=] that takes a transformed data document
+([=JSON object=] |transformedDocument|) and canonical proof configuration
+([=struct=] |canonicalProofConfig|) as input and returns a [=byte sequence=],
+or an error, as output.
```
dlongley: I need to look more closely at this, but it seems like it might be slightly misaligned or overly prescriptive for use with selective disclosure cryptosuites that return other cryptosuite-specific outputs (e.g., https://w3c.github.io/vc-di-ecdsa/#base-proof-hashing-ecdsa-sd-2023 returns an object with multiple values, some hashes and others not, as key-value pairs).

I'm concerned we're trying too hard to pin down too much of the interface here rather than providing core functions and a base, loose, architecture for extending cryptosuites to fit into. There is generally a transformation step, which always takes the unsecured data document as an input -- but the output can be specialized to the cryptosuite -- and, generally, a hashing step follows that receives this cryptosuite-specific output and generates its own cryptosuite-specific output, which is then followed by a proof generation/serialization step that finally outputs the data integrity proof object.

In short, there's a pipeline that is cryptosuite specific that always starts with an unsecured document and ends up with a data integrity proof at the end of it, but cryptosuites should be free to define their own intermediates. The fact that the data integrity core spec provides some useful primitives to use there does not mean that every cryptosuite needs to use them exactly as they are for "simple digital signatures". What is required, however, is unsecured document in, and data integrity proof out (for the entire pipeline).

I also think it's useful to keep the "proof configuration options" typing information we've added here, but we might need to relax some of the "pipeline intermediates" to prevent unnecessarily restricting innovation in cryptosuites in that space. It's also just not even aligned with current selective disclosure suites (slightly as mentioned).

dlongley (changes requested):

index.html
jyasskin: +1 to the idea of merging the "transform", "hash", and "serializeProof" algorithms for the purposes of this interface. https://github.com/w3c/vc-data-integrity/pull/226#discussion_r1419986552 listed a couple options for improving the flexibility here, but since the only use of this interface just calls the operations as a pipeline, merging them seems best.

index.html
```diff
@@ -2435,37 +2435,58 @@ <h2>Cryptographic Suites</h2>
 verification of the data.
         </li>
         <li>
-          The specification MUST define a <dfn>suite selection algorithm</dfn> that accepts a [=JSON
-          object=] |proof| and uses the properties described in [[[#proofs]]] to return either
-          failure or a [=VC cryptographic suite=] [[Infra]] [=struct=]. This algorithm SHOULD be
-          registered in [[[VC-SPECS]]]. A
-          <dfn class="export">VC cryptographic suite</dfn> [=struct=] has the following
```
jyasskin, line 2466: You really should keep the exported definition, so that other specs can refer to the whole struct at once.

index.html
```diff
@@ -2435,37 +2435,58 @@ <h2>Cryptographic Suites</h2>
...
+registered in [[[VC-SPECS]]]. A cryptographic suite [=struct=] has the
+following [=struct/items=].
           <dl data-dfn-for="cryptographic suite">
             <dt><dfn>transform</dfn></dt>
-            <dd>A [=transformation algorithm=] that takes a [=JSON object=] as input and returns a
-            [=JSON object=].</dd>
+            <dd>
+A [=transformation algorithm=] that takes an [=unsecured data document=]
+([=JSON object=] |unsecuredDocument|) and transformation options
+([=struct=] |options|) as input and produces a
```
jyasskin: "struct" shouldn't really be used as the type of an input parameter, any more than you use "struct" as the type of a C variable. Instead, you create concrete types as structs of a particular set of "items", and then input parameters are typed by those concrete types. If you want an unbounded set of keys, use a [map](https://infra.spec.whatwg.org/#maps).

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-02-14-vcwg#section1-1) on 2024-02-14

- no resolutions were taken

#### 1.1. Fix algorithm misalignments using new cryptosuite interface. (pr vc-data-integrity#244)

_See github pull request [vc-data-integrity#244](https://github.com/w3c/vc-data-integrity/pull/244)._




**Manu Sporny:** processed a number of PRs across VCDM, DI, cryptosuites. need to talk about Jeffery Yaskin's PR (#244) to create an interface for all DI specs.  

_See github pull request [vc-di-ecdsa#57](https://github.com/w3c/vc-di-ecdsa/pull/57)._




**Manu Sporny:** that broke interfaces b/w DI specs. trying to get them re-aligned. 2 PRs - 1 for DI, 1 for ECDSA-SD. heads up to the group we're trying to align these interfaces.  
… some misalignment on how they would work. have a plan forward to address this. plan is for an interface in all DI specs that all have 'functions' each cryptosuite executes to create/verify proofs. a standard interface.  
… the functions to expose was under debate. based on discussion we will only define 2 functions on the interface: create proof and verify proof.  
… will require changes to algorithms across these specs. pushing more details into the cryptosuite specs. less in DI the spec. should not impact implementations. we know we will go through a 2nd CR. the interfaces are changing, not the algorithms.  
… are there any concerns/guidance before I start making those changes?  

> *Dave Longley:* +1 to those changes.

**Ivan Herman:** presume that ECDSA then EDDSA and then BBS?  

**Manu Sporny:** correct.  

**Michael Jones:** think it takes us down a bad path to build interfaces that no one will build. we should not be creating APIs, that is out of scope.  

**Manu Sporny:** agree that APIs are out of scope, but that's not what we're creating here.  

**Michael Jones:** that is what you described.  

**Manu Sporny:** have discussed this before. we're discussing interfaces, which is what the w3c does, not in web IDL which would define an API. implementations are implementing in this way. they are abstract, not concrete web IDL interfaces.  

**Michael Jones:** I am missing context. what else are you planning to do?  

**Manu Sporny:** changing the interfaces that we had months ago, which Jefferey asked for. that PR had weeks of review and already went in.  

**Brent Zundel:** any other comments?  

**Manu Sporny:** no - that's the major PR I need feedback on.  

</details>

index.html
msporny: Removing the export was a mistake, I'll fix it in the next revision. 

The one concern was calling it a "VC cryptographic suite" -- the only reason we're doing the Data Integrity work in the VCWG is due to politics. The Data Integrity technology is NOT supposed to be bound at all to the Verifiable Credential work (except that VCs are just ONE use of Data Integrity)... my concern in doing this (many years ago when the work was chartered) was that people would think that DI is tightly coupled to VCs, which is what certain actors wanted during the chartering process, but was never the intention of the work. All that to say "VC cryptographic suite" is wrong, "DI cryptographic suite" might be an ok compromise, but exporting it as "cryptographic suite" doesn't feel wrong either (since the xref would have to refer to the spec to import the language).

Just noting this here to see if you have any thoughts. I'll export the definition and adjust further if you have strong feelings one way or the other.

index.html
msporny: Yes, agreed, I accepted @TallTed's change where he changed "registered" -> "listed".

index.html
```suggestion
([=map=] |options|) as input and produces a
```

msporny: Fixed to use a map.

index.html
msporny: Ok, we [had a conversation about this during the call yesterday](https://github.com/w3c/vc-data-integrity/pull/244#issuecomment-1944257979) and are taking @jyasskin and @dlongley's guidance and fixing up the interface to be simpler / higher level.

index.html
jyasskin: Defining "cryptographic suite" is fine with me. I only picked the more specific term because I wasn't sure it's what the WG intended.

iherman: I would go for "DI  Cryptographic Suite".

jyasskin: FWIW, Bikeshed has xrefs that pull in everything, without needing the spec to be listed, so we'd want to resolve a collision on "cryptographic suite" if one happens, but resolving those isn't the end of the world. If the WG thinks the interface here will work for cryptographic suites outside of this ecosystem (like in https://w3c.github.io/webcrypto/), it'd be fine to claim the term, IMO. If the WG doesn't want to be that general, then picking the more specific term is better.

msporny: > FWIW, Bikeshed has xrefs that pull in everything, without needing the spec to be listed

Ah, well, yes, that would be an issue. I'll go with the longer name for now. We could probably utilize the same concept for webcrypto, but that is far more work than this WG has signed up to do.

msporny: I settled on `cryptographic suite instance` for now, and exported the definition. I tried `data integrity cryptographic suite instance` and repeatedly using that in the refs in the spec was a pain.

index.html
msporny: Ok, reworked the entire PR, but only focused on the interface definition and the usage of the interface in Add Proof and Verify Proof. It was more painful than I was hoping it would be... but it might be closer to what folks want now. I still need to work on the other side of the implementation in the matching ECDSA PR.

We would still need to do alignments w/ the Proof Set/Chain stuff in a later PR.

msporny:
Alright @jyasskin, @dlongley, @iherman, @TallTed, I've made another pass at this and attempted to apply each of your concerns... please re-review and let me know if further adjustments are needed.

index.html
iherman: I understand it is a pain but, thinking a bit further, I do believe it is worth having a reference to DI in the name. If I look at the spec from a _purely Linked Data perspective_ using this spec for _any_ RDF Dataset, then there are some problems that these specs do not really solve (we would get back to the discussions we had with Henry Story, which could be settled properly for VC but not in general).

Solving those issues is not in the charter of this WG, so that is o.k., but that is why I believe a reference to DI would make sense.

index.html
```diff
@@ -2435,37 +2435,28 @@ <h2>Cryptographic Suites</h2>
...
-              <p>A [=struct=] type that needs to be supplied to the [=cryptographic suite/serialize
-              proof=] operation.</p>
-              <p class="note">To sign data, cryptographic suites need to be able to use a private
-              key. This private key will often be implied by a field in the [=proof creation
-              options=], but it is often a better design to avoid passing the key directly. Passing
-              a key identifier or an asynchronous signing algorithm allows implementations to take
-              advantage of hardware security modules (HSMs) to get better security.</p>
+An algorithm that takes an [=unsecured data document=] ([=JSON object=]
+|unsecuredDocument|) and proof options ([=map=] |options|) as input and produces
+a [=secured data document=] ([=JSON object=]), or an error, as output.
```
```suggestion
|unsecuredDocument|) and proof options ([=map=] |options|) as input, and outputs
a [=secured data document=] ([=JSON object=]) or an error.
```

index.html
```diff
@@ -2435,37 +2435,28 @@ <h2>Cryptographic Suites</h2>
...
+            </dd>
+            <dt><dfn>verifyProof</dfn></dt>
+            <dd>
+An algorithm that takes a [=secured data document=] ([=JSON object=]
+|securedDocument|) as input and produces a verification result
+([=map=]), or an error, as output. The result MUST contain a
```
```suggestion
|securedDocument|) as input, and outputs a verification result
([=map=]) or an error. The verification result MUST contain a
```

index.html
```diff
@@ -2486,6 +2477,13 @@ <h2>Cryptographic Suites</h2>
         </li>
       </ul>
 
+      <p>
+A [=cryptographic suite=] is instantiated using a [=cryptographic suite
+instantiation algorithm=] and is made available to algorithms in an
+implementation-specific manner, which MAY utilize [[[VC-SPECS]]] to discover
+known [=cryptographic suite instantiation algorithms=].
```
```suggestion
implementation-specific manner. Implementations MAY use [[[VC-SPECS]]]
[[VC-SPECS]] to discover known [=cryptographic suite instantiation algorithms=].
```

index.html
```diff
@@ -2890,42 +2888,32 @@ <h2>Base Decode</h2>
...
-(|suite:VC cryptographic suite|), and an instance of |suite|'s [=cryptographic suite/proof creation
-options=] (|options|). A <a>secured data document</a> is produced as output. Whenever this algorithm
-encodes strings, it MUST use UTF-8 encoding.
+The following algorithm specifies how to add a digital proof to an
+<a>unsecured data document</a>, which can be used to verify its authenticity and
+integrity. Required inputs are an <a>unsecured data document</a>
+([=JSON object=] |unsecuredDocument|), a [=cryptographic suite instance=]
+([=struct=] |cryptosuite|), and a set of options ([=map=] |options|). A
+<a>secured data document</a>, or an error, is produced as output. Whenever this
+algorithm encodes strings, it MUST use UTF-8 encoding.
```
```suggestion
The following algorithm specifies how a digital proof can be added to an
<a>unsecured data document</a>, and can then be used to verify the output document's
authenticity and integrity. Required inputs are an <a>unsecured data document</a>
([=JSON object=] |unsecuredDocument|), a [=cryptographic suite instance=]
([=struct=] |cryptosuite|), and a set of options ([=map=] |options|). Output is
a <a>secured data document</a> or an error. Whenever this
algorithm encodes strings, it MUST use UTF-8 encoding.
```

index.html
```diff
@@ -2890,42 +2888,32 @@ <h2>Base Decode</h2>
...
+parameter.
           </li>
           <li>
-If the |proof|.|type|,
-|proof|.|verificationMethod|,
-or |proof|.|proofPurpose| values are not set, an error
-MUST be raised and SHOULD convey an error type of
+If the |proof|.|type|, |proof|.|verificationMethod|, or |proof|.|proofPurpose|
+values are not set, an error MUST be raised and SHOULD convey an error type of
 <a href="#PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</a>.
```
```suggestion
If one or more of the |proof|.|type|, |proof|.|verificationMethod|, and
|proof|.|proofPurpose| values is not set, an error MUST be raised and SHOULD convey
an error type of <a href="#PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</a>.
```

index.html
```diff
@@ -2946,27 +2934,17 @@ <h3>Add Proof</h3>
...
 The following algorithm specifies how to incrementally add a proof to a proof
 set or proof chain starting with a secured document containing either a proof or
 proof set/chain. Required inputs are a <a>secured data document</a>
-(|securedDocument|), a <a>VC cryptographic suite</a>
-(|suite:VC cryptographic suite|), and an instance of |suite|'s [=cryptographic suite/proof creation
-options=] (|options|). A new <a>secured data document</a> is produced as
-output. Whenever this algorithm encodes strings, it MUST use UTF-8 encoding.
+(|securedDocument|), a <a>cryptographic suite</a>
+(|suite:cryptographic suite|), and a set of ([=map=] |options|). A new
+<a>secured data document</a> is produced as output. Whenever this algorithm
```
```suggestion
(|suite:cryptographic suite|), and a set of options ([=map=] |options|). Output
is a new <a>secured data document</a>. Whenever this algorithm
```

index.html
```diff
@@ -3031,140 +3009,128 @@ <h3>Verify Proof</h3>
...
-          <li>
-|documentBytes|, a [=byte sequence=] whose media type is |mediaType|
-          </li>
-          <li>
-|expectedProofPurpose|, an optional [=string=], used to ensure that the
-|proof| was generated by the proof creator for the expected reason by the verifier. See
+        <dl>
+          <dt>|mediaType|</dt>
+          <dd>
+A [=MIME type=].
```
TallTed: It's a `MIME media type` or a `media type` or maybe an `IANA media type`; never a `MIME type`. I know `[=MIME type=]` means it's in multiple locations, so I'm leaving all those changes to you.

index.html
```diff
@@ -3031,140 +3009,128 @@ <h3>Verify Proof</h3>
...
+          </dd>
+          <dt>|documentBytes|</dt>
+          <dd>
+A [=byte sequence=] whose media type is |mediaType|.
```
```suggestion
A [=byte sequence=] whose media type is |mediaType|
```

index.html
```diff
@@ -3031,140 +3009,128 @@ <h3>Verify Proof</h3>
...
+          </dd>
+          <dt>|cryptosuite|</dt>
+          <dd>
+A [=cryptographic suite instance=].
```
```suggestion
A [=cryptographic suite instance=]
```

index.html
```diff
@@ -3031,140 +3009,128 @@ <h3>Verify Proof</h3>
...
+          </dd>
+          <dt>|expectedProofPurpose|</dt>
+          <dd>
+An optional [=string=], used to ensure that the |proof| was generated by the
+proof creator for the expected reason by the verifier. See
 [[[#proof-purposes]]] for common values.
```
```suggestion
[[[#proof-purposes]]] for common values
```

index.html
```diff
@@ -3031,140 +3009,128 @@ <h3>Verify Proof</h3>
...
-          <li>
-|challenge|, an optional [=string=] [=challenge=], used by the verifier to ensure that an attacker
-is not replaying previously created proofs
-          </li>
-        </ul>
+          <dt>|domain|</dt>
+          <dd>
+An optional [=set=] of [=strings=], used by the proof creator to lock a proof to
+a particular security domain, and used by the verifier to ensure that a proof is
+not being used across different security domains.
```
```suggestion
not being used across different security domains
```

index.html
```diff
@@ -3031,140 +3009,128 @@ <h3>Verify Proof</h3>
...
+          </dd>
+          <dt>|challenge|</dt>
+          <dd>
+An optional [=string=] [=challenge=], used by the verifier to ensure that an
+attacker is not replaying previously created proofs.
```
```suggestion
attacker is not replaying previously created proofs
```

index.html
```diff
@@ -3031,140 +3009,128 @@ <h3>Verify Proof</h3>
...
         <p>
-This algorithm returns a <dfn>verification result</dfn>, a [=struct=] whose [=struct/items=] are:
+This algorithm returns a <dfn>verification result</dfn>, a [=struct=] whose
+[=struct/items=] are:
         </p>
         <dl>
-          <dt><dfn data-dfn-for="verification result">status</dfn></dt>
-          <dd>"valid" or "invalid"</dd>
+          <dt><dfn data-dfn-for="verification result">verified</dfn></dt>
+          <dd>`true` or `false`.</dd>
```
```suggestion
          <dd>`true` or `false`</dd>
```

index.html
```diff
@@ -3031,140 +3009,128 @@ <h3>Verify Proof</h3>
...
           <dt><dfn data-dfn-for="verification result">document</dfn></dt>
-          <dd>null if [=verification result/status=] is "invalid" or an [=unsecured data
-          document=]</dd>
+          <dd>
+null if [=verification result/verified=] is `false` or an [=unsecured data
+document=].
```
```suggestion
NULL, if [=verification result/verified=] is `false`; otherwise, an [=unsecured data
document=]
```

index.html
```diff
@@ -3031,140 +3009,128 @@ <h3>Verify Proof</h3>
...
+          </dd>
           <dt><dfn data-dfn-for="verification result">media type</dfn></dt>
-          <dd>null if [=verification result/status=] is "invalid" or a [=MIME type=], which may
-          include [=MIME type/parameters=]</dd>
+          <dd>
+null if [=verification result/verified=] is `false` or a [=MIME type=], which
+MAY include [=MIME type/parameters=].
```
```suggestion
NULL, if [=verification result/verified=] is `false`; otherwise, a [=MIME type=],
which MAY include [=MIME type/parameters=]
```

index.html
```diff
@@ -3031,140 +3009,128 @@ <h3>Verify Proof</h3>
...
+          </dd>
           <dt><dfn data-dfn-for="verification result" class="lint-ignore">warnings</dfn></dt>
-          <dd>a [=list=] of <dfn data-cite="VC-DATA-MODEL-2.0#dfn-problemdetails">ProblemDetails</dfn>,
-          defaulting to an empty [=list=]</dd>
+          <dd>
+a [=list=] of
+<dfn data-cite="VC-DATA-MODEL-2.0#dfn-problemdetails">ProblemDetails</dfn>,
+defaulting to an empty [=list=].</dd>
```
```suggestion
which defaults to an empty [=list=]</dd>
```

index.html
```diff
@@ -3031,140 +3009,128 @@ <h3>Verify Proof</h3>
...
           <dt><dfn data-dfn-for="verification result">errors</dfn></dt>
-          <dd>a [=list=] of [=ProblemDetails=], defaulting to an empty [=list=]</dd>
+          <dd>
+a [=list=] of [=ProblemDetails=], defaulting to an empty [=list=].</dd>
```
```suggestion
a [=list=] of [=ProblemDetails=], which defaults to an empty [=list=]</dd>
```

index.html
```diff
@@ -3031,140 +3009,128 @@ <h3>Verify Proof</h3>
...
         </dl>
 
         <p>
-When a step says "an error MUST be raised", it means to return a [=verification result=] with a
-[=verification result/status=] of "invalid" and a non-empty [=verification result/errors=] list.
+When a step says "an error MUST be raised", it means to return a [=verification
+result=] with a [=verification result/verified=] of `false` and a non-empty
+[=verification result/errors=] list.
```
```suggestion
When a step says "an error MUST be raised", it means that a [=verification
result=] MUST be returned with a [=verification result/verified=] of `false`
and a non-empty [=verification result/errors=] list.
```

index.html
```diff
@@ -3031,140 +3009,128 @@ <h3>Verify Proof</h3>
...
           <li>
-If |proof|["`type`"],
-|proof|["`verificationMethod`"], or
-|proof|["`proofPurpose`"] do not [=map/exist=],
-an error MUST be raised and SHOULD convey an error type of
-<a href="#MALFORMED_PROOF_ERROR">MALFORMED_PROOF_ERROR</a>.
+If |securedDocument| is not a [=map=] or |securedDocument|.|proof| is not a
+[=map=], an error MUST be raised and SHOULD convey an error type of
+<a href="https://www.w3.org/TR/VC-DATA-MODEL-2.0#PARSING_ERROR">
+PARSING_ERROR</a>.
```
```suggestion
If either |securedDocument| is not a [=map=] or |securedDocument|.|proof|
is not a [=map=], an error MUST be raised and SHOULD convey an error type of
<a href="https://www.w3.org/TR/VC-DATA-MODEL-2.0#PARSING_ERROR">
PARSING_ERROR</a>.
```

index.html
```diff
@@ -3031,140 +3009,128 @@ <h3>Verify Proof</h3>
...
-If all of these algorithms return failure, return "invalid". Otherwise, let |suite:VC cryptographic
-suite| be an [=implementation-defined=] choice of one of the non-failure results.
+Let |proof:map| be |securedDocument|.|proof|.
           </li>
           <li>
-If |suite| requires the
-|proof|["`created`"] value, and it does not [=map/exist=],
+If |proof|.|type|,
+|proof|.|verificationMethod|, or
+|proof|.|proofPurpose| do not [=map/exist=],
```
```suggestion
If one or more of |proof|.|type|,
|proof|.|verificationMethod|, and
|proof|.|proofPurpose| does not [=map/exist=],
```

TallTed (changes requested):

index.html
msporny: I agree with you. Unfortunately, the exported reference in INFRA is "MIME type" (I have no idea why they did that)... I'll see if I can alias it to something else.

index.html
msporny: Done in 63631f56a77792ac48021dd472fb59cb2cf9db76.

index.html
msporny: Fixed in 47e280e0a1abc91b5f9ff7a3623fb39cabecc72e.

Wind4Greg (approved):

index.html
```diff
@@ -2435,37 +2435,30 @@ <h2>Cryptographic Suites</h2>
...
-            <dt><dfn>hash</dfn></dt>
-            <dd>A [=hashing algorithm=] that takes a [=JSON object=] as input and returns a [=byte
-            sequence=].</dd>
-            <dt><dfn>proof creation options</dfn></dt>
+The specification MUST define a <dfn class="export"
+data-lt="cryptosuite instantiation algorithm">data integrity cryptographic
+suite instantiation algorithm</dfn> that accepts a set of options
+([=map=] |options|) and returns a [=cryptosuite instance=]
+([=struct=] |cryptosuite|). This algorithm SHOULD be listed in [[[VC-SPECS]]].
+A <dfn class="export" data-lt="cryptosuite instance">data integrity
```
jyasskin: Note that [`data-lt`](https://respec.org/docs/#data-lt) will capture links to [=cryptosuite instance=] from any spec. If you just want a shorthand for this spec to use, use [`data-local-lt`](https://respec.org/docs/#data-local-lt).

index.html
```diff
@@ -3026,145 +3004,134 @@ <h3>Add Proof Set/Chain</h3>
...
 If |challenge| was given, and it does not match
-|proof|["`challenge`"],  an error MUST be raised and SHOULD
+|proof|.|challenge|, an error MUST be raised and SHOULD
 convey an error type of
 <a href="#INVALID_CHALLENGE_ERROR">INVALID_CHALLENGE_ERROR</a>.
           </li>
           <li>
+Let |cryptosuiteVerificationResult| be the result of running the
+|cryptosuite|.[=cryptosuite instance/verifyProof=] algorithm with
+|securedDocument| provided as input.
```
jyasskin, line 3163: `verifyProof` is defined to be able to return an error. This algorithm needs to handle that error.

index.html
```diff
@@ -2435,37 +2435,30 @@ <h2>Cryptographic Suites</h2>
...
+An algorithm that takes an [=unsecured data document=] ([=map=]
+|unsecuredDocument|) and proof options ([=map=] |options|) as input, and
+produces a [=data integrity proof=] ([=map=]) or an error.
+            </dd>
+            <dt><dfn>verifyProof</dfn></dt>
+            <dd>
+An algorithm that takes a [=secured data document=] ([=map=]
+|securedDocument|) as input, and produces a verification result
+([=map=]) or an error. The verification result MUST contain a
+verification status ([=boolean=] |verified|) value and an [=unsecured data
```
jyasskin: This sets up 3 kinds of return values:
1. error
1. non-error with _verified_==false
1. non-error with _verified_==true

If the first two are equivalent, you should collapse them, probably by removing the _verified_ field. If they're different, you should explain what the distinction is.

If the _verified_ field sticks around, you should probably require the _verifiedDocument_ to only be present if it's true?

index.html
```diff
@@ -2435,37 +2435,30 @@ <h2>Cryptographic Suites</h2>
...
-              a key identifier or an asynchronous signing algorithm allows implementations to take
-              advantage of hardware security modules (HSMs) to get better security.</p>
+An algorithm that takes an [=unsecured data document=] ([=map=]
+|unsecuredDocument|) and proof options ([=map=] |options|) as input, and
+produces a [=data integrity proof=] ([=map=]) or an error.
+            </dd>
+            <dt><dfn>verifyProof</dfn></dt>
+            <dd>
+An algorithm that takes a [=secured data document=] ([=map=]
+|securedDocument|) as input, and produces a verification result
```
jyasskin: I'd usually use "returns" for an algorithm instead of "produces" or "outputs" (see https://infra.spec.whatwg.org/#algorithm-declaration), but I also don't insist on it.

```suggestion
|securedDocument|) as input, and returns a verification result
```

index.html
```diff
@@ -2435,37 +2435,30 @@ <h2>Cryptographic Suites</h2>
...
+([=map=]) or an error. The verification result MUST contain a
+verification status ([=boolean=] |verified|) value and an [=unsecured data
+document=] ([=map=] |verifiedDocument|), and MAY contain information
```
jyasskin: I'd make "verifiedDocument" into a `<dfn>`. If there's more than one defined key, I'd then use a `<dl>` to organize them, but you may only have the 1 after handling my previous comment.

```suggestion
<dfn data-dfn-for="cryptosuite instance/verifyProof">verifiedDocument</dfn>
[=map/key=] whose [=map/value=] is an [=unsecured data
document=]. Other keys MAY contain information
```

jyasskin (commented):
I haven't done a thorough review yet, but please don't block on me.

iherman (approved):

index.html
dlongley: I still haven't done a thorough review here yet, but wanted to note that the first two here aren't equivalent -- so we should explain this a bit more (as @jyasskin says). An error means the verification check couldn't even be completed (because some error prevented it), whereas a non-error with `verified=false` means the process was completed and verification was *false*. This might be meaningful in cases where a network error occurred, but verification could actually become *true* if, for example, network access were restored.

index.html
msporny: Thanks, that's better, fixed in c46e47dab642204822d4648b7daeb47780afe7b1.

index.html
msporny: Fixed in 229663465813a71b117db52dad797591d73147e3.

index.html
msporny: I was trying to avoid "returns" as we haven't defined an error structure to return; we depend on functions throwing when there are errors/exceptions. It's debatable whether or not that is a "return" value or an exception value since programming environments typically have different syntax for handling each path.

I admit to having struggled with this particular part of documenting lower-level functions... I don't want to insist that developers use return structures for errors (because languages that prefer throwing exceptions exist). "Produces" feels like it gives the implementer more latitude to make the choice, where "returns" feels like we're taking exception handling off of the table.

I know that's a bit esoteric, but that's the current approach that many of the algorithms have taken. Changing it to "return" would require a rewrite to most/all of the algorithms. Do we have a best practices wrt. this particular concern at W3C?

index.html
msporny: Fixed in 1f7ee96701cccc2c1e00992ef61bc397d3bb8388.

index.html
msporny: Fixed in 1f7ee96701cccc2c1e00992ef61bc397d3bb8388.

msporny:

index.html
jyasskin: I think the Infra spec strives to document whatever best practices we have, but it's fine for WGs to override that in particular cases. Note that whatever you write, you're not insisting that developers use any particular structures for errors, due to https://infra.spec.whatwg.org/#algorithm-conformance:~:text=Conformance%20requirements%20phrased%20as%20algorithms%20or%20specific%20steps%20may%20be%20implemented%20in%20any%20manner%2C%20so%20long%20as%20the%20end%20result%20is%20equivalent. The  most egregious example I know of is https://html.spec.whatwg.org/multipage/infrastructure.html#starting-a-new-parallel-queue, which is specified as a busy wait. But again, I'm not insisting on changing this.

PR #259, opened 2024-03-29
(Minor, editorial) there was an error in the vocabulary diagram

iherman:
Tiny PR.

PR #261, opened 2024-04-28
Wrong reference for multibase in the diagram

iherman:
This is to replace #259 at @msporny's request


msporny (approved):

msporny:
Editorial, multiple positive reviews (via PR #259), no changes requested, no objections, merging.

davidlehn:
This PR changed most/all of the SVG `<text>` nodes to `<image>`s.  Was that intended?  Seems like it should have only changed a single word, like the previous PR was attempting to do.

iherman:
Sigh, thanks @davidlehn. I was too quick (but I saw that @msporny was eager to close down all pending PRs...)

Actually, the texts were also there. Draw.io has an option when generating svg, to copy some images into the svg _alongside_ the text itself (I presume as an alternative if the player cannot handle fonts via the foreign object construct for HTML snippets). The annoying thing is that this option is 'on' by default, and I forgot to uncheck it...

See #262 which should be o.k. now.

Thanks again.

PR #266, opened 2024-05-30
Update proof section properties

### LINKED ISSUE #232: cryptosuite missing from property list in 2.1 Proofs
See https://www.w3.org/TR/vc-data-integrity/#proofs

My understanding is that `cryptosuite` is mandatory for the proof elements.

wes-smith:
This issue is an attempt to address PR #232 by doing the following things: 

- add `cryptosuite` as a required property of a proof.
- update discussion of `verificationMethod`.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/266.html" title="Last updated on Jun 9, 2024, 3:47 PM UTC (15ff00b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/266/7de1987...15ff00b.html" title="Last updated on Jun 9, 2024, 3:47 PM UTC (15ff00b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -784,9 +784,9 @@ which was to merely logging into a website.
 
           <dt>verificationMethod</dt>
           <dd>
-If included, `verificationMethod`, the means and information needed to verify the proof,
-MUST be specified as a string that maps to a [[URL]]. An example of a verification method
-is a link to a [=public key=] which includes cryptographic material that is used
+A verification method is the means and information needed to verify the proof.
+If included, the value MUST be a string that maps to a [[URL]]. An example of a 
+verification method is a link to a [=public key=] which includes cryptographic material that is used
 by a verifier during the verification process. Inclusion of `verificationMethod` is OPTIONAL, 
 but if it is not included, other properties such as `cryptosuite` MUST provide a mechanism
 by which to obtain the required information to verify the proof.
@@ -794,8 +794,10 @@ by which to obtain the required information to verify the proof.
 
           <dt>cryptosuite</dt>
           <dd>
-Information about the cryptographic suite that was used to generate the proof. See
-[[[#cryptographic-suites]]] for more information. `cryptosuite` is REQUIRED.
+An identifier for the cryptographic suite that can be used to verify the proof. See
+[[[#cryptographic-suites]]] for more information. If the proof type is
+`DataIntegrityProof`, `cryptosuite` MUST be specified; otherwise,  `cryptosuite`
+MAY be specified. If specified, its value MUST be a string.
           </dd>
 
           <dt><dfn class="lint-ignore">created</dfn></dt>
@@ -4533,7 +4535,7 @@ Narrowed date time values to use `dateTimeStamp`.
 Ensure that base URL is set to null when processing as RDF.
         </li>
         <li>
-Updated requirements for `proof` objects.
+Added requirement that `DataIntegrityProof` objects need to contain the `cryptosuite` property.
         </li>
       </ul>
     </section>
--- a/index.html
+++ b/index.html
@@ -4532,6 +4532,9 @@ Narrowed date time values to use `dateTimeStamp`.
         <li>
 Ensure that base URL is set to null when processing as RDF.
         </li>
+        <li>
+Updated requirements for `proof` objects.
+        </li>
       </ul>
     </section>
     <section class="appendix informative">
--- a/index.html
+++ b/index.html
@@ -785,11 +785,12 @@ which was to merely logging into a website.
           <dt>verificationMethod</dt>
           <dd>
 A verification method is the means and information needed to verify the proof.
-If included, the value MUST be a string that maps to a [[URL]]. An example of a 
-verification method is a link to a [=public key=] which includes cryptographic material that is used
-by a verifier during the verification process. Inclusion of `verificationMethod` is OPTIONAL, 
-but if it is not included, other properties such as `cryptosuite` MUST provide a mechanism
-by which to obtain the required information to verify the proof.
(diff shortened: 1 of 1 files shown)


dlongley (approved):

index.html
```diff
@@ -784,10 +784,18 @@ <h3>Proofs</h3>
...
-The means and information needed to verify the proof MUST be specified as a
-string that maps to a [[URL]]. An example of a verification method is a
-link to a [=public key=] which includes cryptographic material that is used
-by a verifier during the verification process.
+If included, `verificationMethod`, the means and information needed to verify the proof,
+MUST be specified as a string that maps to a [[URL]]. An example of a verification method
+is a link to a [=public key=] which includes cryptographic material that is used
+by a verifier during the verification process. Inclusion of `verificationMethod` is OPTIONAL, 
+but if it is not included, other properties such as `cryptosuite` MUST provide a mechanism
+by which to obtain the required information to verify the proof.
```
msporny, line 792:
```suggestion
A verification method is the means and information needed to verify the proof.
If included, the value MUST be a string that maps to a [[URL]]. An example of a 
verification method is a link to a [=public key=] which includes cryptographic material that is used
by a verifier during the verification process. Inclusion of `verificationMethod` is OPTIONAL, 
but if it is not included, other properties such as `cryptosuite` MUST provide a mechanism
by which to obtain the required information to verify the proof.
```

index.html
```diff
@@ -784,10 +784,18 @@ <h3>Proofs</h3>
...
           <dt>verificationMethod</dt>
           <dd>
-The means and information needed to verify the proof MUST be specified as a
-string that maps to a [[URL]]. An example of a verification method is a
-link to a [=public key=] which includes cryptographic material that is used
-by a verifier during the verification process.
+If included, `verificationMethod`, the means and information needed to verify the proof,
+MUST be specified as a string that maps to a [[URL]]. An example of a verification method
+is a link to a [=public key=] which includes cryptographic material that is used
+by a verifier during the verification process. Inclusion of `verificationMethod` is OPTIONAL,
```
```suggestion
A verification method is the means and information needed to verify the proof.
If included, the value MUST be a string that maps to a [[URL]]. An example of a 
verification method is a link to a [=public key=] which includes cryptographic material that is used
by a verifier during the verification process. Inclusion of `verificationMethod` is OPTIONAL, 
```

index.html
msporny: This "MUST" statement is not testable. How would we write a test to verify that a system is conformant to this mandatory requirement? This seems to be a normative statement on a specification requirement, which we should avoid unless it's absolutely necessary.

What is this language attempting to achieve?

index.html
```diff
@@ -784,10 +784,18 @@ <h3>Proofs</h3>
...
+is a link to a [=public key=] which includes cryptographic material that is used
+by a verifier during the verification process. Inclusion of `verificationMethod` is OPTIONAL, 
+but if it is not included, other properties such as `cryptosuite` MUST provide a mechanism
+by which to obtain the required information to verify the proof.
+          </dd>
+
+          <dt>cryptosuite</dt>
+          <dd>
+Information about the cryptographic suite that was used to generate the proof. See
+[[[#cryptographic-suites]]] for more information. `cryptosuite` is REQUIRED.
```
```suggestion
An identifier for the cryptographic suite that can be used to verify the proof. See
[[[#cryptographic-suites]]] for more information. `cryptosuite` MUST be specified
and its value MUST be a string.
```

index.html
```diff
@@ -4524,6 +4532,9 @@ <h2>Revision History</h2>
         <li>
 Ensure that base URL is set to null when processing as RDF.
         </li>
+        <li>
+Updated requirements for `proof` objects.
```
```suggestion
Added requirement that `DataIntegrityProof` objects need to contain the `cryptosuite` property.
```

index.html
msporny: Hmm, on second glance, this is wrong. `cryptosuite` is not a requirement for general Proof objects, they are a requirement for `DataIntegrityProofs` which is a specific type of Proof object. This requirement would break all legacy implementations.

msporny (changes requested):
This PR seems like its incorrect, it adds a `cryptosuite` requirement to all Proof objects, which would break all legacy implementations. We need to ensure that the `cryptosuite` property is only required for `DataIntegrityProof` -- am I misreading this PR?

index.html
dlongley: Perhaps this needs to say that a cryptosuite specification is required to provide the details for how to obtain the verification material, not using MUST language, but still making it a requirement on a cryptosuite spec (and not a detail that could be tested against an implementation).

index.html
dlongley: Yeah, missed this, `cryptosuite` is only required when the type is `DataIntegrityProof`.

TallTed: Does this work, @msporny, @dlongley?
```suggestion
An identifier for the cryptographic suite that can be used to verify the proof. See
[[[#cryptographic-suites]]] for more information. If the proof type is
`DataIntegrityProof`, `cryptosuite` MUST be specified; otherwise,  `cryptosuite`
MAY be specified. If specified, its value MUST be a string.
```

TallTed (changes requested):

index.html
iherman: Agree with @dlongley. The change proposed by @TallTed below seems to cover the type of normative requirement we have, so a MUST language here seems to be superfluous...

msporny: Fixed in 15ff00b19f64b27ccf42cb6d20a1a2f6189b638a.

msporny (approved):

msporny:
Debatably normative, multiple reviews, changes requested and made, no objections, merging.

Issue #190, opened 2023-09-14
Add explanation about why proof graphs are used


msporny:
The VC Data Model added an explanation related to VC graphs in PR https://github.com/w3c/vc-data-model/pull/1280. The Data Integrity specification should do the same.

msporny:
This request is also related to a request by @iherman in this comment: https://github.com/w3c/vc-data-model/pull/1403#pullrequestreview-1797364583

iherman:
This is also related to the discussion in https://github.com/w3c/vc-data-model/pull/1397#pullrequestreview-1815677016: the fact that the proof property's value is a named graph, i.e., that the property refers to a container property in JSON-LD parlance, must be clearly specified in the DI specification (it was, so far, part of the VCDM spec, but has been removed as part of https://github.com/w3c/vc-data-model/pull/1397).

msporny:
PR #270 has been raised to address this issue. This issue will be closed once PR #270 has been merged.

msporny:
PR #270 has been merged, closing.

PR #270, opened 2024-06-20
Add explanation about proof graphs

msporny:
This PR is an attempt to address issue #190 by adding an explanation about why proof graphs are used.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/270.html" title="Last updated on Jun 29, 2024, 12:26 PM UTC (931419a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/270/f8b1155...931419a.html" title="Last updated on Jun 29, 2024, 12:26 PM UTC (931419a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1066,14 +1066,14 @@ another proof with a `previousProof` value that identifies the previous proof.
         <h3>Proof Graphs</h3>
 
         <p>
-When securing a document, it is important to clearly delinate the data being
+When securing a document, it is important to clearly delineate the data being
 protected, called the <dfn class="export">default graph</dfn>, from the data
 associated with the securing mechanism, called a <dfn class="export">proof
 graph</dfn>. Doing so enables the processing algorithms to be able to
 deterministically protect and verify a secured document. This is especially
 important because a single document can contain multiple proofs that might be
 independent of one another and a mechanism needs to be used to clearly
-dilineate one proof from another.
+delineate one proof from another.
         </p>
 
         <p>
@@ -1089,7 +1089,7 @@ property of the document is a separate graph, called a
         </p>
         <p>
 Using these graphs has a concrete effect when performing JSON-LD processing,
-which properly separates graph node identifiers in one graph from those in
+which properly separates statements expressed in one graph from those in
 another graph. Implementers that limit their inputs to other document types,
 such as JSON, YAML, or CBOR, will also need to keep this in mind if they merge
 data from one document with data from another, such as when an `id` value is the
--- a/index.html
+++ b/index.html
@@ -1061,7 +1061,46 @@ another proof with a `previousProof` value that identifies the previous proof.
         </pre>
 
       </section>
+
+      <section>
+        <h3>Proof Graphs</h3>
+
+        <p>
+When securing a document, it is important to clearly delinate the data being
+protected, called the <dfn class="export">default graph</dfn>, from the data
+associated with the securing mechanism, called a <dfn class="export">proof
+graph</dfn>. Doing so enables the processing algorithms to be able to
+deterministically protect and verify a secured document. This is especially
+important because a single document can contain multiple proofs that might be
+independent of one another and a mechanism needs to be used to clearly
+dilineate one proof from another.
+        </p>
+
+        <p>
+To ensure that information from different [=data integrity proofs=] are not
+accidentally co-mingled, the concept of a [=proof graph=]
+is used to encapsulate each [=data integrity proof=]. For simple documents,
+i.e., when the document contains some data with, possibly, associated proofs,
+this graph is the [=default graph=]. Each value associated with the `proof`
+property of the document is a separate graph, called a
+<dfn class="export">named graph</dfn>, of type
+<dfn class="export">ProofGraph</dfn>, which contains a single
+[=data integrity proof=].
+        </p>
+        <p>
+Using these graphs has a concrete effect when performing JSON-LD processing,
+which properly separates graph node identifiers in one graph from those in
+another graph. Implementers that limit their inputs to other document types,
+such as JSON, YAML, or CBOR, will also need to keep this in mind if they merge
+data from one document with data from another, such as when an `id` value is the
+same in both documents. It is important to not merge objects that seem to have
+similar properties but do not contain an `id` property, or other type of global
+identifier such as a URL, as it is not possible to tell if two unidentified
+objects are expressing information about the same object or not.
+        </p>
       </section>
+
+    </section>
 
       <section>
         <h3>Proof Purposes</h3>
--- a/index.html
+++ b/index.html
@@ -1067,20 +1067,19 @@ another proof with a `previousProof` value that identifies the previous proof.
 
         <p>
 When securing data in a document, it is important to clearly delineate the data
-being protected, called the <dfn class="export">default graph</dfn>, from the
-data associated with a securing mechanism, called a
-<dfn class="export">proof graph</dfn>. Creating this separation enables the
-processing algorithms to be able to deterministically protect and verify a
-secured document.
+being protected, which is every graph expressed in the document except the
+one containing the data associated with a securing mechanism, which is called
+a <dfn class="export">proof graph</dfn>. Creating this separation enables the
+processing algorithms to be able to deterministically protect and verify a secured
+document.
         </p>
 
         <p>
 For an unsecured document, the information contained in the document before a
-[=data integrity proof=] is added to the document is called the [=default
-graph=]. A single document can contain multiple proof graphs. To ensure that
-information from different [=data integrity proofs=] are not accidentally
-co-mingled, the concept of a [=proof graph=] is used to encapsulate each [=data
-integrity proof=]. Each value associated with the `proof` property of the
+[=data integrity proof=] is added to the document is expressed in one or more
+graphs. To ensure that information from different [=data integrity proofs=] are not
+accidentally co-mingled, the concept of a [=proof graph=] is used to encapsulate
+each [=data integrity proof=]. Each value associated with the `proof` property of the
 document is a separate graph, which is sometimes referred to as a
 <dfn class="export">named graph</dfn>, of type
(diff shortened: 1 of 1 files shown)


msporny:
@iherman I'll note that we're duplicating terminology between VCDM and this spec in this PR, but I have to get the exports working in this spec to re-direct the VCDM terminology references to this specification. All that to say, there will be some duplication until this PR is merged and the VCDM adjustments are made. Also note that the concept of a ProofGraph has been added, where we might need a reference added to the Security vocabulary?

index.html
```diff
@@ -1058,8 +1058,47 @@ <h3>Proof Chains</h3>
...
+i.e., when the document contains some data with, possibly, associated proofs,
+this graph is the [=default graph=]. Each value associated with the `proof`
+property of the document is a separate graph, called a
+<dfn class="export">named graph</dfn>, of type
+<dfn class="export">ProofGraph</dfn>, which contains a single
+[=data integrity proof=].
+        </p>
+        <p>
+Using these graphs has a concrete effect when performing JSON-LD processing,
+which properly separates graph node identifiers in one graph from those in
```
```suggestion
which properly separates statements expressed in one graph from those in
```

dlongley (approved):

longpd (approved):

Wind4Greg (approved):

index.html
```diff
@@ -1058,8 +1058,47 @@ <h3>Proof Chains</h3>
         </pre>
 
       </section>
+
+      <section>
+        <h3>Proof Graphs</h3>
+
+        <p>
+When securing a document, it is important to clearly delinate the data being
```
peacekeeper:
```suggestion
When securing a document, it is important to clearly delineate the data being
```

index.html
```diff
@@ -1058,8 +1058,47 @@ <h3>Proof Chains</h3>
...
+protected, called the <dfn class="export">default graph</dfn>, from the data
+associated with the securing mechanism, called a <dfn class="export">proof
+graph</dfn>. Doing so enables the processing algorithms to be able to
+deterministically protect and verify a secured document. This is especially
+important because a single document can contain multiple proofs that might be
+independent of one another and a mechanism needs to be used to clearly
+dilineate one proof from another.
```
peacekeeper:
```suggestion
delineate one proof from another.
```

peacekeeper (approved):
Two typos I think? Otherwise, this feels like a useful clarification.

iherman:
> @iherman I'll note that we're duplicating terminology between VCDM and this spec in this PR, but I have to get the exports working in this spec to re-direct the VCDM terminology references to this specification. All that to say, there will be some duplication until this PR is merged and the VCDM adjustments are made. 

Understood. 

But the current text is, I believe, wrong 😒. It says:

> When securing a document, it is important to clearly delinate the data being protected, called the  default graph,… 

That, in general, is not (necessarily) true. It is only true for simple Credentials (or similar tree-formed linked data without the usage of named graphs of their own), when the set of claims to be secured _happens to be_ the default graph. But this is not the case for a Verifiable Presentation, whose default graph only contains the claims for the presentation as a whole, the credential within the presentation is a separate (named) graph, and then there are proof graphs of that credential that must also be proven by the presentation's proof graph. See [figure 9](https://www.w3.org/TR/vc-data-model-2.0/#info-graph-vp) in the VC spec. The [relevant section in the VCDM](https://www.w3.org/TR/vc-data-model-2.0/#securing-mechanism-specifications) describes the mechanism more precisely.

*If* we want to formulate it in more general terms in the DI spec, we may be on a slippery slope. The DI spec, so far, does not say how one designates the "subject dataset" to be secured by a proof. It is formally defined for VCs and VPs in the aforementioned section but, in general, we do not have a standard mechanism. (This was the essence of the discussion with the late Henry Story.)

I am not sure how you want to play this at this point. One approach is to say in the DI spec, that any application area have to provide their own definition on what exactly is secured by a `ProofGraph`. We can refer back (probably within a note, not to make it normative) to the aforementioned section for the VC and VP cases. 

Alternatively, we move the whole definition here from the VCDM, but this then becomes a major surgery, because most of the diagrams (and their accompanying texts) in §3 would become more difficult to understand, and may have to be moved to the DI spec...

> Also note that the concept of a ProofGraph has been added, where we might need a reference added to the Security vocabulary?

I am not sure what you mean. The security vocabulary already has the notion of a [`ProofGraph`](https://w3c.github.io/vc-data-integrity/vocab/security/vocabulary.html#ProofGraph). Do you mean that the definition of that vocabulary (which is very terse) should refer to this new section? Then yes, this is to be done, possibly as part of this PR. I am happy to do it.

msporny:
@iherman wrote:
> The security vocabulary already has the notion of a ProofGraph. 

Excellent! I forgot that existed. :)  -- (I'm sure you're seeing a pattern for today, Ivan -- my head is clearly not screwed on straight today)

> But the current text is, I believe, wrong 😒. 

Yes, you're right... let me try to see if I can do something in DI that doesn't require the major surgery you mention above.

msporny:
> we do not have a standard mechanism. (This was the essence of the discussion with the late Henry Story.)
> One approach is to say in the DI spec, that any application area have to provide their own definition on what exactly is secured by a ProofGraph.

@iherman, I have made an attempt at a fix in 8544fa74b77552e5f31a2f46b1f8967656c935e0 that defines the "default graph" (as far as DI is concerned) as "For an unsecured document, the information contained in the document before a [=data integrity proof=] is added to the document". So, we don't run afoul of where the boundary is (the boundary is the document and goes no further than that, avoiding Henry and Dan's concerns, IIRC)... and what is contained (it's everything in the document before it was secured, which might include references to other named graphs, which are also secured).

Does that formulation work for you? If not, please suggest some text that might fix it. For now, ignore what we say in VCDM and see if this text can stand on its own... if it can, perhaps we can add text that says "other specs using DI can more accurately define what the default graph entails and what the proof graph secures. Thoughts?

thereisnogabe (approved):

iherman:
> 
> @iherman, I have made an attempt at a fix in [8544fa7](https://github.com/w3c/vc-data-integrity/commit/8544fa74b77552e5f31a2f46b1f8967656c935e0) that defines the "default graph" (as far as DI is concerned) as "For an unsecured document, the information contained in the document before a [=data integrity proof=] is added to the document". So, we don't run afoul of where the boundary is (the boundary is the document and goes no further than that, avoiding Henry and Dan's concerns, IIRC)... and what is contained (it's everything in the document before it was secured, which might include references to other named graphs, which are also secured).
> 
> Does that formulation work for you? 

Almost, but not fully. The whole section could work, _except for the term "default graph"._ It does not, because:

The term "default graph" is formally defined in the [RDF specification](https://www.w3.org/TR/rdf12-concepts/#dfn-default-graph). We should not redefine that term, and we should refer to it if and only if the reference is correct. (We were very cautious in the VCDM spec in this respect!) And the problem is that, in the case of a Verifiable Presentation, that "thing" is not the default graph but, rather, an [(RDF) dataset](https://www.w3.org/TR/rdf12-concepts/#dfn-rdf-dataset).

So the text works for me, _but only if we use a different term_. 

I must admit, I am not sure what the right term could be. AFAIK, the crypto people sometimes use the term "plaintext" for the "thing" that is being signed, so maybe something like the "plain graph" may work (I am holding my nose over the fact that it is not a graph...)? Or, if I forget about holding my nose, "plain graphs" (knowing that the plural is not necessarily justified)? But that is just an idea, I trust you will find a better term...

index.html
```diff
@@ -1061,8 +1061,47 @@ <h3>Proof Chains</h3>
...
+      <section>
+        <h3>Proof Graphs</h3>
+
+        <p>
+When securing data in a document, it is important to clearly delineate the data
+being protected, called the <dfn class="export">default graph</dfn>, from the
+data associated with a securing mechanism, called a
+<dfn class="export">proof graph</dfn>. Creating this separation enables the
+processing algorithms to be able to deterministically protect and verify a
+secured document.
```
dlongley: Attempting to address Ivan's concern:

```suggestion
When securing data in a document, it is important to clearly delineate the data
being protected, which is every graph expressed in the document except the
one containing the data associated with a securing mechanism, which is called
a <dfn class="export">proof graph</dfn>. Creating this separation enables the
processing algorithms to be able to deterministically protect and verify a secured
document.
```

index.html
```diff
@@ -1061,8 +1061,47 @@ <h3>Proof Chains</h3>
...
+For an unsecured document, the information contained in the document before a
+[=data integrity proof=] is added to the document is called the [=default
+graph=]. A single document can contain multiple proof graphs. To ensure that
+information from different [=data integrity proofs=] are not accidentally
+co-mingled, the concept of a [=proof graph=] is used to encapsulate each [=data
+integrity proof=]. Each value associated with the `proof` property of the
+document is a separate graph, which is sometimes referred to as a
+<dfn class="export">named graph</dfn>, of type
+<dfn class="export">ProofGraph</dfn>, which contains a single [=data integrity
+proof=].
```
dlongley: Also aimed at addressing Ivan's concerns:

```suggestion
For an unsecured document, the information contained in the document before a
[=data integrity proof=] is added to the document is expressed in one or more
graphs. To ensure that information from different [=data integrity proofs=] are not
accidentally co-mingled, the concept of a [=proof graph=] is used to encapsulate
each [=data integrity proof=]. Each value associated with the `proof` property of the
document is a separate graph, which is sometimes referred to as a
<dfn class="export">named graph</dfn>, of type
<dfn class="export">ProofGraph</dfn>, which contains a single [=data integrity
proof=].
```

iherman:
To be clear, the proposed changes by @dlongley (in https://github.com/w3c/vc-data-integrity/pull/270#discussion_r1659102575 and https://github.com/w3c/vc-data-integrity/pull/270#discussion_r1659107254) solve my problem.

Thanks @dlongley!

msporny:

Issue #288, opened 2024-07-16
Clarify the processing model


msporny:
From https://github.com/w3c/vc-data-integrity/issues/272#issuecomment-2212258255, @decentralgabe wrote:
> Update DI spec language to provide a transparent, step-by-step processing model for verifiers that emphasizes secure handling of contexts and transformation of documents to trusted forms before interpretation. We can highlight the potential attacks/vulnerabilities at each step and recommend mitigations.

msporny:
PR #291 has been raised to address this issue. This issue will be closed once PR #291 has been merged.

msporny:
PR #291 has been merged, closing.

PR #291, opened 2024-07-21
Add sections on Processing Model and Context Validation

msporny:
This PR attempts to address issue #288 by adding sections on Processing Model and Context Validation.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/291.html" title="Last updated on Jul 27, 2024, 5:19 PM UTC (334d1de)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/291/acba0d7...334d1de.html" title="Last updated on Jul 27, 2024, 5:19 PM UTC (334d1de)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1451,6 +1451,70 @@ Verifiable Credentials v2.0: Vocabularies</a>.
         </p>
 
         <section>
+          <h3>Validating Contexts</h3>
+
+          <p>
+It is necessary to ensure that a consuming application has explicitly approved
+of the types, and therefore the semantics, of input documents that it will
+process. Not checking JSON-LD context values against known good values can lead
+to security vulnerabilities, due to variance in the semantics that they convey.
+Applications MUST use the algorithm in Section [[[#context-validation]]], or one
+that achieves equivalent protections, to validate contexts in a [=conforming
+secured document=]. Context validation MUST be run <em>after</em> running the
+applicable algorithm in either Section [[[#verify-proof]]] or Section
+[[[#verify-proof-sets-and-chains]]].
+          </p>
+
+          <p>
+While the algorithm described in Section [[[#context-validation]]] provides one
+way of checking context values, and one optional way of safely processing
+unknown context values, implementers MAY use alternative approaches, or a
+different ordering of the steps, that provide the same protections.
+          </p>
+
+          <p>
+For example, if no JSON-LD processing is to occur, then, rather than performing
+this check, an application could follow the guidance in whatever trusted
+documentation is provided out of band for properly understanding the semantics
+of that type of document.
+          </p>
+
+          <p>
+Another approach would be to configure an application to use a JSON-LD Context
+loader, sometimes referred to as a document loader, to use only local copies of
+approved context files. This would guarantee that neither the context files nor
+their cryptographic hashes would ever change, effectively resulting in the same
+result as the algorithm in Section [[[#context-validation]]].
+          </p>
+
+          <p>
+Another alternative approach, also effectively equivalent to the algorithm in
+Section [[[#context-validation]]], would be for an application to keep a list of
+well known context URLs and their associated approved cryptographic hashes,
+without storing every context file locally. This would allow these contexts to
+be safely loaded from the network without compromising the security expectations
+of the application.
+          </p>
+
+          <p>
+Yet another valid approach would be for a transmitting application to
+<a data-cite="JSON-LD11-API#compaction-algorithm">compact</a> a document to
+exactly what a receiving application requests, via a protocol such as one
+requesting a [=verifiable presentation=], omitting additional sender-specific
+context values that were used when securing the original document. As long as
+the cryptography suite's verification algorithm provides a successful
+verification result, such transformations are valid and would result in full
+URLs for terms that were previously compacted by the omitted context. That is, a
+term that was previously compacted to `foo` based on a sender-supplied context
+that is unknown to a receiver (e.g., ``https://ontology.example/v1`) would
+instead be "expanded" to a URL like `https://ontology.example#foo`, which would
+then be "compacted" to the same URL, once the unknown context is omitted and the
+JSON-LD compaction algorithm is applied by the receiving application.
+          </p>
+
+        </section>
+
+        <section>
           <h3>Context Injection</h3>
 
           <p>
@@ -2237,14 +2301,10 @@ Return |combinedVerificationResult|, |successfulVerificationResults|.
         <h3>Context Validation</h3>
 
         <p>
-The purpose of the algorithm in this section is to ensure that a consuming
-application has explicitly approved of the types, and therefore the semantics,
-of input documents that it will process. Not checking JSON-LD context values
-against known good values can lead to security vulnerabilities, due to variance
-in the semantics that they convey. Applications MAY use this context validation
-algorithm to validate contexts in a [=conforming secured document=]. If used,
-it MUST be run <em>after</em> running the applicable algorithm in either
-Section [[[#verify-proof]]] or Section [[[#verify-proof-sets-and-chains]]].
+The following algorithm provides one mechanism that can be used to ensure that
+an application understands the contexts associated with a document before it
+executed business rules specific to the input in the document. For more
+rationale related to this algorithm, see Section [[[#validating-contexts]]].
 This algorithm takes inputs of a document ([=map=] |inputDocument|), a set of
 approved JSON-LD Contexts ([=map=] |approvedContexts|), and a boolean to
 recompact when unknown contexts are detected ([=boolean=] |recompact|), and
@@ -2306,51 +2366,6 @@ If |result|.|errors| is empty, set |result|.|status| to `true`; otherwise, set
 Return the value of |result|.
           </li>
         </ol>
-
-        <p>
-While the algorithm above provides one way of checking these values, and one
-optional way of safely processing unknown values, implementers MAY use
-alternative approaches, or a different ordering of the steps, that provide the
-same protections.
-        </p>
-
-        <p>
-For example, if no JSON-LD processing is to occur, then, rather than performing
-this check, an application could follow the guidance in whatever trusted
-documentation is provided out of band for properly understanding the semantics
-of that type of document.
-        </p>
-
-        <p>
-Another approach would be to configure an application to use a JSON-LD Context
-loader, sometimes referred to as a document loader, to use only local copies of
-approved context files. This would guarantee that neither the context files nor
-their cryptographic hashes would ever change, effectively resulting in the
-same result as the algorithm above.
-        </p>
-
-        <p>
-Another alternative approach, also effectively equivalent to the algorithm provided
-in this section, would be for an application to keep a list of well known context
-URLs and their associated approved cryptographic hashes, without storing every
-context file locally. This would allow these contexts to be safely loaded from
-the network without compromising the security expectations of the application.
-        </p>
-
-        <p>
-Yet another valid approach would be for a transmitting application to
-<a data-cite="JSON-LD11-API#compaction-algorithm">compact</a> a document to
-exactly what a receiving application requests, omitting additional
-sender-specific context values that were used when securing the original document.
-As long as the cryptography suite's verification algorithm provides a successful
-verification result, such transformations are valid and would result in full
-URLs for terms that were previously compacted by the omitted context. That is, a
-term that was previously compacted to `foo` based on a sender-supplied context
-that is unknown to a receiver (e.g., ``https://ontology.example`) would instead
-be "expanded" to a URL like `https://ontology.example#foo`, which would then
-be "compacted" to the same URL, once the unknown context is omitted and the
-JSON-LD compaction algorithm is applied by the receiving application.
-        </p>
 
       </section>
 
--- a/index.html
+++ b/index.html
@@ -1837,6 +1837,59 @@ measures that could contribute to a safer and more efficient ecosystem.
       </p>
 
       <section class="normative">
+        <h3>Processing Model</h3>
+
+        <p>
+The processing model used by a [=conforming processor=] and its
+application-specific software is described in this section. When software
+needs to transmit information in a way that protects the data from being
+tampered with, it:
+        </p>
+
+        <ol>
+          <li>
+Arranges the information into a document, such as a JSON or JSON-LD document.
+          </li>
+          <li>
+Selects one or more JSON-LD Contexts and expresses them using the `@context`
+property if the document is a JSON-LD document.
+          </li>
+          <li>
+Selects a cryptography suite that meets the needs of the use case, such as
+one that uses acceptable cryptographic key material and provides full
+disclosure, selective disclosure, or unlinkable disclosure functionality.
+          </li>
+          <li>
+Adds one or more proofs using the algorithms in Section [[[#add-proof]]] or
+Section [[[#add-proof-set-chain]]].
+          </li>
+        </ol>
+
+        <p>
+When software needs to use information that was transmitted to it using
+a mechanism described by this specification, it:
+        </p>
+
+        <ol>
+          <li>
+Transforms the incoming data into a document that can be understood by the
+algorithms provided in Section [[[#verify-proof]]] or Section
+[[[#verify-proof-sets-and-chains]]].
+          </li>
+          <li>
+Verifies the integrity of the incoming document using the algorithms provided in
+Section [[[#verify-proof]]] or Section [[[#verify-proof-sets-and-chains]]].
+          </li>
+          <li>
+Validates all JSON-LD Context values used in the document using the
+algorithm provided in Section [[[#context-validation]]] if the document is a
+JSON-LD document.
+          </li>
+        </ol>
+
+      </section>
+
+      <section class="normative">
         <h3>Add Proof</h3>
 
         <p>
--- a/index.html
+++ b/index.html
@@ -1842,51 +1842,53 @@ measures that could contribute to a safer and more efficient ecosystem.
         <p>
 The processing model used by a [=conforming processor=] and its
 application-specific software is described in this section. When software
-needs to ensure information is tamper-evident, it:
+is to ensure information is tamper-evident, it performs the following steps:
         </p>
 
         <ol>
           <li>
-Arranges the information into a document, such as a JSON or JSON-LD document.
+The software arranges the information into a document, such as a JSON or
+JSON-LD document.
           </li>
           <li>
-If the document is a JSON-LD document, selects one or more JSON-LD Contexts
-and expresses them using the `@context` property.
+If the document is a JSON-LD document, the software selects one or more
+JSON-LD Contexts and expresses them using the `@context` property.
           </li>
           <li>
-Selects a cryptography suite that meets the needs of the use case, such as
-one that uses acceptable cryptographic key material and provides full
-disclosure, selective disclosure, or unlinkable disclosure functionality.
+The software selects one or more cryptography suites that meet the needs of
+the use case, such as one that provides full, selective, or unlinkable
+disclosure, using acceptable cryptographic key material.
           </li>
           <li>
-Adds one or more proofs using the algorithms in Section [[[#add-proof]]] or
-Section [[[#add-proof-set-chain]]].
+The software uses the applicable algorithms in either Section [[[#add-proof]]]
+or Section [[[#add-proof-set-chain]]] to add one or more proofs.
           </li>
         </ol>
 
         <p>
 When software needs to use information that was transmitted to it using
-a mechanism described by this specification, it:
+a mechanism described by this specification, it performs the following steps:
         </p>
 
         <ol>
           <li>
-Transforms the incoming data into a document that can be understood by the
-algorithms provided in Section [[[#verify-proof]]] or Section
+The software transforms the incoming data into a document that can be understood
+by the applicable algorithm provided in Section [[[#verify-proof]]] or Section
 [[[#verify-proof-sets-and-chains]]].
           </li>
           <li>
-Validates that the incoming document follows an expected schema used by
-the application via JSON Schema or equivalent mechanism.
+The software uses JSON Schema or equivalent mechanism to validate that the
+incoming document follows an expected schema used by the application.
           </li>
           <li>
-Verifies the integrity of the incoming document using the algorithms provided in
-Section [[[#verify-proof]]] or Section [[[#verify-proof-sets-and-chains]]].
+The software uses the applicable algorithms provided in Section [[[#verify-proof]]]
+or Section [[[#verify-proof-sets-and-chains]]] to verify the integrity of the
+incoming document.
           </li>
           <li>
-If the document is a JSON-LD document, validates all JSON-LD Context
-values used in the document using the algorithm provided in Section
-[[[#context-validation]]].
+If the document is a JSON-LD document, the software uses the algorithm provided
+in Section [[[#context-validation]]] to validate all JSON-LD Context values used
+in the document.
           </li>
         </ol>
 
--- a/index.html
+++ b/index.html
@@ -2306,7 +2306,7 @@ an application understands the contexts associated with a document before it
 executed business rules specific to the input in the document. For more
 rationale related to this algorithm, see Section [[[#validating-contexts]]].
 This algorithm takes inputs of a document ([=map=] |inputDocument|), a set of
-approved JSON-LD Contexts ([=map=] |approvedContexts|), and a boolean to
+approved JSON-LD Contexts ([=map=] |expectedContext|), and a boolean to
 recompact when unknown contexts are detected ([=boolean=] |recompact|), and
 returns a [=map=] that contains the following:
         </p>
@@ -2366,6 +2366,12 @@ If |result|.|errors| is empty, set |result|.|status| to `true`; otherwise, set
 Return the value of |result|.
           </li>
         </ol>
+
+        <p>
+Implementations MAY include additional warnings or errors that enforce
+further validation rules that are specific to the implementation or a
+particular use case.
+        </p>
 
       </section>
 
--- a/index.html
+++ b/index.html
@@ -2267,21 +2267,31 @@ The context validation algorithm is as follows:
 Set |result|.|status| to `false`.
           </li>
           <li>
-For each |contextValue| expressed in every `@context` property, ensure that
-either:
-            <ol class="algorithm">
+For each |contextValue| expressed in every `@context` property:
+            <ul class="algorithm">
               <li>
-The |contextValue| exists in |approvedContexts| and matches a cryptographic hash
-associated with the matching approved context, or
+If the |contextValue| exists in |approvedContexts|:
+                <ol class="algorithm">
+                  <li>
+confirm that it matches a cryptographic hash associated with the matching
+approved context.
+                  </li>
+                </ol>
               </li>
               <li>
-optionally, for the context values that do not exist in |approvedContexts|,
-remove the |contextValue| from the `@context` property values. Set
-|result|.|document| to the result of running the
-<a data-cite="JSON-LD11-API#compaction-algorithm">
-JSON-LD Compaction Algorithm</a> with the modified |inputDocument| as input.
+If the context value does not exist in |approvedContexts|:
+                <ol class="algorithm">
+                  <li>
+Optionally, remove the |contextValue| from the `@context` property values. 
+                  </li>
+                  <li>
+Set |result|.|document| to the result of running the
+<a data-cite="JSON-LD11-API#compaction-algorithm"> JSON-LD Compaction
+Algorithm</a> with the modified |inputDocument| as input.
+                  </li>
+                </ol>
               </li>
-            </ol>
+            </ul>
           </li>
           <li>
 If no errors occurred in the previous step, set |result|.|status| to `true` and,
--- a/index.html
+++ b/index.html
@@ -2241,14 +2241,14 @@ The purpose of the algorithm in this section is to ensure that a consuming
 application has explicitly approved of the types, and therefore the semantics,
 of input documents that it will process. Not checking JSON-LD context values
 against known good values can lead to security vulnerabilities, due to variance
-in the semantics that they convey. The context validation algorithm is one
-that applications validating a [=conforming secured document=] MUST run after
-running the algorithm in Section [[[#verify-proof]]] or Section
-[[[#verify-proof-sets-and-chains]]]. This algorithm takes a document ([=map=]
-|inputDocument|), a set of approved JSON-LD Contexts ([=map=]
-|approvedContexts|), and a boolean to recompact when unknown contexts are
-detected ([=boolean=] |recompact|) and returns a [=map=] that contains the
-following:
+in the semantics that they convey. Applications MAY use this context validation
+algorithm to validate contexts in a [=conforming secured document=]. If used,
+it MUST be run <em>after</em> running the applicable algorithm in either
+Section [[[#verify-proof]]] or Section [[[#verify-proof-sets-and-chains]]].
+This algorithm takes inputs of a document ([=map=] |inputDocument|), a set of
+approved JSON-LD Contexts ([=map=] |approvedContexts|), and a boolean to
+recompact when unknown contexts are detected ([=boolean=] |recompact|), and
+returns a [=map=] that contains the following:
         </p>
 
         <ul>
@@ -2273,7 +2273,7 @@ The context validation algorithm is as follows:
         <ol class="algorithm">
           <li>
 Set |result|.|status| to `false`, |result|.|warnings| to an empty list,
-|result|.|errors| to an empty list, |compactionContext| to an empty list,
+|result|.|errors| to an empty list, |compactionContext| to an empty list;
 and clone |inputDocument| to |result|.|document|.
           </li>
           <li>
--- a/index.html
+++ b/index.html
@@ -2275,33 +2275,29 @@ Set |result|.|status| to `false`, |result|.|warnings| to an empty list,
 and clone |inputDocument| to |result|.|document|.
           </li>
           <li>
-For each subtree in |result|.|document| that contains a `@context` property:
+Let |contextValue| be the value of the `@context` property of |result|.|document|,
+which might be undefined.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1836,6 +1836,63 @@ <h2>Algorithms</h2>
...
       </p>
 
+      <section class="normative">
+        <h3>Processing Model</h3>
+
+        <p>
+The processing model used by a [=conforming processor=] and its
+application-specific software is described in this section. When software
+needs to transmit information in a way that protects the data from being
+tampered with, it:
```
```suggestion
application-specific software is described in this section. When software
needs to ensure information is tamper-evident, it:
```

index.html
```diff
@@ -2175,6 +2232,116 @@ <h3>Verify Proof Sets and Chains</h3>
...
+              <li>
+The |contextValue| exists in |approvedContexts| and matches a cryptographic hash
+associated with the matching approved context, or
+              </li>
+              <li>
+optionally, for the context values that do not exist in |approvedContexts|,
+remove the |contextValue| from the `@context` property values. Set
+|result|.|document| to the result of running the
+<a data-cite="JSON-LD11-API#compaction-algorithm">
+JSON-LD Compaction Algorithm</a> with the modified |inputDocument| as input.
```
dlongley: This will need the same adjustments as VC data model [PR 1535](https://github.com/w3c/vc-data-model/pull/1535).

index.html
```diff
@@ -2175,6 +2232,116 @@ <h3>Verify Proof Sets and Chains</h3>
...
+explicitly approved of the types of input documents that it will process. Not
+checking JSON-LD Context values against known good values, and the semantics
+that they convey, can lead to security vulnerabilities. While the algorithm
+above provides one way of checking these values, and one optional way of safely
+processing unknown values, there are other approaches that provide the same
+protections.
+        </p>
+
+        <p>
+For example, if no JSON-LD processing is to occur, this check can be ignored.
```
```suggestion
For example, if no JSON-LD processing is to occur, this check must be replaced
by using whatever trusted documentation is provided out of band for
properly understanding the semantics of the document.
```

index.html
```diff
@@ -1836,6 +1836,63 @@ <h2>Algorithms</h2>
...
+tampered with, it:
+        </p>
+
+        <ol>
+          <li>
+Arranges the information into a document, such as a JSON or JSON-LD document.
+          </li>
+          <li>
+Selects one or more JSON-LD Contexts and expresses them using the `@context`
+property if the document is a JSON-LD document.
```
```suggestion
If the document is a JSON-LD document, selects one or more JSON-LD Contexts
and expresses them using the `@context` property.
```

index.html
```diff
@@ -1836,6 +1836,63 @@ <h2>Algorithms</h2>
...
+the application via JSON Schema or equivalent mechanism.
+          </li>
+          <li>
+Verifies the integrity of the incoming document using the algorithms provided in
+Section [[[#verify-proof]]] or Section [[[#verify-proof-sets-and-chains]]].
+          </li>
+          <li>
+Validates all JSON-LD Context values used in the document using the
+algorithm provided in Section [[[#context-validation]]] if the document is a
+JSON-LD document.
```
```suggestion
If the document is a JSON-LD document, validates all JSON-LD Context
values used in the document using the algorithm provided in Section
[[[#context-validation]]].
```

index.html
```diff
@@ -2175,6 +2232,116 @@ <h3>Verify Proof Sets and Chains</h3>
...
+      <section class="normative">
+        <h3>Context Validation</h3>
+
+        <p>
+This section contains an algorithm that applications MUST run after running the
+algorithm in Section [[[#verify-proof]]] or Section
+[[[#verify-proof-sets-and-chains]]], when validating a [=conforming secured
+document=]. This algorithm takes a document ([=map=] |inputDocument|) and a set
+of approved JSON-LD Contexts ([=map=] |approvedContexts|) and returns a [=map=]
+that contains the following:
```
```suggestion
This section contains an algorithm that applications validating a [=conforming
secured document=] MUST run after running the algorithm in Section
[[[#verify-proof]]] or Section [[[#verify-proof-sets-and-chains]]]. This
algorithm takes a document ([=map=] |inputDocument|) and a set of approved
JSON-LD Contexts ([=map=] |approvedContexts|), and returns a [=map=] that
contains the following:
```

index.html
```diff
@@ -2175,6 +2232,116 @@ <h3>Verify Proof Sets and Chains</h3>
...
+        </p>
+
+        <ul>
+          <li>
+a status ([=boolean=] |status|)
+          </li>
+          <li>
+a validated document ([=map=] |document|)
```
TallTed: Do any `errors` prevent return of `a validated document`? That is, is `a validated document` *always* returned in the `map`, regardless of `errors`?

index.html
```diff
@@ -2175,6 +2232,116 @@ <h3>Verify Proof Sets and Chains</h3>
...
+occurred, set |result|.|status| to `false`, remove the |document| property from
+|result|, and add at least one error to |result|.|errors|.
+          </li>
+          <li>
+Return the value of |result|.
+          </li>
+        </ol>
+
+        <p>
+The purpose of the algorithm above is to ensure that a consuming application has
```
TallTed: This explanation (all paragraphs of it) would be better presented before the algorithm it discusses, than following it. Making this relocation will mandate some rephrasing. It might be worthwhile to put this description into `<h4>Context Validation Description</h4>` (or similar title) followed by `<h4>Context Validation Algorithm</h4>`, both being within `<h3>Context Validation</h4>`.

index.html
```diff
@@ -2175,6 +2232,116 @@ <h3>Verify Proof Sets and Chains</h3>
...
+associated with the matching approved context, or
+              </li>
+              <li>
+optionally, for the context values that do not exist in |approvedContexts|,
+remove the |contextValue| from the `@context` property values. Set
+|result|.|document| to the result of running the
+<a data-cite="JSON-LD11-API#compaction-algorithm">
+JSON-LD Compaction Algorithm</a> with the modified |inputDocument| as input.
+              </li>
+            </ol>
```
```suggestion
For each |contextValue| expressed in every `@context` property:
            <ul class="algorithm">
              <li>
If the |contextValue| exists in |approvedContexts|:
                <ol class="algorithm">
                  <li>
confirm that it matches a cryptographic hash associated with the matching
approved context.
                  </li>
                </ol>
              </li>
              <li>
If the context value does not exist in |approvedContexts|:
                <ol class="algorithm">
                  <li>
Optionally, remove the |contextValue| from the `@context` property values. 
                  </li>
                  <li>
Set |result|.|document| to the result of running the
<a data-cite="JSON-LD11-API#compaction-algorithm"> JSON-LD Compaction
Algorithm</a> with the modified |inputDocument| as input.
                  </li>
                </ol>
              </li>
            </ul>
```

index.html
```diff
@@ -2175,6 +2232,116 @@ <h3>Verify Proof Sets and Chains</h3>
...
+        <p>
+For example, if no JSON-LD processing is to occur, this check can be ignored.
+        </p>
+
+        <p>
+Another approach would be to configure an application to use a JSON-LD Context
+loader, sometimes referred to as a document loader, to use only local copies of
+approved context files which would guarantee that 1) context files will never
+change, and 2) the cryptographic hashes will never change, effectively resulting
+in the same result as the algorithm above.
```
```suggestion
approved context files. This would guarantee that neither context files nor
their cryptographic hashes would ever change, effectively resulting in the
same result as the algorithm above.
```

index.html
```diff
@@ -2175,6 +2232,116 @@ <h3>Verify Proof Sets and Chains</h3>
...
+        </ol>
+
+        <p>
+The purpose of the algorithm above is to ensure that a consuming application has
+explicitly approved of the types of input documents that it will process. Not
+checking JSON-LD Context values against known good values, and the semantics
+that they convey, can lead to security vulnerabilities. While the algorithm
+above provides one way of checking these values, and one optional way of safely
+processing unknown values, there are other approaches that provide the same
+protections.
```
```suggestion
The purpose of the algorithm above is to ensure that a consuming application has
explicitly approved of the types, and therefore the semantics, of input documents
that it will process. Not checking JSON-LD context values against known good values
can lead to security vulnerabilities, due to variance in the semantics that they
convey. While the algorithm above provides one way of checking these context values,
and one optional way of safely processing unknown values, there are other approaches
that can provide the same protections.
```

index.html
```suggestion
For example, if no JSON-LD processing is to occur, then, rather than performing
this check, an application could follow the guidance in whatever trusted
documentation is provided out of band for properly understanding the semantics
of that type of document.
```

index.html
```diff
@@ -2175,6 +2232,116 @@ <h3>Verify Proof Sets and Chains</h3>
...
+in the same result as the algorithm above.
+        </p>
+
+        <p>
+An alternative approach an application could use would be to use a list of well
+known context URLs and their associated approved cryptographic hashes, without
+storing every context file locally, which would allow contexts to be safely
+loaded and cached from the network without compromising the security
+expectations of the application. This approach would also be effectively
+equivalent to the algorithm provided in this section.
```
```suggestion
An alternative approach, also effectively equivalent to the algorithm provided in
this section, would be for an application to keep a list of well known context
URLs and their associated approved cryptographic hashes, without storing every
context file locally. This would allow contexts to be safely loaded from the
network without compromising the security expectations of the application.
```

Wind4Greg (approved):

thereisnogabe (approved):

index.html
```diff
@@ -2175,6 +2232,116 @@ <h3>Verify Proof Sets and Chains</h3>
...
+<a data-cite="JSON-LD11-API#compaction-algorithm">compact</a> a document to
+exactly what a receiving application requests, omitting additional
+sender-specific context values that it used when securing the original document.
+As long as the cryptography suite's verification algorithm provides a successful
+verification result, such transformations are valid and would result in full
+URLs for terms that were previously compacted by the omitted context. That is, a
+term that was previously compacted to `foo` based on a sender-supplied context
+that is unknown to a receiver would instead be "compacted" to something like
+`https://vocabulary.example#foo` once the unknown context is omitted and the
+compaction algorithm is applied by the receiving application.
```
TallTed: This needs more detailed description. `foo` becoming `https://vocabulary.example#foo` looks like an expansion, not a compaction. Can we (or an update to JSON-LD docs) show a full step-by-step example, starting with a JSON-LD document that includes "unrecognized" `@context` value(s), and resulting in a JSON-LD document that has only "recognized" `@context` value(s) (including )?

Also, it is unclear here what actions are taken by the transmitting application (and whether the transmitter is a holder or issuer — the former suggested by `values that it used when securing`, which I've changed to `values that were used when securing` because I think the transmitter is *either* holder or issuer), and what actions are taken by the receiving application (only being a verifier, I think?). Significantly, I don't see any indication of how a receiving application can request specific context(s) from a transmitting application, which would be necessary for the transmitter to omit `additional sender-specific context values`...
```suggestion
Yet another valid approach would be for a transmitting application to
<a data-cite="JSON-LD11-API#compaction-algorithm">compact</a> a document to
exactly what a receiving application requests, omitting additional
sender-specific context values that were used when securing the original document.
As long as the cryptography suite's verification algorithm provides a successful
verification result, such transformations are valid and would result in full
URLs for terms that were previously compacted by the omitted context. That is, a
term that was previously compacted to `foo` based on a sender-supplied context
that is unknown to a receiver would instead be "compacted" to something like
`https://vocabulary.example#foo` once the unknown context is omitted and the
compaction algorithm is applied by the receiving application.
```

TallTed (changes requested):

longpd (approved):

index.html
```diff
@@ -2175,6 +2232,116 @@ <h3>Verify Proof Sets and Chains</h3>
...
       </section>
 
+      <section class="normative">
+        <h3>Context Validation</h3>
+
+        <p>
+This section contains an algorithm that applications MUST run after running the
+algorithm in Section [[[#verify-proof]]] or Section
+[[[#verify-proof-sets-and-chains]]], when validating a [=conforming secured
+document=]. This algorithm takes a document ([=map=] |inputDocument|) and a set
```
iherman: What is the reason for this order? Instinctively, my choice would be to first check whether the document is all right in terms of JSON-LD, ie, run the context validation, _before_ running it through the validating steps (that may also involve running RDFC on the RDF graphs generated by the JSON-LD processing).

(Note if the order is reversed, it also has to be updated in the new processing model section.)

TallTed:
There's a lot of repeated text, here and in https://github.com/w3c/vc-data-model/pull/1535, which needs similar if not identical changes (which suggestions from myself and others do not currently reflect), possibly being adjusted to be *said* in one place, and *referenced* in the other.

msporny:
@TallTed wrote:
> There's a lot of repeated text, here and in [w3c/vc-data-model#1535](https://github.com/w3c/vc-data-model/pull/1535), which needs similar if not identical changes (which suggestions from myself and others do not currently reflect), possibly being adjusted to be _said_ in one place, and _referenced_ in the other.

See the reasoning here: https://github.com/w3c/vc-data-model/pull/1535#issuecomment-2248905296

index.html
msporny: No, errors will result in an `undefined` value for `document`, which is noted in the algorithm below.

index.html
msporny: Fixed in 3107ccb1a830ae90efa417df7ccf4d735021bcb6 (without creating more sections, just moved some text around).

index.html
msporny: Done in 0f234e8cd98edd2a245669653ab6bf0d20f09e88 ... which drastically changed the algorithm given @TallTed's changes. The previous algorithm didn't work for multiply nested documents w/ multiple contexts. The update attempts to address that use case. Please check my work @dlongley @davidlehn... I think I got all the corner cases, but need more eyes on it.

index.html
msporny: Fixed in 3107ccb1a830ae90efa417df7ccf4d735021bcb6.

iherman (approved):

index.html
```diff
@@ -1836,6 +1836,62 @@ <h2>Algorithms</h2>
...
+        <ol>
+          <li>
+Arranges the information into a document, such as a JSON or JSON-LD document.
+          </li>
+          <li>
+If the document is a JSON-LD document, selects one or more JSON-LD Contexts
+and expresses them using the `@context` property.
+          </li>
+          <li>
+Selects a cryptography suite that meets the needs of the use case, such as
```
```suggestion
Selects one or more cryptography suites that meet the needs of the use case, such as
```

index.html
```diff
@@ -2175,6 +2231,130 @@ <h3>Verify Proof Sets and Chains</h3>
...
+        <ul>
+          <li>
+a status ([=boolean=] |status|)
+          </li>
+          <li>
+a validated document ([=map=] |document|) if no errors occurred
+          </li>
+          <li>
+zero or more warnings ([=list=] of [=ProblemDetails=] |warnings|)
+          </li>
```
dlongley, line 2323: This isn't used in the algorithm.

```suggestion
```

index.html
```diff
@@ -2175,6 +2231,130 @@ <h3>Verify Proof Sets and Chains</h3>
...
+of input documents that it will process. Not checking JSON-LD context values
+against known good values can lead to security vulnerabilities, due to variance
+in the semantics that they convey. The context validation algorithm is one
+that applications validating a [=conforming secured document=] MUST run after
+running the algorithm in Section [[[#verify-proof]]] or Section
+[[[#verify-proof-sets-and-chains]]]. This algorithm takes a document ([=map=]
+|inputDocument|), a set of approved JSON-LD Contexts ([=map=]
+|approvedContexts|), and a boolean to recompact when unknown contexts are
+detected ([=boolean=] |recompact|) and returns a [=map=] that contains the
+following:
```
dlongley: I couldn't understand if the MUST meant that this particular algorithm MUST be run by any application that performs context validation (which I don't think should be the case), or if it meant that "this particular algorithm (if run), MUST be run after these other algorithms" (which makes sense to me).

```suggestion
in the semantics that they convey. This context validation algorithm is one that
applications MAY use to validate contexts in a [=conforming secured document=].
If used, it MUST be run after running the algorithm in Section [[[#verify-proof]]]
or Section [[[#verify-proof-sets-and-chains]]]. This algorithm takes a document
([=map=] |inputDocument|), a set of approved JSON-LD Contexts ([=map=]
|approvedContexts|), and a boolean to recompact when unknown contexts are
detected ([=boolean=] |recompact|) and returns a [=map=] that contains the
following:
```

index.html
```diff
@@ -2175,6 +2231,130 @@ <h3>Verify Proof Sets and Chains</h3>
...
+
+        <p>
+The context validation algorithm is as follows:
+        </p>
+
+        <ol class="algorithm">
+          <li>
+Set |result|.|status| to `false`, |result|.|warnings| to an empty list,
+|result|.|errors| to an empty list, |compactionContext| to an empty list,
+and clone |inputDocument| to |result|.|document|.
```
```suggestion
Set |result|.|status| to `false`, |result|.|errors| to an empty list,
|compactionContext| to an empty list, and clone |inputDocument|
to |result|.|document|.
```

index.html
```diff
@@ -2175,6 +2231,130 @@ <h3>Verify Proof Sets and Chains</h3>
...
+            </ol>
+          </li>
+          <li>
+If |result|.|errors| is empty, set |result|.|status| to `true`. Otherwise, set
+|result|.|status| to `false`, and remove the |document| property from |result|.
+          </li>
+          <li>
+Return the value of |result|.
+          </li>
+        </ol>
```
dlongley: I think this algorithm is more complex than it needs to be, is unlikely to be used in very complex use cases anyway as JSON schema checks + compaction will be used more easily leveraged for those, and it won't work properly anyway because of the changing semantics in each nested body (i.e., the "approvedContexts" would need to match the nesting too, which is much better served by and understood through the use of JSON schema).

I also think that this algorithm (when `recompact=false`) can result in a document that cannot be consumed because it contained no approved contexts, but somehow the cryptographic hashes matched known good values (this could only happen with a URL scheme that would allow hashes as a component of the context URL). But this sort of document still cannot be consumed by the application because the contexts aren't approved.

I recommend simplifying this algorithm for the less complex cases that expect a particular top-level context in the document and make `approvedContexts` simpler too: it can just become become an expected context value (a single context value or a list of context values in a particular order) as `expectedContext`. If the document does not have exactly the `expectedContext`, it should be recompacted to `expectedContext` or rejected (depending on the `recompact` flag).

I may be missing one of the aims here, but this approach seems far simpler to me:

```suggestion
          <li>
Let |contextValue| be the value of the `@context` property of |result|.|document|,
which might be undefined.
          </li>
          <li>
If |contextValue| does not deeply equal |expectedContext|, if any
any subtree in |result|.|document| contains an `@context` property,
or if any remote context in |contextValue|, which refers
to a JSON-LD Context file, does not match a known good value or
cryptographic hash:
            <ol class="algorithm">
              <li>
If |recompact| is `true` set |result|.|document| to the result
of running the <a data-cite="JSON-LD11-API#compaction-algorithm">
JSON-LD Compaction Algorithm</a> with the |inputDocument| and
|expectedContext| as inputs. If the compaction fails, add at least one error
to |result|.|errors|.
              </li>
              <li>
Otherwise, add at least one error to |result|.|errors|.
              </li>
            </ol>
          </li>
          <li>
If |result|.|errors| is empty, set |result|.|status| to `true`. Otherwise, set
|result|.|status| to `false`, and remove the |document| property from |result|.
          </li>
          <li>
Return the value of |result|.
          </li>
        </ol>
```

dlongley (changes requested):

index.html
dlongley: The above could possibly be tweaked for readability a bit more, but wanted to get this suggestion out for review.

index.html
TallTed: Building on @dlongley's suggestion
```suggestion
in the semantics that they convey. Applications MAY use this context validation
algorithm to validate contexts in a [=conforming secured document=]. If used,
it MUST be run <em>after</em> running the applicable algorithm in either
Section [[[#verify-proof]]] or Section [[[#verify-proof-sets-and-chains]]].
This algorithm takes inputs of a document ([=map=] |inputDocument|), a set of
approved JSON-LD Contexts ([=map=] |approvedContexts|), and a boolean to
recompact when unknown contexts are detected ([=boolean=] |recompact|), and
returns a [=map=] that contains the following:
```

index.html
```diff
@@ -1836,6 +1836,62 @@ <h2>Algorithms</h2>
 measures that could contribute to a safer and more efficient ecosystem.
       </p>
 
+      <section class="normative">
+        <h3>Processing Model</h3>
+
+        <p>
+The processing model used by a [=conforming processor=] and its
+application-specific software is described in this section. When software
+needs to ensure information is tamper-evident, it:
```
```suggestion
is to ensure information is tamper-evident, it performs the following steps:
```

index.html
```diff
@@ -1836,6 +1836,62 @@ <h2>Algorithms</h2>
...
+        </p>
+
+        <ol>
+          <li>
+Arranges the information into a document, such as a JSON or JSON-LD document.
```
```suggestion
The software arranges the information into a document, such as a JSON or
JSON-LD document.
```

index.html
```diff
@@ -1836,6 +1836,62 @@ <h2>Algorithms</h2>
...
+          </li>
+          <li>
+If the document is a JSON-LD document, selects one or more JSON-LD Contexts
+and expresses them using the `@context` property.
```
```suggestion
If the document is a JSON-LD document, the software selects one or more
JSON-LD Contexts and expresses them using the `@context` property.
```

index.html
```diff
@@ -1836,6 +1836,62 @@ <h2>Algorithms</h2>
...
+          </li>
+          <li>
+Selects a cryptography suite that meets the needs of the use case, such as
+one that uses acceptable cryptographic key material and provides full
+disclosure, selective disclosure, or unlinkable disclosure functionality.
```
TallTed: Including [suggestion](https://github.com/w3c/vc-data-integrity/pull/291/files#r1691623442) from @dlongley 
```suggestion
The software selects one or more cryptography suites that meet the needs of
the use case, such as one that provides full, selective, or unlinkable
disclosure, using acceptable cryptographic key material.
```

index.html
```diff
@@ -1836,6 +1836,62 @@ <h2>Algorithms</h2>
...
+          </li>
+          <li>
+Adds one or more proofs using the algorithms in Section [[[#add-proof]]] or
+Section [[[#add-proof-set-chain]]].
```
```suggestion
The software uses the applicable algorithms in either Section [[[#add-proof]]]
or Section [[[#add-proof-set-chain]]] to add one or more proofs.
```

index.html
```diff
@@ -1836,6 +1836,62 @@ <h2>Algorithms</h2>
...
+          </li>
+        </ol>
+
+        <p>
+When software needs to use information that was transmitted to it using
+a mechanism described by this specification, it:
```
```suggestion
a mechanism described by this specification, it performs the following steps:
```

index.html
```diff
@@ -1836,6 +1836,62 @@ <h2>Algorithms</h2>
...
+        </p>
+
+        <ol>
+          <li>
+Transforms the incoming data into a document that can be understood by the
+algorithms provided in Section [[[#verify-proof]]] or Section
+[[[#verify-proof-sets-and-chains]]].
```
```suggestion
The software transforms the incoming data into a document that can be understood
by the applicable algorithm provided in Section [[[#verify-proof]]] or Section
[[[#verify-proof-sets-and-chains]]].
```

index.html
```diff
@@ -1836,6 +1836,62 @@ <h2>Algorithms</h2>
...
+          </li>
+          <li>
+Validates that the incoming document follows an expected schema used by
+the application via JSON Schema or equivalent mechanism.
```
```suggestion
The software uses JSON Schema or equivalent mechanism to validate that the
incoming document follows an expected schema used by the application.
```

index.html
```diff
@@ -1836,6 +1836,62 @@ <h2>Algorithms</h2>
...
+          </li>
+          <li>
+Verifies the integrity of the incoming document using the algorithms provided in
+Section [[[#verify-proof]]] or Section [[[#verify-proof-sets-and-chains]]].
```
```suggestion
The software uses the applicable algorithms provided in Section [[[#verify-proof]]]
or Section [[[#verify-proof-sets-and-chains]]] to verify the integrity of the
incoming document.
```

index.html
```diff
@@ -1836,6 +1836,62 @@ <h2>Algorithms</h2>
...
+          </li>
+          <li>
+If the document is a JSON-LD document, validates all JSON-LD Context
+values used in the document using the algorithm provided in Section
+[[[#context-validation]]].
```
```suggestion
If the document is a JSON-LD document, the software uses the algorithm provided
in Section [[[#context-validation]]] to validate all JSON-LD Context values used
in the document.
```

index.html
```diff
@@ -2175,6 +2231,130 @@ <h3>Verify Proof Sets and Chains</h3>
...
 
+      <section class="normative">
+        <h3>Context Validation</h3>
+
+        <p>
+The purpose of the algorithm in this section is to ensure that a consuming
+application has explicitly approved of the types, and therefore the semantics,
+of input documents that it will process. Not checking JSON-LD context values
+against known good values can lead to security vulnerabilities, due to variance
+in the semantics that they convey. The context validation algorithm is one
```
```suggestion
The algorithm in this section is used to ensure that a consuming application has
explicitly approved of the types, and therefore the semantics, of input documents
that it will process. Using JSON-LD context values without checking them against
known good values can lead to security vulnerabilities, due to variance in the
semantics that they convey. The context validation algorithm is one
```

index.html
TallTed: Including suggestion from @dlongley... This, or the following...
```suggestion
Set |result|.|status| to `false`, |result|.|warnings| to an empty list,
|result|.|errors| to an empty list, |compactionContext| to an empty list,
and |inputDocument| to a clone of |result|.|document|.
```

TallTed: Including suggestion from @dlongley... This, or the above...
```suggestion
Set |result|.|status| to `false`, |result|.|warnings| to an empty list,
|result|.|errors| to an empty list, |compactionContext| to an empty list;
and clone |inputDocument| to |result|.|document|.
```

index.html
```diff
@@ -2175,6 +2231,130 @@ <h3>Verify Proof Sets and Chains</h3>
...
+        </p>
+
+        <ol class="algorithm">
+          <li>
+Set |result|.|status| to `false`, |result|.|warnings| to an empty list,
+|result|.|errors| to an empty list, |compactionContext| to an empty list,
+and clone |inputDocument| to |result|.|document|.
+          </li>
+          <li>
+For each subtree in |result|.|document| that contains a `@context` property:
```
```suggestion
For each subtree in |result|.|document| that contains a `@context` property,
perform the following actions:
```

index.html
```diff
@@ -2175,6 +2231,130 @@ <h3>Verify Proof Sets and Chains</h3>
...
+            <ol class="algorithm">
+              <li>
+For each value (|contextValue|) associated with the property:
```
```suggestion
For each value (|contextValue|) associated with the property,
perform the following actions:
```

index.html
```diff
@@ -2175,6 +2231,130 @@ <h3>Verify Proof Sets and Chains</h3>
...
+                <ol class="algorithm">
+                  <li>
+If |recompact| is `true` and the |contextValue| exists in |approvedContexts|,
+add the |contextValue| to the |compactionContext|.
+                  </li>
+                  <li>
+Otherwise, ensure that the contents of the |contextValue|, which is a JSON-LD
+Context file, match a known good value or cryptographic hash. If the check
+fails, add at least one error to |result|.|errors|.
```
```suggestion
Otherwise (i.e., if |recompact| is not `true` or the |contextValue| does not
exist in |approvedContexts|), ensure that the contents of the |contextValue|,
which is a JSON-LD Context file, match a known good file or cryptographic
hash. If the check fails, add at least one error to |result|.|errors|.
```

index.html
```diff
@@ -2175,6 +2231,130 @@ <h3>Verify Proof Sets and Chains</h3>
...
+of running the <a data-cite="JSON-LD11-API#compaction-algorithm">
+JSON-LD Compaction Algorithm</a> with the |inputDocument| and
+|compactionContext| as inputs. If the compaction fails, add at least one error
+to |result|.|errors|.
+              </li>
+            </ol>
+          </li>
+          <li>
+If |result|.|errors| is empty, set |result|.|status| to `true`. Otherwise, set
+|result|.|status| to `false`, and remove the |document| property from |result|.
```
```suggestion
If |result|.|errors| is empty, set |result|.|status| to `true`; otherwise, set
|result|.|status| to `false`, and remove the |document| property from |result|.
```

index.html
TallTed: Starting from suggestion by @dlongley... I am leaving "deeply equal" in place, but I think this comparator I've never seen before needs adjustment — could be a link to a definition of this concept, addition of such a definition here, or a different comparator that delivers the intended result.
```suggestion
          <li>
Let |contextValue| be the value of the `@context` property of |result|.|document|,
which might be undefined.
          </li>
          <li>
If |contextValue| does not deeply equal |expectedContext|, then if any subtree in
|result|.|document| contains an `@context` property or any URI in
|contextValue| dereferences to a JSON-LD Context file that does not match a
known good value or cryptographic hash, then perform the applicable action:
            <ol class="algorithm">
              <li>
If |recompact| is `true`, set |result|.|document| to the result
of running the <a data-cite="JSON-LD11-API#compaction-algorithm">
JSON-LD Compaction Algorithm</a> with the |inputDocument| and
|expectedContext| as inputs. If the compaction fails, add at least one error
to |result|.|errors|.
              </li>
              <li>
If |recompact| is not `true`, add at least one error to |result|.|errors|.
              </li>
            </ol>
          </li>
          <li>
If |result|.|errors| is empty, set |result|.|status| to `true`; otherwise, set
|result|.|status| to `false`, and remove the |document| property from |result|.
          </li>
          <li>
Return the value of |result|.
          </li>
        </ol>
```

index.html
```diff
@@ -2175,6 +2231,130 @@ <h3>Verify Proof Sets and Chains</h3>
...
+For example, if no JSON-LD processing is to occur, then, rather than performing
+this check, an application could follow the guidance in whatever trusted
+documentation is provided out of band for properly understanding the semantics
+of that type of document.
+        </p>
+
+        <p>
+Another approach would be to configure an application to use a JSON-LD Context
+loader, sometimes referred to as a document loader, to use only local copies of
+approved context files. This would guarantee that neither context files nor
```
```suggestion
approved context files. This would guarantee that neither the context files nor
```

index.html
```diff
@@ -2175,6 +2231,130 @@ <h3>Verify Proof Sets and Chains</h3>
...
+their cryptographic hashes would ever change, effectively resulting in the
+same result as the algorithm above.
+        </p>
+
+        <p>
+An alternative approach, also effectively equivalent to the algorithm provided in
+this section, would be for an application to keep a list of well known context
+URLs and their associated approved cryptographic hashes, without storing every
+context file locally. This would allow contexts to be safely loaded from the
+network without compromising the security expectations of the application.
```
```suggestion
Another alternative approach, also effectively equivalent to the algorithm provided
in this section, would be for an application to keep a list of well known context
URLs and their associated approved cryptographic hashes, without storing every
context file locally. This would allow these contexts to be safely loaded from
the network without compromising the security expectations of the application.
```

index.html
```diff
@@ -2175,6 +2231,130 @@ <h3>Verify Proof Sets and Chains</h3>
...
+        </p>
+
+        <p>
+Yet another valid approach would be for a transmitting application to
+<a data-cite="JSON-LD11-API#compaction-algorithm">compact</a> a document to
+exactly what a receiving application requests, omitting additional
```
TallTed: How does a receiving application request (a) specific context(s)?

index.html
```diff
@@ -2175,6 +2231,130 @@ <h3>Verify Proof Sets and Chains</h3>
...
+sender-specific context values that were used when securing the original document.
+As long as the cryptography suite's verification algorithm provides a successful
+verification result, such transformations are valid and would result in full
+URLs for terms that were previously compacted by the omitted context. That is, a
+term that was previously compacted to `foo` based on a sender-supplied context
+that is unknown to a receiver would instead be "compacted" to something like
+`https://vocabulary.example#foo` once the unknown context is omitted and the
+compaction algorithm is applied by the receiving application.
```
```suggestion
term that was previously compacted to `foo` based on a sender-supplied context
that is unknown to a receiver (e.g., ``https://ontology.example`) would instead
be "expanded" to a URL like `https://ontology.example#foo`, which would then
be "compacted" to the same URL, once the unknown context is omitted and the
JSON-LD compaction algorithm is applied by the receiving application.
```

TallTed (changes requested):

index.html
TallTed: @msporny — https://github.com/w3c/vc-data-integrity/commit/3107ccb1a830ae90efa417df7ccf4d735021bcb6 is a good start. I think lines 2312 through 2353 (using the line numbers visible at that commit) should also be moved to ~2245 (preceding the list of algorithm inputs) or possibly ~2264 (following the list of algorithm outputs). After that move, some further rephrasing may be appropriate. (Also note that there are change suggestions in flight on these sections, which I think should be made before the relocation, so the line numbers referenced here may also change before the relocation happens.)

index.html
msporny: Yes, but it's left in there so that implementations can add warnings if they would like (and there is a standards interface for them doing so. I'll add some text to note that implementations MAY add additional warnings/errors.

index.html
msporny: Fixed by explaining that it can happen via the query protocol.

index.html
msporny: Fixed in 04584e7b095ab6afbbaa83a6221f01db5f64ef2c.

msporny:

index.html
```diff
@@ -2175,6 +2297,84 @@ <h3>Verify Proof Sets and Chains</h3>
...
+Set |result|.|status| to `false`, |result|.|warnings| to an empty list,
+|result|.|errors| to an empty list, |compactionContext| to an empty list;
+and clone |inputDocument| to |result|.|document|.
+          </li>
+          <li>
+Let |contextValue| be the value of the `@context` property of |result|.|document|,
+which might be undefined.
+          </li>
+          <li>
+If |contextValue| does not deeply equal |expectedContext|, then if any subtree in
```
dlongley, line 2344: Slight mixup here in the grammar fixes, this step should read "or if" not "then if".

Issue #287, opened 2024-07-09
Error in the definition of the "unsecured document"


iherman:
(This issue comes from w3c/vc-data-model#1519, more exactly https://github.com/w3c/vc-data-model/pull/1519#discussion_r1667664777).

The term `unsecuredDocument` is defined in [§4 Algorithms](https://www.w3.org/TR/vc-data-integrity/#dfn-unsecured-data-document) of the DI specification as follows:

> An unsecured data document is a [map](https://infra.spec.whatwg.org/#ordered-map) with no proof [key](https://infra.spec.whatwg.org/#map-key).

This definition goes wrong for the case when proofs are chained (and if the intention is to prove the previous `proof` entry of the chain). See again https://github.com/w3c/vc-data-model/pull/1519#discussion_r1667664777 for more details on the effects on cryptosuites.

Also, it is not immediately obvious what happens in the case of a Verifiable Presentation: in a VP, on the top level there is no proof key in the data, because all other instantiations of the `proof` key (or equivalent) appear only "behind" the `verifiableCredential `key, so to say. Which does mean that the `unsecuredDocument` does in fact _include_ the proofs for the VCs in the Presentation, but it may be worth spelling this out.

I would also note that there is a duplication of definitions between this and the definitions in [§5.12 Securing Mechanism Specifications](https://www.w3.org/TR/vc-data-model-2.0/#securing-mechanism-specifications) (whether w3c/vc-data-model#1519 gets merged or not). This is a possible source of errors. I would propose we choose one of the sections for the formal definition, clean it up, and then refer to it from the other place.

cc @dlongley @msporny @TallTed

iherman:
Note, on the choice of where we want to have the final definition:

- §5.12 contains normative text on the definition of securing mechanisms in general
- At the moment, ecdsa and eddsa (I have not checked bbs) refer to the concept of `unsecuredDocument` (defined in §4) in their specifications on _what_ to secure

This means that

1. If we define the normative behavior v.a.v. chained proofs in §5.12, that will be valid _for all cryptosuites_. 
2. At the moment, cryptosuites _may_ decide to ignore the concept of `unsecureDocument` in their specification, i.e., not to secure chained proofs. 

I.e, if we think we want to keep it open for cryptosuites (i.e., keep (2)) then the formal definition should not be in §5.12 but, rather, in §4, and a remark should be added in §5.12 making it clear that cryptosuites may choose something else than what is in §4.

[...]

PR #292, opened 2024-07-21
Clarify meaning of unsecured/secured document.

msporny:
This PR is an attempt to address issue #287 by clarifying the meaning of unsecured/secured document..

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/292.html" title="Last updated on Jul 27, 2024, 6:46 PM UTC (c07caa3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/292/0eb5396...c07caa3.html" title="Last updated on Jul 27, 2024, 6:46 PM UTC (c07caa3)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1887,7 +1887,7 @@ The algorithms defined below operate on documents represented as <dfn
 data-cite="RFC8259#section-4">JSON objects</dfn>. This specification follows the
 [[[JSON-LD11-API]]] specification in representing a JSON object as an
 [[[Infra]]] [=map=]. An <dfn class="export">unsecured data document</dfn> is a
-[[[Infra]]] [=map=]. In this scenario, an <dfn class="export">unsecured data document</dfn> is a [=map=] that has not yet had the current proof added to it, but it MAY contain a
+[[[Infra]]] [=map=] that has not yet had the current proof added to it, but it MAY contain a
 proof value added to it by a previous process. A <dfn class="export">secured
 data document</dfn> is a [=map=] that has had the current proof(s) added to it.
       </p>
--- a/index.html
+++ b/index.html
@@ -509,8 +509,8 @@ A <dfn>conforming secured document</dfn> is any [=byte sequence=] that can be
 converted to a
 <a data-cite="INFRA#parse-json-bytes-to-a-javascript-value">
 JSON document</a> that follows the relevant normative requirements in
-Sections [[[#proofs]]], [[[#proof-purposes]]], [[[#contexts-and-vocabularies]]],
-and [[[#dataintegrityproof]]].
+Sections [[[#proofs]]], [[[#proof-purposes]]], [[[#resource-integrity]]],
+[[[#contexts-and-vocabularies]]], and [[[#dataintegrityproof]]].
         </p>
 
         <p>
@@ -1040,7 +1040,7 @@ document.
         </p>
 
         <p>
-The information contained in an unsecured document before a
+The information contained in an input document before a
 [=data integrity proof=] is added to the document is expressed in one or more
 graphs. To ensure that information from different [=data integrity proofs=] is not
 accidentally co-mingled, the concept of a [=proof graph=] is used to encapsulate
@@ -1744,8 +1744,8 @@ cryptographic suite instance</dfn> [=struct=] has the following
           <dl data-dfn-for="cryptosuite instance">
             <dt><dfn>createProof</dfn></dt>
             <dd>
-An algorithm that takes an [=unsecured data document=] ([=map=]
-|unsecuredDocument|) and proof options ([=map=] |options|) as input, and
+An algorithm that takes an [=input document=] ([=map=]
+|inputDocument|) and proof options ([=map=] |options|) as input, and
 produces a [=data integrity proof=] ([=map=]) or an error.
             </dd>
             <dt><dfn>verifyProof</dfn></dt>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1820,11 +1820,12 @@ <h2>Algorithms</h2>
...
-[[JSON-LD-API]] in representing a JSON object as an [[Infra]] [=map=]. An
-<dfn class="export">unsecured data document</dfn> is a [=map=] with no
-`proof` [=map/key=]. A <dfn class="export">secured data document</dfn> is a
-[=map=] with a `proof` [=map/key=].
+data-cite="RFC8259#section-4">JSON objects</dfn>. This specification follows the
+[[[JSON-LD11-API]]] specification in representing a JSON object as an
+[[[Infra]]] [=map=]. An <dfn class="export">unsecured data document</dfn> is a
+[=map=] that has not yet had the desired proof added to it, but it MAY contain a
+proof value added to it by a previous process. A <dfn class="export">secured
+data document</dfn> is a [=map=] that has had the desired proof(s) added to it.
```
dlongley: Instead of "desired" ... since maybe the document already has some desirable proof in it ... perhaps "impending" to identify the proof that is about to be added? Perhaps "target proof" could be used if "impending" is not preferred.

```suggestion
[=map=] that has not yet had the impending proof added to it, but it MAY contain a
proof value added to it by a previous process. A <dfn class="export">secured
data document</dfn> is a [=map=] that has had the impending proof(s) added to it.
```

dlongley: Alternative "target":

```suggestion
[=map=] that has not yet had the target proof added to it, but it MAY contain a
proof value added to it by a previous process. A <dfn class="export">secured
data document</dfn> is a [=map=] that has had the target proof(s) added to it.
```

Wind4Greg (approved):

thereisnogabe (approved):

index.html
TallTed: To me, "target" implies "thing to which other things are being applied" rather than "thing being applied to another thing". So, I think, not `target`.

TallTed: `impending` is good for the first instance, but not for the second, because `impending` suggests an action is yet to be taken, while the second instance is after that action. Perhaps (note that this would also replace line 1825, which really wants to say `In this context` instead of `In this scenario`, but that's guaranteed to confuse our readers) —
```suggestion
[[[Infra]]] [=map=]. In this scenario, an <dfn class="export">unsecured data document</dfn> is a [=map=] that has not yet had the current proof added to it, but it MAY contain a
proof value added to it by a previous process. A <dfn class="export">secured
data document</dfn> is a [=map=] that has had the current proof(s) added to it.
```

TallTed (changes requested):

iherman:
Looking back at the previous comments, my vote would be for @TallTed's  formulation in https://github.com/w3c/vc-data-integrity/pull/292#discussion_r1686902419.

iherman (approved):
I just want to be picky so that we are sure what we want to say…

The section stands by itself within the context (sic!) of the DI specification. This means it defines everything based on the `proof` property and its usage. However, the VCDM specification refers to DI as one possible [embedded securing mechanism](https://www.w3.org/TR/vc-data-model-2.0/#securing-mechanisms); there may be others. I.e., if someone comes with a DI-bis specification with a securing mechanism based on the usage of a, say, `verify` property (which could be used side-by-side with the `proof` property) nothing is said on how the two would work together.

We *could* describe the whole algorithm, and the "secured document" term, in a slightly more general way, referring to _any_ embedded proof mechanism. I am not saying we should (it is probably too late for that anyway) but I want to be sure that we are all aware of this fact.

index.html
```diff
@@ -1820,11 +1820,12 @@ <h2>Algorithms</h2>
...
-[[JSON-LD-API]] in representing a JSON object as an [[Infra]] [=map=]. An
-<dfn class="export">unsecured data document</dfn> is a [=map=] with no
-`proof` [=map/key=]. A <dfn class="export">secured data document</dfn> is a
-[=map=] with a `proof` [=map/key=].
+data-cite="RFC8259#section-4">JSON objects</dfn>. This specification follows the
+[[[JSON-LD11-API]]] specification in representing a JSON object as an
+[[[Infra]]] [=map=]. An <dfn class="export">unsecured data document</dfn> is a
+[[[Infra]]] [=map=]. In this scenario, an <dfn class="export">unsecured data document</dfn> is a [=map=] that has not yet had the current proof added to it, but it MAY contain a
+proof value added to it by a previous process. A <dfn class="export">secured
+data document</dfn> is a [=map=] that has had the current proof(s) added to it.
```
```suggestion
data-cite="RFC8259#section-4">JSON objects</dfn>. This specification follows the
[[[JSON-LD11-API]]] specification in representing a JSON object as an
[[[Infra]]] [=map=]. An <dfn class="export">unsecured data document</dfn> is a
[[[Infra]]] [=map=] that has not yet had the current proof added to it, but it MAY contain a
proof value added to it by a previous process. A <dfn class="export">secured
data document</dfn> is a [=map=] that has had the current proof(s) added to it.
```

dlongley (approved):

index.html
```diff
@@ -1820,11 +1820,12 @@ <h2>Algorithms</h2>
...
-[[JSON-LD-API]] in representing a JSON object as an [[Infra]] [=map=]. An
-<dfn class="export">unsecured data document</dfn> is a [=map=] with no
-`proof` [=map/key=]. A <dfn class="export">secured data document</dfn> is a
-[=map=] with a `proof` [=map/key=].
+data-cite="RFC8259#section-4">JSON objects</dfn>. This specification follows the
+[[[JSON-LD11-API]]] specification in representing a JSON object as an
+[[[Infra]]] [=map=]. An <dfn class="export">unsecured data document</dfn> is a
+[[[Infra]]] [=map=] that has not yet had the current proof added to it, but it MAY contain a
+proof value added to it by a previous process. A <dfn class="export">secured
+data document</dfn> is a [=map=] that has had the current proof(s) added to it.
```
```suggestion
[[[Infra]]] [=map=]. An <dfn class="export">unsecured data document</dfn> is an
[[[Infra]]] [=map=] that has not yet had the current proof added to it, but it
MAY contain a proof value that was added to it by a previous process. A <dfn
class="export">secured data document</dfn> is a [=map=] that has had the current
proof(s) added to it.
```

index.html
msporny: I've updated the spec to speak to "unsecured data documents", "input documents", and "secured data documents". The new formulation feels like it addresses various concerns raised throughout this PR.

msporny:
> I am not saying we should (it is probably too late for that anyway) but I want to be sure that we are all aware of this fact.

Yes, it's a good point, though without an alternative I hesitate to try and create language that would be inclusive of that alternative. I did make a pass to try to figure out how else to say it, and it's difficult to say it w/o also pointing to the alternative.

Ultimately, this language is in the Algorithms section and that applies directly to DI and its usage of `proof`. I'm not sure we need to be concerned w/ other specifications and their "proof" values, because DI proofs have nothing to do with those other securing mechanisms (we don't have language on how to process those other things... we only know of our own properties and algorithms).

msporny:

Issue #281, opened 2024-07-05
Reconsider `@context` injection feature based on implementer feedback


msporny:
After performing an implementation, and working through a PR, @rflechtner has proposed that we remove `@context` injection feature from the Data Integrity specification in this comment:

https://github.com/w3c/vc-data-integrity/issues/231#issuecomment-2209299341

The feature was added based on feedback that "implementers should be able to inject a context". The current text doesn't quite do that; it's a bit of a half-measure, and after some discussion, there does not seem to be support to continue to support the injection feature in its current form (or at all).

This issue is meant to track that concern and decide what to do. Among the options to consider are:

1. Normatively, continue to support context injection, and provide much more clear rules on how to support it across all cryptosuites.
2. Normatively, do not support context injection at all.
2. Normatively, do not prohibit context injection, but informatively note that its up to applications to do it after they have carefully considered the ramifications.

jandrieu:
Can we get clarification on what "context injection" is?

rflechtner:
> Can we get clarification on what "context injection" is?

Context injection is described in https://www.w3.org/TR/vc-data-integrity/#context-injection and refers to amending the `@context` property of the secured document with a context value that maps the terms contained in a proof. In practice, this means that the document is altered in the process of proof creation - or even of proof verification, the specification is not entirely clear here. My assumption is that this practice was initially introduced under the assumption that the `@context` property is not secured by the proof (see https://github.com/w3c/vc-data-integrity/issues/272 where this has been discussed) and thus can be manipulated without negative consequences, an assumption that however does not hold true when using cryptosuites that treat the document as regular JSON, e.g. those using the JCS transformation (such as [eddsa-jcs-2022](https://www.w3.org/TR/vc-di-eddsa/#eddsa-jcs-2022)).

We've tried addressing the issue by means of a workaround on JCS cryptosuites (https://github.com/w3c/vc-data-integrity/issues/225), but as I outlined in https://github.com/w3c/vc-data-integrity/issues/231 there are more accommodations that would need to be made to make sure that JCS-based proofs can reliably be verified. These accommodations, if they are even possible, need to be very carefully designed to make sure that not only reliably verification is guaranteed, but also an effective protection of the semantics expressed in `@context` values originally added by the issuer.

[...]

PR #301, opened 2024-07-29
Make `@context` injection optional.

msporny:
This PR has been raised to address issue #281 by making context injection optional during the securing process. 

/cc @rflechtner @silverpill

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/301.html" title="Last updated on Aug 3, 2024, 9:21 PM UTC (17b1521)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/301/a065cfc...17b1521.html" title="Last updated on Aug 3, 2024, 9:21 PM UTC (17b1521)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1411,20 +1411,22 @@ as `DataIntegrityProof`, are used.
           </p>
 
           <p>
-When securing a document, if an `@context` property is not provided in the document
-or the Data Integrity terms used in the document are not mapped by
-existing values in the `@context` property, implementations MUST inject or add
-an `@context` property with a value of
-`https://w3id.org/security/data-integrity/v2`.
+When an application is securing a document, if an `@context` property is not
+provided in the document or the Data Integrity terms used in the document are
+not mapped by existing values in the `@context` property, implementations SHOULD
+inject or add a `@context` property with a value of
+`https://w3id.org/security/data-integrity/v2` or one or more contexts with at
+least the same declarations, such as the Verifiable Credential Data Model v2.0
+context (`https://www.w3.org/ns/credentials/v2`).
           </p>
 
           <p>
-Context injection is expected to be unnecessary sometimes, such as when the Verifiable
-Credential Data Model v2.0 context (`https://www.w3.org/ns/credentials/v2`)
-exists as a value in the `@context` property, as that context maps all of the
-necessary Data Integrity terms that were previously mapped by
-`https://w3id.org/security/data-integrity/v2`.
+Implementations that do no desire to use JSON-LD processing MAY choose to not
+include a `@context` declaration at the top-level of the document. If a
+`@context` declaration is not provided, extensions to this specification, or
+corresponding cryptosuites MUST NOT be made.
           </p>
+
         </section>
 
         <section>
--- a/index.html
+++ b/index.html
@@ -1421,11 +1421,11 @@ context (`https://www.w3.org/ns/credentials/v2`).
           </p>
 
           <p>
-Implementations that do not desire to use JSON-LD processing MAY choose to not
-include a `@context` declaration at the top-level of the document. If a
-`@context` declaration is not provided, extensions (such as the addition of new
-properties) related to this specification or corresponding cryptosuites MUST NOT
-be made.
+Implementations that do not intend to use JSON-LD processing MAY choose to not
+include an `@context` declaration at the top-level of the document. However, if
+an `@context` declaration is not included, extensions (such as the addition of
+new properties) related to this specification or corresponding cryptosuites
+MUST NOT be made.
           </p>
 
         </section>
--- a/index.html
+++ b/index.html
@@ -1414,17 +1414,18 @@ as `DataIntegrityProof`, are used.
 When an application is securing a document, if an `@context` property is not
 provided in the document or the Data Integrity terms used in the document are
 not mapped by existing values in the `@context` property, implementations SHOULD
-inject or add a `@context` property with a value of
+inject or append an `@context` property with a value of
 `https://w3id.org/security/data-integrity/v2` or one or more contexts with at
 least the same declarations, such as the Verifiable Credential Data Model v2.0
 context (`https://www.w3.org/ns/credentials/v2`).
           </p>
 
           <p>
-Implementations that do no desire to use JSON-LD processing MAY choose to not
+Implementations that do not desire to use JSON-LD processing MAY choose to not
 include a `@context` declaration at the top-level of the document. If a
-`@context` declaration is not provided, extensions to this specification, or
-corresponding cryptosuites MUST NOT be made.
+`@context` declaration is not provided, extensions (such as the addition of new
+properties) related to this specification or corresponding cryptosuites MUST NOT
+be made.
           </p>
 
         </section>


index.html
```diff
@@ -1411,20 +1411,23 @@ <h3>Context Injection</h3>
...
-Context injection is expected to be unnecessary sometimes, such as when the Verifiable
-Credential Data Model v2.0 context (`https://www.w3.org/ns/credentials/v2`)
-exists as a value in the `@context` property, as that context maps all of the
-necessary Data Integrity terms that were previously mapped by
-`https://w3id.org/security/data-integrity/v2`.
+Implementations that do not desire to use JSON-LD processing MAY choose to not
+include a `@context` declaration at the top-level of the document. If a
+`@context` declaration is not provided, extensions (such as the addition of new
+properties) related to this specification or corresponding cryptosuites MUST NOT
+be made.
```
```suggestion
Implementations that do not intend to use JSON-LD processing MAY choose to not
include an `@context` declaration at the top-level of the document. However, if
an `@context` declaration is not included, extensions (such as the addition of
new properties) related to this specification or corresponding cryptosuites
MUST NOT be made.
```

Wind4Greg (approved):

dlongley (approved):

silverpill (approved):

index.html
```diff
@@ -1411,20 +1411,23 @@ <h3>Context Injection</h3>
...
           </p>
 
           <p>
-Context injection is expected to be unnecessary sometimes, such as when the Verifiable
-Credential Data Model v2.0 context (`https://www.w3.org/ns/credentials/v2`)
-exists as a value in the `@context` property, as that context maps all of the
-necessary Data Integrity terms that were previously mapped by
-`https://w3id.org/security/data-integrity/v2`.
+Implementations that do not desire to use JSON-LD processing MAY choose to not
+include a `@context` declaration at the top-level of the document. If a
```
silverpill: Does this mean that step (2) of [Create Proof (eddsa-jcs-2022)](https://w3c.github.io/vc-di-eddsa/#create-proof-eddsa-jcs-2022) algorithm will be optional too?

>Set `proof.@context` to `unsecuredDocument.@context`.

dlongley: I don't think it's optional, it just results in a no-op because `unsecuredDocument.@context` would be undefined in the cases where it isn't present (and thus so would `proof.@context`, which is already acceptable in the verification algorithm). If we feel it's necessary, we could say: "If `unsecuredDocument.@context` is present, set `proof.@context` to `unsecuredDocument.@context`".

msporny: I'd be fine to add @dlongley's suggestion if @silverpill wants that clarification. It would be a simple editorial change (don't need to do it in this PR).

index.html
```diff
@@ -1411,20 +1411,23 @@ <h3>Context Injection</h3>
           </p>
 
           <p>
-When securing a document, if an `@context` property is not provided in the document
-or the Data Integrity terms used in the document are not mapped by
-existing values in the `@context` property, implementations MUST inject or add
-an `@context` property with a value of
-`https://w3id.org/security/data-integrity/v2`.
+When an application is securing a document, if an `@context` property is not
+provided in the document or the Data Integrity terms used in the document are
```
iherman, line 1415:
```
When an application is securing a document, if an `@context` property is not
provided in the document or the Data Integrity terms used in the document are
```

Excuse me if I am overly picky at this point...

Per the [Interpreting JSON as JSON-LD section](https://www.w3.org/TR/json-ld11/#interpreting-json-as-json-ld) of the JSON-LD spec, it is possible to transfer the context through an HTTP Link Header. This means that the term "provided in the document" may not be absolutely precise.

Interestingly, the aforementioned JSON-LD section also includes this statement:

> Please note that [JSON-LD documents](https://www.w3.org/TR/json-ld11/#dfn-json-ld-document) served with the application/ld+json media type MUST have all context information, including references to external contexts, within the body of the document. Contexts linked via a http://www.w3.org/ns/json-ld#context [HTTP Link Header](https://tools.ietf.org/html/rfc8288#section-3) MUST be ignored for such documents.

So far, this statement applied to VC documents in JSON-LD, so we were fine. However, the official media type is now `/vc`, meaning that the JSON-LD statement does not apply anymore.

(Actually, a possibility might be to take over the JSON-LD statement and add it to the media type specification...)

index.html
silverpill: >If `unsecuredDocument.@context` is present, set `proof.@context` to `unsecuredDocument.@context`

+1

I think such clarification would be helpful

msporny: @silverpill, applied your (editorial) request to all JCS cryptosuites:

* https://github.com/w3c/vc-di-ecdsa/commit/7c91cb9286d30d302adbec2cdbe16f065a44d197
* https://github.com/w3c/vc-di-eddsa/commit/c728ae3edeec57271a07b2484d2d91353215f9b8

index.html
msporny: There are two problems here:

1. That guidance in JSON-LD 1.1 (rel-linking to a context) seems to be something that is rarely, if ever used. It seems to be creating unnecessary complexity without much gain.
2. We don't use the rel-linking style anywhere in VCDM and DI, and should probably keep it that way (to avoid unnecessary complexity).

All that to say, we should probably fix JSON-LD v1.1 and not further complicate VCDM or DI with that rel-linking feature. I'm presuming your comment is a no-op on this PR, @iherman... if you think we should do something about it, it probably deserves an issue for discussion.

index.html
msporny: I did quite a bit of editorial clean up to the JCS Create Proof and Verify Proof algorithms, you might just want to look at the latest live Editor's drafts instead.

msporny:

index.html
iherman: > There are two problems here:
> 
> 1. That guidance in JSON-LD 1.1 (rel-linking to a context) seems to be something that is rarely, if ever used. It seems to be creating unnecessary complexity without much gain.

Maybe. But "hardly ever used" is a bit vague for the JSON-LD WG to act upon. At this point, if the JSON-LD WG proposed to remove the feature, at the minimum it should ask the community to see if there are users of the feature, ie, whether removing it would break applications. In the absence of this, I would be opposed to remove it.

> 2. We don't use the rel-linking style anywhere in VCDM and DI, and should probably keep it that way (to avoid unnecessary complexity).

I actually agree. But while that was automatically the case so far, it is not anymore 😒.

> 
> All that to say, we should probably fix JSON-LD v1.1 and not further complicate VCDM or DI with that rel-linking feature.

This is where we disagree. We have a problem _now_, and it is very uncertain that the JSON-LD WG will change that in the near future. I believe we _must_ document this restriction.

> I'm presuming your comment is a no-op on this PR, @iherman... if you think we should do something about it, it probably deserves an issue for discussion.

Yes, I agree it is a no-op on this PR. If we agree that the best place to handle this is to add a remark to the media type documentations (i.e., [Appendix C of the VCDM spec](https://www.w3.org/TR/vc-data-model-2.0/#iana-considerations)) then we should move ahead with a PR on the VCDM right away. If you prefer to do it, fine, if you want me to do it, I can try.

iherman: Actually, @msporny, there is a statement in the media type specification:

> A [[JSON-LD11](https://www.w3.org/TR/vc-data-model-2.0/#bib-json-ld11)] context is expected to be present in the body of the document. The credential is expected to be a valid [JSON-LD document](https://www.w3.org/TR/json-ld11/#dfn-json-ld-document).

Which may be interpreted as saying that the link is not allowed. I still believe making this explicit should be done.

msporny: Sure, making it explicit would be helpful. Do you want to raise an issue/PR to do that, @iherman? or should I?

iherman: > Sure, making it explicit would be helpful. Do you want to raise an issue/PR to do that, @iherman? or should I?

I will do it, probably tomorrow

iherman: See https://github.com/w3c/vc-data-model/pull/1539, @msporny

PR #320, opened 2024-11-16
Define ControllerDocument class.

msporny:
This PR is an attempt to address issue https://github.com/w3c/vc-data-integrity/issues/322 by defining a ControllerDocument class. 

/cc @filip26

### DIFF
--- a/vocab/security/template.html
+++ b/vocab/security/template.html
@@ -270,7 +270,9 @@
         </p>
         <p>
           Each of these sections has an ellipse at the top, labeled as "Proof", "VerificationMethod", 
-          and "VerificationRelationship", respectively.
+          and "VerificationRelationship", respectively. Additionally, the Verification Method section also
+          includes an additional ellipse, labeled as "ControllerDocument", side-by-side with the ellipse
+          labeled as "VerificationMethod".
         </p>
         <section>
           <h3>Proof Section</h3>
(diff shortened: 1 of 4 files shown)


dlongley (approved):

iherman:
The PR is incomplete, because the diagram must be updated as well and. It is also a question whether the domain of the "controller" property shouldn't be specified as a union (ie, logical 'or') for a controller document class and the verification method class.

@msporny I think I should take it over, because I have more practice in creating/changing the diagram. I will clone your branch and take it from there if you are fine with it.

msporny:
> It is also a question whether the domain of the "controller" property shouldn't be specified as a union (ie, logical 'or') for a controller document class and the verification method class.

I don't think we should limit the domain just yet... controller might make sense in many more use cases.

> @msporny I think I should take it over, because I have more practice in creating/changing the diagram. I will clone your branch and take it from there if you are fine with it.

Yes, please take it over @iherman ... and thank you! :)

iherman:
> > It is also a question whether the domain of the "controller" property shouldn't be specified as a union (ie, logical 'or') for a controller document class and the verification method class.
> 
> I don't think we should limit the domain just yet... controller might make sense in many more use cases.
> 
> > @msporny I think I should take it over, because I have more practice in creating/changing the diagram. I will clone your branch and take it from there if you are fine with it.
> 
> Yes, please take it over @iherman ... and thank you! :)

Done.

iherman (approved):

vocab/security/template.html
```diff
@@ -270,7 +270,9 @@ <h2>Diagram description</h2>
         </p>
         <p>
           Each of these sections has an ellipse at the top, labeled as "Proof", "VerificationMethod", 
-          and "VerificationRelationship", respectively.
+          and "VerificationRelationship", respectively. Additionally, the Verification Method section also
+          includes an additional ellipse, labeled as "ControllerDocument", side-by-side with the ellipse
+          labeled as "VerificationMethod".
```
TallTed:
```suggestion
          and "VerificationRelationship", respectively. The Verification Method section
          includes an additional ellipse, labeled as "ControllerDocument", side-by-side with the ellipse
          labeled as "VerificationMethod".
```

vocab/security/template.html
```diff
@@ -304,16 +306,17 @@ <h3>Verification Method Section</h3>
...
           <p>
             The left side of this section contains a column of three labeled boxes, all styled as "Property". 
-            The labels, from top to bottom, are "expires", "controller", and "revoked". 
+            The labels, from top to bottom, are "controller", "expires", and "revoked".
             Each of these is connected to the "VerificationMethod" ellipse, with a line styled as "Domain Of".
-            The "expires" "Property" box is also connected to the "Proof" ellipse (in the Proof section), 
-            with a line styled as "Domain Of".
+            The "controller" "Property" box is also connected to the "ControllerDocument" ellipse, while
+            the "expires" "Property" box is also connected to the "Proof" ellipse (in the Proof section);
+            both these extra connections are also with a line styled as "Domain Of".
```
TallTed:
```suggestion
            both of these extra connecting lines are styled as "Domain Of".
```

TallTed (changes requested):
The redundant department of redundancy department thanks you for your support.

msporny:

PR #327, opened 2025-01-06
Remove the numerical error codes to be in sync with the VCDM specification

iherman:
Following up on https://github.com/w3c/cid/pull/134#discussion_r1884295424: 

- Removed the error codes from the processing errors (both in the spec and in the vocabulary). I have also synched with the corresponding VCDM spec which _require_ some terms in the error.
- In the vocabulary, the links to the controller document was still present for the error code; I have changed that.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/327.html" title="Last updated on Jan 20, 2025, 7:11 AM UTC (13b9a61)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/327/5b24b45...13b9a61.html" title="Last updated on Jan 20, 2025, 7:11 AM UTC (13b9a61)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2366,7 +2366,7 @@ particular use case.
 The algorithms described in this specification, as well as in various cryptographic
 suite specifications, throw specific types of errors. Implementers might find
 it useful to convey these errors to other libraries or software systems. This
-section provides specific URLs, descriptions, and error codes for the errors,
+section provides specific URLs and descriptions for the errors,
 such that an ecosystem implementing technologies described by this
 specification might interoperate more effectively when errors occur.
         </p>
@@ -2384,38 +2384,34 @@ The `type` value of the error object MUST be a URL that starts with the value
 below.
           </li>
           <li>
-The `code` value MUST be the integer code described in the table below
-(in parentheses, beside the type name).
-          </li>
-          <li>
-The `title` value SHOULD provide a short but specific human-readable [=string=] for
+The `title` value MUST be present and SHOULD provide a short but specific human-readable [=string=] for
 the error.
           </li>
           <li>
-The `detail` value SHOULD provide a longer human-readable [=string=] for the error.
+The `detail` value MUST be present and SHOULD provide a longer human-readable [=string=] for the error.
           </li>
         </ul>
 
         <dl>
-          <dt id="PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR (-16)</dt>
+          <dt id="PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</dt>
           <dd>
 A request to generate a proof failed. See Section [[[#add-proof]]], and Section
 [[[#add-proof-set-chain]]].
           </dd>
-          <dt id="PROOF_VERIFICATION_ERROR">PROOF_VERIFICATION_ERROR (-17)</dt>
+          <dt id="PROOF_VERIFICATION_ERROR">PROOF_VERIFICATION_ERROR</dt>
           <dd>
 An error was encountered during proof verification. See Section [[[#verify-proof]]].
           </dd>
-          <dt id="PROOF_TRANSFORMATION_ERROR">PROOF_TRANSFORMATION_ERROR  (-18)</dt>
+          <dt id="PROOF_TRANSFORMATION_ERROR">PROOF_TRANSFORMATION_ERROR </dt>
           <dd>
 An error was encountered during the transformation process.
           </dd>
-          <dt id="INVALID_DOMAIN_ERROR">INVALID_DOMAIN_ERROR (-19)</dt>
+          <dt id="INVALID_DOMAIN_ERROR">INVALID_DOMAIN_ERROR</dt>
           <dd>
 The `domain` value in a proof did not match the expected value. See Section
 [[[#verify-proof]]].
           </dd>
-          <dt id="INVALID_CHALLENGE_ERROR">INVALID_CHALLENGE_ERROR (-20)</dt>
+          <dt id="INVALID_CHALLENGE_ERROR">INVALID_CHALLENGE_ERROR</dt>
           <dd>
 The `challenge` value in a proof did not match the expected value. See Section
 [[[#verify-proof]]].
--- a/index.html
+++ b/index.html
@@ -2388,7 +2388,7 @@ The `title` value SHOULD provide a short but specific human-readable [=string=] 
 the error.
           </li>
           <li>
-The `detail` value MUST be present and SHOULD provide a longer human-readable [=string=] for the error.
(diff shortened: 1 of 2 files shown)


Wind4Greg (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2025-01-08-vcwg#section3-1) on 2025-01-08

- no resolutions were taken

#### 3.1. Remove the numerical error codes to be in sync with the VCDM specification (pr cid#140)

_See github pull request [cid#140](https://github.com/w3c/cid/pull/140)._




**Brent Zundel:** Remove numerical error codes to be in sync with the VCDM spec.  

> *Manu Sporny:* +1 to this change.

**Brent Zundel:** It does what it says. It removes the error codes while retaining the names of the error codes. Anyone want to say we don't want to do this, please say so, can take comments briefly, etc. if needed.  

**Ivan Herman:** Just remarking that there is a sister PR in the DI spec which does the same. It also makes the changes in the vocab definition file. These two PRs should go hand-in-hand.  

_See github pull request [vc-data-integrity#327](https://github.com/w3c/vc-data-integrity/pull/327)._




**Manu Sporny:** Yes, +1 -- we made a decision in the group to remove the error codes in the group and this is just Ivan making sure we follow that guidance.  

</details>

index.html
```diff
@@ -2384,38 +2384,34 @@ <h3>Processing Errors</h3>
 below.
           </li>
           <li>
-The `code` value MUST be the integer code described in the table below
-(in parentheses, beside the type name).
-          </li>
-          <li>
-The `title` value SHOULD provide a short but specific human-readable [=string=] for
+The `title` value MUST be present and SHOULD provide a short but specific human-readable [=string=] for
```
dlongley: This should match what's in the other specs and not add a new MUST: https://w3c.github.io/vc-bitstring-status-list/#processing-errors

```suggestion
The `title` value SHOULD provide a short but specific human-readable [=string=] for
```

index.html
```diff
@@ -2384,38 +2384,34 @@ <h3>Processing Errors</h3>
...
 the error.
           </li>
           <li>
-The `detail` value SHOULD provide a longer human-readable [=string=] for the error.
+The `detail` value MUST be present and SHOULD provide a longer human-readable [=string=] for the error.
```
dlongley: This should match what's in the other specs and not add a new MUST: https://w3c.github.io/vc-bitstring-status-list/#processing-errors

```suggestion
The `detail` value SHOULD provide a longer human-readable [=string=] for the error.
```

dlongley (changes requested):
Minor but important clean up then +1. I see VCDM's problem details isn't in sync with bitstring status list either, looks like we should just loosen VCDM and then everything will be in sync and will also match the RFC requirements (advisory only), see: https://www.rfc-editor.org/rfc/rfc9457#name-title.

index.html
iherman: Well, actually... I have added the "MUST be present and" statement exactly to be consistent with the VCDM spec, see https://www.w3.org/TR/vc-data-model-2.0/#problem-details

I do not really care either way, we have to decide whether we align with the VCDM or we align the VCDM with this...

index.html
iherman: See my comment above

index.html
dlongley: Right, I think we should align the VCDM with this more flexible approach (no MUSTs) that is also closer to the RFC.

iherman: I am neutral on this. But this should be a decision of the WG, probably to be discussed on a call.

@brentzundel

msporny: Hmm, looks like we have a consistency mismatch between VCDM, DI, and BSL :(. VCDM says each property MUST be present (this language). DI and BSL say SHOULD. RFC9457 says SHOULD as well.

I think we were trying to make the error messages useful by using MUST for certain properties existing (`type`, `title`, and `detail`). I still think that's a good idea, that is -- if you are going to raise an error, you MUST include at least those properties for the error to be immediately useful for a developer).

Let's discuss during the next WG call.

index.html
msporny: See comment here: https://github.com/w3c/vc-data-integrity/pull/327/files#r1912460821

index.html
dlongley: > I think we were trying to make the error messages useful by using MUST for certain properties existing (type, title, and detail).

The `type` property is a MUST everywhere (in all the specs already) and should stay that way. It's only the properties that are advisory from the RFC (e.g., `title` and `detail`) that I believe should be SHOULDs. I think only `type` is really necessary for interoperability and is what software should be using to make any branching decisions. But, these other fields are useful for developers -- and the fact that they are will drive whether or not they are included, i.e., the SHOULD provides flexibility but systems will probably include them for the best DX without us having to demand it for interop, which doesn't make sense, IMO.

msporny: I'm fine w/ what @dlongley says above -- @iherman, if you're fine with that approach, can you please revert your change to MUSTs? 

If not, can we make the SHOULD/MUST changes into a separate PR. If this PR just removes the numerical error codes, that's non-controversial and we can get that in sooner than later.

iherman:
@msporny 

> I'm fine w/ what @dlongley says above -- @iherman, if you're fine with that approach, can you please revert your change to MUSTs?

I have reversed the SHOULD/MUST change; as I said, I am fine either way, and my only concern was consistency. I have also reversed that part of the PR in CID: https://github.com/w3c/cid/pull/140.

Furthermore, I have also raised a separate PR in the VCDM repo: https://github.com/w3c/vc-data-model/pull/1587. That PR would bring the VCDM spec in sync with DI and CID. Indeed, the arguments in this repo, namely https://github.com/w3c/vc-data-integrity/pull/327#discussion_r1912460821 and https://github.com/w3c/vc-data-integrity/pull/327#discussion_r1912575769 apply to VCDM as well.

>
> If not, can we make the SHOULD/MUST changes into a separate PR. If this PR just removes the numerical error codes, that's non-controversial and we can get that in sooner than later.

I have a separate PR in the VCDM now: https://github.com/w3c/vc-data-model/pull/1587. If that PR is accepted, then we are all set. If it is rejected, then separate PRs for DI and CID will have to be raised to sync up with VCDM (if we come to that, I am happy to do them). In any case, this PR (and its [CID equivalent](https://github.com/w3c/cid/pull/140)) are ready to be merged imho.

dlongley (approved):

iherman:
@msporny I believe this PR can be merged right away, it is the counterpart of https://github.com/w3c/cid/pull/140 (which has been merged).

msporny (approved):

msporny:

PR #351, opened 2026-07-09
Add basic threat model to specification

msporny:
This PR adds a basic threat model for Data Integrity as well as cryptosuites. We will still need to add many more threats and responses (I just focused on one threat per category for now), but it is largely focused on just getting the basic DFD and structure down before we start adding threats and responses from the WG.

A few new things I'm trying out in this PR:

* Using DFD3 iconography (though, DFD3 doesn't have icon suggestions for flow or object elements). @simoneonofri and @jandrieu -- I need more guidance here, I don't know what colors to use, I don't know what box shapes to use for flows and objects. Maybe trapezoids for objects and a right-facing path marker (fat arrow / chonky arrow) for flows? Something like this: https://github.com/w3c/vc-recognized-entities/blob/main/threat-model/diagrams/dfd.png
* Using YAML to express the outline and threats (having contributors have to hand-edit Javascript doesn't feel like a good contributor experience... just editing a paragraph of text does feel like a good experience). @jandrieu I like what I've done w/ the YAML in this PR more than the JS we have right now.
* Using this threat model as a basis for the Forgery Defense threat model (or not having to do one there since the threats are going to be described here). @wes-smith, I'm trying to figure out how to get the FD spec to just point to this threat model and the BSL threat model instead of creating an entirely new one.

@jandrieu @KDean-Dolphin @eric-schuh we are now repeating ourselves across multiple threat models -- holder, verifier, issuer, issuer system, verifier system, holder system, all being repeated... and the definitions are slightly different than the ones in the VCDM (they add more color based on the DFD). Lots of duplication going on at this point.


threat-model/threats/t1-signed-document-tampering.yaml
```diff
@@ -0,0 +1,50 @@
...
+  [=F2|transmitted to a verifier=]) modifies its contents to alter the
+  claims it expresses. For example, an attacker might change the subject of
+  a university degree credential, alter the amount on a signed invoice, or
+  insert statements that the [=E1|signer=] never made. A related variant
+  strips the proof from a [=O2|signed document=] entirely, or removes
+  individual proofs from a proof set, in the hope that a downstream system
+  will process the document without noticing the missing protection.
+  Detecting modification of signed information, wherever it occurs in the
+  document's lifecycle, is the central capability that the Data Integrity
+  specification was designed to provide.
```
dlongley, line 19: I do wonder if we need to say more about how holder software *is required* to strip proofs that are not understood and how that is different from this threat.

dlongley (approved):
Approving as a good starting point.

threat-model/threats/t2-canonicalization-resource-exhaustion.yaml
```diff
@@ -0,0 +1,54 @@
...
+    name: Poisoned Dataset Detection
+    type: Reduce
+    description: >-
+      Conforming RDF Dataset Canonicalization implementations are required
+      to detect potentially poisoned datasets and abort processing rather
+      than attempt to canonicalize them, and the canonicalization test
+      suite exercises this behavior. Implementers confirm that the
+      canonicalization library they deploy enforces these protections,
+      since bounded work is still consumed before detection triggers and
+      detection heuristics are necessarily imperfect.
```
dlongley, line 32: Something that's missing here as another mitigation is that a basic JSON schema check run prior to verification will catch any deviation from what is expected as well -- weeding out bad reconstructions by an attacker (in some use cases, entirely so). JSON schema checks are absolutely required anyway, no software should assume that what was submitted (even if digitally signed) matches what is expected (this is covered in another threat that's been discussed, but is related here).

Wind4Greg (approved):
Any way to get "PR preview" to work  with  threat modeling?  Also we will need another model  later that includes selective disclosure functionality.

msporny:
> Any way to get "PR preview" to work with threat modeling? Also we will need another model later that includes selective disclosure functionality.

Unfortunately, "it's complicated". PR Preview was never meant to preview multiple docs in a PR, it expects a top-level index.html and if it doesn't find that, it explodes, IIUC. Someone would have to do a major overhaul to get it to work.

wes-smith (approved):
+1 to merge, noting that the threat model is currently not comprehensive, but iterating from here is preferable to trying to finalize it in this PR.

simoneonofri (approved):
+1 to merge. As @wes-smith noted, a threat model does not need to be comprehensive before it can be published. This is already a useful baseline, and iterating from here is preferable to trying to finalize everything in this PR.

To answer the **DFD question**:  
 - I normally use a slightly modified version of DFD3, as in the [Threat Model for Decentralized Credentials](https://www.w3.org/TR/threat-model-decentralized-credentials/#diagram). This is only my personal convention: the Threat Modeling Guide is descriptive rather than prescriptive about notation (and in the updated version of the TMG there is a mapping between the richer notation from the TMG and DFD3). 
 - The current DFD has a good size and level of abstraction for starting the analysis.
 - I would label the currently unlabeled flows, such as those between P1 and P2, and connect the external entities to the processes they control. For human actors, these flows often represent a decision or action, such as providing information, confirming something, or choosing not to proceed.

As Threat Models are live documents, I normally use a “**Did We Do a Good Enough Job?**” section as a stage check:
 - What we did. Describe the model, threats, and responses already covered.
 - What remains. Record gaps, assumptions, and the next practical steps.
This makes the maturity and status of the threat model visible to reviewers and other readers.

Some ideas:
 - As a later improvement, I would also add an exact specification-section reference to each dictionary entry. This would make it clear where each model element comes from and provide traceability from the specification to the DFD, threats, and responses.
 - One question for the next stage: does this threat model already represent the as-is specification? In other words, was the specification reviewed to reverse-engineer its existing threats, security decisions, and responses, or do some entries describe candidate or future responses? If this is already the as-is model, plus some additional observations, this is great at this stage. Otherwise, I would make that as-is pass the next operational step.
 - After that, the model can be expanded through additional elicitation using STRIDE per element, or other useful lenses.

We could bring it to SING or the [Threat Modeling Community Group](https://www.w3.org/groups/cg/tmcg/) for a joint pass, or review it together asynchronously by email.

threat-model/threats/outline.yaml
```diff
@@ -0,0 +1,56 @@
+# Registry for the Data Integrity threat model.
+#
+# This file defines:
+#   - categories: the threat categories (from the W3C Threat Modeling
+#     Guide) and which threat IDs belong to each, in display order;
+#   - elements:   display labels for the data flow diagram elements
+#     defined in the Dictionary section of index.html;
+#   - files:      the threat definition files to load, relative to this
+#     file's directory.
```
TallTed, line 9:
```suggestion
#   - categories — the threat categories (from the W3C Threat Modeling
#         Guide) and which threat IDs belong to each, in display order
#   - elements — display labels for the data flow diagram elements
#         defined in the Dictionary section of index.html
#   - files — the threat definition files to load, relative to this
#         file's directory
```

threat-model/threats/t1-signed-document-tampering.yaml
TallTed: Yes, more should be said about that. :-)

wes-smith:
The group discussed on the 2026-07-21 VCB+DI task force call. We are in agreement that the PR should be merged pending editorial passes from some group members. This PR will be merged EOD (after incorporating new editorial updates) on 2026-07-22.

threat-model/threats/t2-canonicalization-resource-exhaustion.yaml
```diff
@@ -0,0 +1,54 @@
...
+  document=] supplied by a third party. The Data Integrity specification
+  requires that this class of attack be mitigated, but leaves the selection
+  and configuration of mitigations (detection heuristics, resource limits,
+  timeouts) to implementers.
+response:
+  - id: R3
+    name: Poisoned Dataset Detection
+    type: Reduce
+    description: >-
+      Conforming RDF Dataset Canonicalization implementations are required
```
iherman, line 26:
```suggestion
      Conforming RDF Dataset Canonicalization implementations are expected
```
Implementation are "expected", but not "required", which is an important difference. There is a [note](https://www.w3.org/TR/rdf-canon/#canon-algo-algo) in the RDFC spec, and the issue is listed as a [security issue](https://www.w3.org/TR/rdf-canon/#dataset-poisoning) but, in theory, an RDFC implementation may be conformant and not implement this defense feature. (In truth, all implementations that took part in the CR testing phase for RDFC do it.)

threat-model/threats/t5-crqc-breaks-signature-suites.yaml
```diff
@@ -0,0 +1,56 @@
...
+  - id: R13
+    name: Cryptographic Agility
+    type: Reduce
+    description: >-
+      The [=cryptosuite=] architecture separates the proof from the document
+      it protects and identifies the suite in every proof, so deployments
+      that remain connected can migrate to a post-quantum signature suite
+      through a configuration change once one is standardized. Post-quantum
+      Data Integrity [=cryptosuites=] based on NIST-standardized algorithms,
+      such as ML-DSA, are under active incubation. Agility enables
```
iherman, line 31:
```suggestion
      such as ML-DSA, are under active development. Agility enables
```

In the W3C jargon "incubation" is different. This is active development in a bona fide /TR document of the WG...

threat-model/threats/t5-crqc-breaks-signature-suites.yaml
```diff
@@ -0,0 +1,56 @@
...
+  Shor's algorithm would recover private keys from public keys, allowing an
+  attacker to perform the [=P3|sign=] process for any [=E1|signer=] whose
+  public key it has observed and to forge [=O2|signed documents=] that
+  [=P5|verify=] as authentic. This collapses the mitigation on which the
+  tamper-detection guarantee of this specification rests (see T1), and it
+  applies retroactively: documents signed before migration become forgeable
+  the moment the capability exists, even though the hash functions used by
+  the [=P2|hash=] process are not practically broken by known quantum
+  algorithms.
+response:
```
iherman, line 21: Shouldn't we have a separate reduction entry mentioning the forgery defense work? Or should the motivation of that work be a separate threat entry?

iherman (approved):

TallTed:
> This PR will be merged EOD (after incorporating new editorial updates) on 2026-07-22.

Well. I suppose 11am ET is EOD somewhere.

I was just about to ask for more time before the merge, since I got head down on the barcodes threat model thinking *that* was the one merging EOD today.

But it's apparently 6 hours past EOD for this PR, though for me it's only some minutes past.

Let's please remember that W3C and all its components are global in scope, and Anywhere On Earth (AOE) is the only timezone that's guaranteed to work for TZ-less times like EOD, Noon, Midnight, etc.

Wind4Greg:
Sorry @TallTed I must have merged a day earlier than you planned. This was when I thought I was told to merge.  You can either (a) put in a PR with your fixes, (b) wait for the next update to the Threat Model.

Issue #348, opened 2026-06-25
Move Common Algorthms from vc-di-quantum-resistant to vc-data-integrity and update vc-di-ecdsa and vc-di-eddsa


Wind4Greg:
This issues addresses commonality in the non-selective disclosure case, there is already an issue for the selective disclosure case.

When dealing with multiple new PQC signature algorithms in the  vc-di-quantum-resistant specification common functionality  was factored out into five parameterized [common algorithms](https://w3c.github.io/vc-di-quantum-resistant/#CommonAlgorithms): 

1. Hashing
2. Proof Configuration
3. Transformation
4. Proof Serialization
5. Proof Verification

The result of this was that it only took a page or two to specific credential creation and verification. All current  "data integrity" based  cryptosuires vc-di-ecdsa and vc-di-eddsa  follow a similar pattern and can reuse these common algorithms.  This would simplify the specifications and development of existing and future cryptosuites.

dlongley:
+1 to make these algorithms common, only written once, and parameterized for use by different cryptosuites.

Wind4Greg:
Proposed outline update to Data Integrity Algorithms section:

**Algorithms**
1. Processing Model
2. Add Proof
3. Add Proof Set/Chain
4. Verify Proof
5. Verify Proof Sets and Chains
6. Common  Algorithms (**new**  from [vc-di-quantum-resistant](https://w3c.github.io/vc-di-quantum-resistant), 5 items)
7. Selective Disclosure Functions (**new** from  [vc-di-ecdsa](https://w3c.github.io/vc-di-ecdsa/#selective-disclosure-functions) lots of items)
8. Context Validation  (*existing* just moved down)
9. Processing Errors (*existing* just moved down)

Wind4Greg:
Based on discussions from the VC-Barcodes/VC-DI task force  call of 2026-06-30 the preference was for two additional sections rather than more sub-sections, i.e., 

1. Algorithms  (Other names: Base Algs?, Core Algs?)
    1. Processing Model
    2. Add Proof
    3. Add Proof Set/Chain
    4. Verify Proof
    5. Verify Proof Sets and Chains
    6. Context Validation 
    7. Processing Errors
2. Common  Algorithms
    1.  Intro -- How to use common algs in proof creation/verification, with diagrams if possible
    2. Hashing
    3. Proof Configuration
    4. Transformation
    5.  Proof Serialization
    6. Proof Verification
3. Selective Disclosure Functions
[...]

PR #353, opened 2026-07-21
Add Cryptosuite Common Algorithms

Wind4Greg:
Per issue https://github.com/w3c/vc-data-integrity/issues/348 this PR add cryptosuite common algorithms, both high level: create  proof, and verify proof and well as supporting: Transformation, Hashing, etc... These can be reused in the  VC-DI-Quantum-Resistant, VC-DI-ECDSA, and VC-DI-EdDSA specifications.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/pull/353.html" title="Last updated on Jul 21, 2026, 6:27 PM UTC (fd18752)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/353/0253ea5...fd18752.html" title="Last updated on Jul 21, 2026, 6:27 PM UTC (fd18752)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -103,7 +103,7 @@
           // document unless you know what you're doing. If in doubt ask your friendly neighbourhood
           // Team Contact.
           //wgPatentURI:  "",
-          maxTocLevel: 3,
+          maxTocLevel: 4,
           /*preProcess: [ ],
           alternateFormats: [ {uri: "diff-20111214.html", label: "diff to previous version"} ],
           */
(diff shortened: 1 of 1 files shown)


iherman:
On §2.4.4. `cryptosuiteString` Datatype: I presume the main reason of introducing this datatype is to restrict the value of the `cryptosuite` to ASCII. Is it worth it? Would an implementation really go wrong if the name of a cryptosuite was expressed in, say, Chinese characters?

I wonder whether it is not simpler to say that the value of the cryptosuite is a string (ie, unicode), and leave it at that.

Wind4Greg:
Hi @iherman section 2.4.4 wasn't changed in this PR (or at  least  I didn't intend to change it). 

> On §2.4.4. `cryptosuiteString` Datatype: I presume the main reason of introducing this datatype is to restrict the value of the `cryptosuite` to ASCII. Is it worth it? Would an implementation really go wrong if the name of a cryptosuite was expressed in, say, Chinese characters?
> 
> I wonder whether it is not simpler to say that the value of the cryptosuite is a string (ie, unicode), and leave it at that.

dlongley:
+1 this PR doesn't change `cryptosuite`, so if that needs a discussion it should happen elsewhere.

That being said:

> Datatype: I presume the main reason of introducing this datatype is to restrict the value of the cryptosuite to ASCII.

It helps with CBOR-LD compression.

dlongley (approved):

iherman:
> Hi @iherman section 2.4.4 wasn't changed in this PR (or at least I didn't intend to change it).
> 
> > On §2.4.4. `cryptosuiteString` Datatype: I presume the main reason of introducing this datatype is to restrict the value of the `cryptosuite` to ASCII. Is it worth it? Would an implementation really go wrong if the name of a cryptosuite was expressed in, say, Chinese characters?
> > I wonder whether it is not simpler to say that the value of the cryptosuite is a string (ie, unicode), and leave it at that.

My bad. The diff highlighted it, but only in the TOC (I presume the exact position has changed), and it then caught my attention.

~~I will raise a separate issue.~~

iherman (approved):

Issue #344, opened 2026-01-21
Move General Selective Disclosure Functions from VC-DI-ECDSA to VC-Data-Integrity


Wind4Greg:
The current [DI-ECDSA][] cryptosuite contains general processing functions for enabling *selective disclosure* functionality. In addition to being used to provide selective disclosure for the ECDSA cryptosuite, these have already been used in the draft [DI-BBS][] cryptosuite and could be used in providing selective disclosure functionality in future versions of cyptosuites such as [DI-EDDSA][] and [DI-PQC][]. This issue proposes moving these functions to the VC Data Integrity specification.

This issue contains concrete proposals for changes to the VC Data Integrity specification, and VC-DI-ECDSA specifications. Included are annotated outlines of current [DI-ECDSA][], and [DI-BBS][] specifications. As a whole these changes do **NOT** change any functionality of the current set of specifications, but shift functionality between specifications as part of **editorial** specification refinement, usability, and readability enhancements.

Please provide comments in general and on the proposed outlines!

[VC-DI]: https://w3c.github.io/vc-data-integrity/ "VC Data Integrity"
[DI-ECDSA]: https://w3c.github.io/vc-di-ecdsa/ "VC DI ECDSA"
[DI-EDDSA]: https://w3c.github.io/vc-di-eddsa/ "VC DI EDDSA"
[DI-BBS]: https://w3c.github.io/vc-di-bbs/ "VC DI BBS"
[DI-PQC]: https://w3c-ccg.github.io/di-quantum-safe/ "Quantum Safe Cryptosuites" 

## Concrete Proposal for Moving General SD Functions to Data Integrity Specification

Recommend adding a new subsection on "Selective Disclosure Functions" to the Algorithms section after [Section 4.7 Processing Errors](https://w3c.github.io/vc-data-integrity/#processing-errors) with the following annotated structure:

Annotation guide: the algorithm/algorithm name is followed by an id in parentheses, a braced list of other algorithms/functions that use the function and a braced list of dependencies. Algorithm/function Ids are given in the reference sections on the existing document outlines. **Note**: annotations are for *editorial purposes* and not part of proposed changes.

Externally used algorithms/functions are shown in bold and listed first.

1. Selective Disclosure Functions
    1. **labelReplacementCanonicalizeJsonLd** (2SD) used by {EC9F, BBSF8} -- depends on: {Ext2, 1SD}
[...]

PR #358, opened 2026-08-12
Selective Disclosure Refactored

Wind4Greg:
This  PR is in response to issues https://github.com/w3c/vc-data-integrity/issues/344 and https://github.com/w3c/vc-data-integrity/issues/348. 

This PR refactors the selective disclosure functionality from VC-DI-ECDSA, VC-DI-Quantum-Resistant, and VC-DI-BBS into common functionality and "approach specific" functionality. It identifies and factors  out three distinct ***approaches*** to  SD: (1) *Signed Individual Claims* as used in VC-DI-ECDSA (SD), (2) *Salted Hash of Claims*  as in VC-DI-Quantum-Resistant, and (3) *Multi Claim Signature* as use in VC-DI-BBS.

Much of the common functionality comes from extracting *approach*  specific serialization, parsing, and verification into separate  functions, leaving common reusable functionality across all approaches.

***
<a href="https://pr-preview.s3.amazonaws.com/Wind4Greg/vc-data-integrity/pull/358.html" title="Last updated on Aug 25, 2026, 9:01 PM UTC (e9f0dcc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-integrity/358/8668a95...Wind4Greg:e9f0dcc.html" title="Last updated on Aug 25, 2026, 9:01 PM UTC (e9f0dcc)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3705,9 +3705,6 @@ data-cite="INFRA#nulls">Null</a>
         <h3>Salted Hash of Claims (SHoC) Specific Sub-Algorithms</h3>
         <section id="SaltedHashSD">
           <h4>SaltedHashSD</h4>
-          <p class="ednote">
-WORK IN PROGRESS.
-          </p>
 
           <p>
 The following algorithm specifies how to create salted hash data.
@@ -3750,9 +3747,77 @@ NIST SP-800-90A,B,C documents and/or BSI AIS 20 and AIS 31 documents.
         </section>
         <section  id="SerializeBase-SHoC">
           <h4>SerializeBase-SHoC</h4>
+
+          <p>
+The following algorithm specifies how to serialize SHoC approach base proof; 
+called by an
+issuer of an Quantum-Safe-SD-protected Verifiable Credential.
+The base proof is to be
+given only to the holder, who is responsible for generating a derived proof from
+it, exposing only selectively disclosed details in the proof to a verifier. This
+algorithm is designed to be used in conjunction with the algorithms defined
+in the Data Integrity [[VC-DATA-INTEGRITY]] specification,
+<a data-cite="vc-data-integrity#algorithms">
+Section 4: Algorithms</a>. Required inputs are
+cryptographic hash data (|hashData|) and
+<em>proof options</em> (|options|). The
+<em>proof options</em> MUST contain a type identifier for the
+<a data-cite="vc-data-integrity#dfn-cryptosuite">
+cryptographic suite</a> (|type|) and MAY contain a cryptosuite
+identifier (|cryptosuite|). A single <em>digital proof</em> value
+represented as series of bytes is produced as output.
+          </p>
+
+          <ol class="algorithm">
+            <li>
+Initialize |proofHash|, |mandatoryPointers|, |mandatoryHash|, |salts|,
+|saltedHashes|,
+and |hmacKey| to the values associated with their property names
+|hashData|.
+            </li>
+            <li>
+Initialize |toSign| the hash of the concatenation of |proofHash|,
+|mandatoryHash|, |salts|,
+and |saltedHashes| where the order of the concatenation MUST be followed. The
+cryptographic hash algorithm is that specified for signature algorithm in
+[[[#HashSigTable]]], e.g., SHA-256 for a security category 1 or 2 algorithms.
+            </li>
+            <li>
+Initialize |signature| to the result of digitally signing |toSign| using the
+private key associated with the base proof verification method.
+            </li>
+            <li>
+Initialize a byte array, |proofValue|, that starts with the SHoC approach
+base proof header bytes 0xd9, 0x5d, and 0x10.
+            </li>
+            <li>
+Initialize |components| to an array with five elements containing the values of:
+|signature|, |hmacKey|, |salts|, |saltedHashes|, and |mandatoryPointers|.
+            </li>
+            <li>
+CBOR-encode |components| per [[RFC8949]] where CBOR tagging MUST NOT be used on
+any of the |components|. Append the produced encoded value to |proofValue|.
+            </li>
+
+            <li>
+Return |proofValue| as <em>proof bytes</em>.
+            </li>
+          </ol>
+
+          <p class="ednote">
+The SHoC approach base proof header bytes 0xd9, 0x5d, and 0x10 were chosen
+to not collide with any of the [[VC-DI-ECDSA]] or [[VC-DI-BBS]] values. They are
+currently tentative.
+          </p>
+
         </section>
         <section id="ParseBase-SHoC">
           <h4>ParseBase-SHoC</h4>
+
+          <p class="ednote">
+WORK IN PROGRESS.
+          </p>
+
         </section>
         <section id="SerializeDerived-SHoC">
           <h4>SerializeDerived-SHoC</h4>
--- a/index.html
+++ b/index.html
@@ -2794,6 +2794,1296 @@ Let |proof| be a clone of the proof options, |options|.
                 </section>
         </section>
         </section>
+
+<section id="SDAlgorithms">
+        <h3>Selective Disclosure Algorithms</h3>
+        <section class="informative">
+          <h3>Introduction</h3>
+          <p>
+There are three high level algorithms used with credentials that support
+cryptographic selective disclosure. First, the issuer uses the <em>Add Base 
+Proof</em> algorithm to create a signed credential that supports selective 
+disclosure. The holder then uses the <em>Add Derived Proof</em> algorithm to
+create a selectively disclosed version of the issuer signed credential along 
+with a derived proof. Finally the verifier uses the <em>Verify Derived 
+Proof</em> algorithms to verify the received selectively disclose credential
+with derived proof against the <strong>issuer's</strong> verification 
+information.
+          </p>
+          <p>
+This specification supports three different implementation approaches to 
+selective disclosure that work primarily at the level of the <em>claims</em> 
+within a verifiable credential, that is, allowing the holder to selectively 
+disclose, or not, a  particular claim from a credential. The three approaches 
+take into account the efficiency of the fundamental cryptographic signature 
+algorithm used. Only one, (to be) standardized,  cryptographic signature 
+algorithm, BBS, natively supports selective disclosure, i.e., it produces a base 
+signature over an ordered list of claims, and facilitates the creation 
+of a "derived signature" over a subset of that list of claims. For other 
+cryptographic signature algorithms two approaches are typically used: 
+(1) provide individual signatures for each claim, or (2) sign a list of 
+"salted hashes" for all claims. The first we  will refer to as SIC 
+(Sign Individual Claims), the second SHoC (Salted Hashes of Claims), and the 
+case where the signature algorithm naturally supports multiple 
+claims - MCS (Multiple Claim Support).
+          </p>
+          <p>
+All approaches utilize similar mechanisms to protect the proof related meta 
+data, and minimize data leakage of non-disclosed information from the derived 
+proof. They all incur roughly similar overhead in the base and derived proofs
+for these purposes. Similarly, all approaches utilize the same encoding of the
+individual data items that go into a proof, CBOR, and the final encoding of raw
+octets into a string, base64-url. Hence this information is not shown in the
+summary [[[#SDApproachTable]]].
+          </p>
+          <p>
+The SIC approaches signs individual claims after the transformation step.
+Hence its base proof size scales as the number of claims times the size of the 
+signature produced by the chosen signature algorithm. Hence this approach is
+good for signature algorithms that produce small signatures. The size of the
+derived proof scales as the number of revealed claims times the signature size.
+This is a nice property, not shared by the other two approaches, i.e., the
+fewer claims reveal the smaller the proof.
+          </p>
+          <p>
+In the SHoC a salt value, random nonce, is created for each claim, that salt  
+value is concatenated with the claim and a cryptographic collision 
+resistant hash function is applied to produce a list of hashes, one for each
+claim.  The lists of salts and hashes are combined and signed with a selected 
+signature algorithm. Only one signature is needed, hence this aproach is good 
+for signature algorithms with larger signature sizes. The downside is that the 
+complete list of salts and hashes must be conveyed to the holder and verifier. 
+To verify a revealed claim the verifier would first verify the signature for the
+combined list of salts and hashes, then they must be told the index of the claim 
+in the list of salts and hashes, then using the appropriate salt value from the
+list they verify the value of hash of the salt concatenated with the received 
+claim and the hash from the table.
+          </p>
+          <p>
+In one MCS (Multi  Claim  Support) signature algorithm case, BBS, the produced 
+base proof is extremely short, 80 bytes. While the derived proof grows as 32 
+bytes per non-revealed claim.
+          </p>
+
+          <p>
+We sumarize the properties of these approaches in 
+[[[#SDApproachTable]]]. 
+          </p>
+          <table id="SDApproachTable" class="data numbered">
+            <caption><em>Selective Disclosure Approaches.</em> </caption>
+            <thead>
+              <tr>
+                <th>Name</th>
+                <th>Base Proof Contents</th>
+                <th>Base Proof Size</th>
+                <th>Derived Proof Contents</th>
+                <th>Derived Proof Size</th>
+              </tr>
+            </thead>
+            <tbody>
+              <tr>
+                <td>Signed Individual Claims (SIC)</td>
+                <td>list of signatures, one for each claim.</td>
+                <td>N<sub>claims</sub> &times; (signature size) </td>
+                <td>sub-list of signatures, one for each revealed claim.</td>
+                <td>N<sub>revealed claims</sub> &times; (signature size) </td>
+              </tr>
+              <tr>
+                <td>Signed Hash of Claims (SHoC)</td>
+                <td>list of salts, list of salted hashes, one for each claim, 
+                  plus signature over these two lists.</td>
+                <td>N<sub>claims</sub> &times; (salt size + hash size) + signature size</td>
+                <td>list of salts, list of salted hashes, one for each claim, 
+                  plus one signature over these two lists.</td>
+                <td>N<sub>claims</sub> &times; (salt size + hash size) + signature size</td>
+              </tr>
+              <tr>
+                <td>Multi Claim Signature, BBS</td>
+                <td>BBS signature</td>
+                <td>80 bytes</td>
+                <td>BBS <em>proof</em></td>
+                <td>272 + 32 &times;  (N<sub>claims</sub> - N<sub>revealed claims</sub>) bytes</td>
+              </tr>
+            </tbody>
+          </table>
+
+          <p>
+In the following sections we define the three main selective disclosure 
+algorithms in terms of sub-algorithms. These sub-algorithms are of two types: 
+those that are independent of the selective disclosure implementation approaches
+discussed above and those that have  approach specific implementations.
+          </p>
+        </section>  <!-- end Introduction to SD Algorithms section -->
+
+        <section>
+          <h3>Create Base Proof</h3>
+          <p class="ednote">
+This general procedure can be decomposed into: (a) generic Proof Configuration, 
+(b) generic "Base Proof Transformation", (c) generic "hashing", and optional
+specific "Extended Proof Hashing", and (d) approach specific 
+"Base Proof Serialization".
+          </p>
+          <p class="ednote">
+TODO: After precise defininitions of sub-algorithms, check all the parameter names,
+values, and return values for completeness and consistency.
+          </p>
+
+          <p>
+The following algorithm specifies how to create a [=data integrity proof=] given
+an <a>unsecured data document</a>. Required inputs are an
+<a>unsecured data document</a> ([=map=] |unsecuredDocument|), and a set of proof
+options ([=map=] |options|). A [=data integrity proof=] ([=map=]), or an error,
+is produced as output.
+          </p>
+
+          <ol class="algorithm">
+            <li>
+Let |proof| be a clone of the proof options, |options|.
+            </li>
+            <li>
+Let |proofConfig| be the result of running the algorithm in
+Section [[[#ProofConfigurationAlg]]] with
+|options| passed as a parameter.
+            </li>
+            <li>
+Let |transformedData| be the result of running the algorithm in
+[[[#TransformSD]]] with |unsecuredDocument|,
+|proofConfig|, and |options| passed as parameters.
+            </li>
+            <li>
+Let |hashData| be the result of running the algorithm in
+[[[#HashSD]]] with |transformedData| and |proofConfig|
+passed as a parameters.
+            </li>
+            <li>
+In the case of a SHoC approach, add to |hashData| the |saltedHash| data that 
+results from running the algorithm in [[[#SaltedHashSD]]].
+            </li>
+            <li>
+Let |proofBytes| be the result of running the approach specific 
+<em>SerializeBase</em>algorithm. For example, for the SIC approach use 
+[[[#SerializeBase-SIC]]] and for the SHoC approach use 
+[[[#SerializeBase-SHoC]]]. In all cases pass |hashData| and
+|options| as parameters.
+            </li>
+            <li>
+Let |proof|.|proofValue| be a <a data-cite="CID#multibase-0">
+base64-url-encoded Multibase value</a> of the |proofBytes|.
+            </li>
+            <li>
+Return |proof| as the [=data integrity proof=].
+            </li>
+          </ol>
+        </section>
+
+        <section>
+          <h3>Add Derived Proof</h3>
+          <p class="ednote">
+This general procedure can be decomposed into: (a) approach specific 
+"parse base proof", (b) generic "create disclosure data", and (c) approach 
+specific "serialize derived proof value".
+          </p>
+          <p class="ednote">
+TODO: After precise defininitions of sub-algorithms, check all the parameter names,
+values, and return values for completeness and consistency.
+          </p>
+
+          <p>
+The following algorithm creates a selective disclosure derived proof; called by
+a holder of a [=verifiable credential=] protected with a selective disclosure 
+'base proof'.
+The derived proof is to be given to the [=verifier=]. The inputs include a
+JSON-LD document (|document|), an selecitve disclosure base proof
+(|proof|), an array of JSON pointers to use to selectively disclose
+statements (|selectivePointers|), and any custom JSON-LD API options,
+such as a document loader. A single <em>selectively revealed document</em>
+value, represented as an object, is produced as output.
+          </p>
+
+          <ol>
+            <li>
+Initialize the |proofData| object to the result returned from the approach  
+specific <em>ParseBase</em> sub-algorithm. For example, for the SIC approach 
+use [[[#ParseBase-SIC]]] and for the SHoC approach use [[[#ParseBase-SHoC]]].
+At a minimum the |proofData| object will contain the following fields: 
+|hmacKey| and |mandatoryPointers|.
+            </li>
+            <li>
+Initialize |disclosureData| to the object returned when calling the algorithm in
+[[[#CreateDisclosureData]]], passing the |document|, |proofData|,
+|selectivePointers|, and any custom JSON-LD API options, such as a document
+loader. The |disclosureData| object will contain the following fields: 
+|revealDocument|, |mandatoryIndexes|, |selectiveIndexes|, |verifierLabelMap|
+|mandatory|, and |nonMandatory|.
+            </li>
+            <li>
+Initialize |newProof| to a shallow copy of |proof|.
+            </li>
+            <li>
+Replace |proofValue| in |newProof| with the result of calling the approach 
+specific  <em>SerializeDerived</em> algorithm. For example, for the SIC approach
+use [[[#SerializeDerived-SIC]]] and for the SHoC approach use 
+[[[#SerializeDerived-SHoC]]]. In all cases passing
+|proofData|,and |disclosureData| as parameters.
+            </li>
+            <li>
+Set the value of the "proof" property in |revealDocument| to |newProof|.
+            </li>
+            <li>
+Return |revealDocument| as the <em>selectively revealed document</em>.
+            </li>
+          </ol>
+
+
+        </section>
+
+        <section>
+          <h3>Verify Derived Proof</h3>
+          <p class="ednote">
+This procedure can  be decomposed into: (a) approach specific "parse derived proof"
+(b) generic "create verify data", and (c) approach specific verification procedure.
+          </p>
+          <p class="ednote">
+TODO: After precise defininitions of sub-algorithms, check all the parameter names,
+values, and return values for completeness and consistency.
+          </p>
+
+
+          <p>
+The following algorithm attempts verification of a selecitive disclosure derived
+proof. This algorithm is called by a verifier of an selective disclosure 
+protected [=verifiable credential=]. The inputs include a JSON-LD document
+(|document|), a selective disclosure derived proof (|proof|), and any
+custom JSON-LD API options, such as a document loader. This algorithm returns
+a [=verification result=]:
+          </p>
+
+          <ol class="algorithm">
+            <li>
+Let |unsecuredDocument| be a copy of |document| with the `proof` value removed.
+            </li>
+            <li>
+Initialize the |proofData| object to the result returned from the approach  
+specific <em>ParseDerived</em> sub-algorithm. For example, for the SIC approach 
+use [[[#ParseDerived-SIC]]] and for the SHoC approach use [[[#ParseDerived-SHoC]]].
+At a minimum the |proofData| object will contain the following fields: 
+|labelMap| and |mandatoryIndexes|.
+            </li>            
+            <li>
+Initialize the |verifyData| object to the value returned when calling the 
+algorithm in Section [[[#CreateVerifyData]]], passing the |document|, 
+|proofData|, and any custom JSON-LD API options, such as a document loader.
+The |verifyData| object will contain the following fields: |proofHash|, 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2794,6 +2794,2223 @@ <h4>Proof Verification</h4>
...
+        <section class="informative">
+          <h3>Introduction</h3>
+          <p>
+There are three high-level algorithms used with credentials that support
+cryptographic selective disclosure. First, the issuer uses the <em>Add Base 
+Proof</em> algorithm to create a signed credential that supports selective 
+disclosure. The holder then uses the <em>Add Derived Proof</em> algorithm to
+create a selectively disclosed version of the issuer-signed credential along 
+with a derived proof. Finally, the verifier uses the <em>Verify Derived 
+Proof</em> algorithms to verify the received selectively disclosed credential
```
TallTed: Either this is singular (as I've suggested) *or* the first line `three high-level algorithms` must change to `four` or whatever 2 + the number of `Verify Derived 
Proof algorithms` is.
```suggestion
Proof</em> algorithm to verify the received selectively disclosed credential
```

index.html
```diff
@@ -2794,6 +2794,2223 @@ <h4>Proof Verification</h4>
...
+with derived proof against the <strong>issuer's</strong> verification 
+information.
+          </p>
+          <p>
+This specification supports three different implementation approaches to
```
```suggestion
This specification allows for three different implementation approaches to 
```

index.html
```diff
@@ -2794,6 +2794,2223 @@ <h4>Proof Verification</h4>
...
+with a derived proof. Finally, the verifier uses the <em>Verify Derived 
+Proof</em> algorithms to verify the received selectively disclosed credential
+with derived proof against the <strong>issuer's</strong> verification 
+information.
+          </p>
+          <p>
+This specification supports three different implementation approaches to 
+selective disclosure that work primarily at the level of the <em>claims</em> 
+within a verifiable credential, that is, allowing the holder to selectively 
+disclose, or not, a  particular claim from a credential. The three approaches
```
```suggestion
within a verifiable credential; that is, they allow the holder to selectively 
disclose, or not, each particular claim in the credential. The three approaches 
```

index.html
```diff
@@ -2794,6 +2794,2223 @@ <h4>Proof Verification</h4>
...
+information.
+          </p>
+          <p>
+This specification supports three different implementation approaches to 
+selective disclosure that work primarily at the level of the <em>claims</em> 
+within a verifiable credential, that is, allowing the holder to selectively 
+disclose, or not, a  particular claim from a credential. The three approaches 
+take into account the efficiency of the fundamental cryptographic signature 
+algorithm used. Only one (to be) standardized cryptographic signature 
+algorithm, BBS, natively supports selective disclosure, i.e., it produces a base
```
TallTed: BBS cries out to be linked via something like `[[BBS]]` or `[[[BBS]]]` or at least a `data-cite`.

index.html
```diff
@@ -2794,6 +2794,2223 @@ <h4>Proof Verification</h4>
...
+          <p>
+This specification supports three different implementation approaches to 
+selective disclosure that work primarily at the level of the <em>claims</em> 
+within a verifiable credential, that is, allowing the holder to selectively 
+disclose, or not, a  particular claim from a credential. The three approaches 
+take into account the efficiency of the fundamental cryptographic signature 
+algorithm used. Only one (to be) standardized cryptographic signature 
+algorithm, BBS, natively supports selective disclosure, i.e., it produces a base 
+signature over an ordered list of claims and facilitates the creation 
+of a "derived signature" over a subset of that list of claims. For other
```
```suggestion
of a "derived signature" over a subset of that list of claims. In other 
```

index.html
```diff
@@ -2794,6 +2794,2223 @@ <h4>Proof Verification</h4>
...
+algorithm used. Only one (to be) standardized cryptographic signature 
+algorithm, BBS, natively supports selective disclosure, i.e., it produces a base 
+signature over an ordered list of claims and facilitates the creation 
+of a "derived signature" over a subset of that list of claims. For other 
+cryptographic signature algorithms, two approaches are typically used: 
+(1) provide individual signatures for each claim, or (2) sign a list of 
+"salted hashes" for all claims. The first we  will refer to as SIC 
+(Sign Individual Claims), the second SHoC (Salted Hashes of Claims), and the 
+case where the signature algorithm naturally supports multiple 
+claims - MCS (Multiple Claim Support).
```
```suggestion
"salted hashes" for all claims. We will refer to the first as `SIC` (Sign
Individual Claims), and the second as `SHoC` (Salted Hashes of Claims).
Cases where the signature algorithm naturally supports multiple claims,
will be referred to as `MCS` (Multiple Claim Support).
```

index.html
```diff
@@ -2794,6 +2794,2223 @@ <h4>Proof Verification</h4>
...
+          </p>
+          <p>
+All approaches utilize similar mechanisms to protect the proof-related metadata, 
+and minimize data leakage of non-disclosed information from the derived proof. 
+They all incur roughly similar overhead in the base and derived proofs
+for these purposes. Similarly, all approaches utilize the same encoding of the
+individual data items that go into a proof, CBOR, and the final encoding of raw
+octets into a string, base64-url. Hence, this information is not shown in the
+summary [[[#SDApproachTable]]].
```
```suggestion
All approaches use similar mechanisms to protect the proof-related metadata,
and to minimize data leakage of non-disclosed information from the derived
proof. They all incur roughly similar overhead in the base and derived proofs
for these purposes. Similarly, all approaches use the same encoding of the
individual data items that go into a proof, CBOR, and the final encoding of
raw octets into a string, `base64-url`. Hence, this information is not shown
in the summary [[[#SDApproachTable]]].
```

index.html
```diff
@@ -2794,6 +2794,2223 @@ <h4>Proof Verification</h4>
...
+          </p>
+          <p>
+The SIC approaches signs individual claims after the transformation step.
+Hence, its base proof size scales as the number of claims times the size of the 
+signature produced by the chosen signature algorithm. Hence, this approach is
+good for signature algorithms that produce small signatures. The size of the
+derived proof scales as the number of revealed claims times the signature size.
+This is a nice property, not shared by the other two approaches, i.e., the
+fewer claims reveal the smaller the proof.
```
```suggestion
The SIC approaches sign individual claims after the transformation step. Hence,
their base proof sizes scale as the number of claims times the size of the 
signature produced by the chosen signature algorithm. Thus this approach is
good for signature algorithms that produce small signatures. Along similar
lines, the size of the derived proof scales as the number of revealed claims
times the signature size, i.e., the fewer claims revealed, the smaller the
proof. This is a nice property, not shared by the other two approaches.
```

index.html
```diff
@@ -2794,6 +2794,2223 @@ <h4>Proof Verification</h4>
...
+for each claim.
+The lists of salts and hashes are combined and signed with a selected 
+signature algorithm. Only one signature is needed; hence, this approach is good 
+for signature algorithms with larger signature sizes. The downside is that the 
+complete list of salts and hashes must be conveyed to the holder and verifier. 
+To verify a revealed claim, the verifier would first verify the signature for the
+combined list of salts and hashes, then they must be told the index of the claim 
+in the list of salts and hashes, and then using the appropriate salt value from 
+the list, they verify the value of the hash of the salt concatenated with the 
+received claim and the hash from the table.
```
```suggestion
When following the SHoC approach, a salt value (random nonce) is created for
each claim. That salt value is concatenated with the claim, and a cryptographic
collision-resistant hash function is applied to the list of paired salts+claims,
to produce a list of hashes, one for each claim. The newly paired lists of salts
and hashes are combined and signed with a selected signature algorithm. Only one
signature is needed; thus, this approach is good for signature algorithms with
larger signature sizes. A downside of this approach is that the complete list of
salts and hashes must be conveyed to both holder and verifier. To verify a
revealed claim, the verifier needs first to verify the signature for the combined
list of salts and hashes; then they need to be told the index of the claim in
that list; and then, using the appropriate salt value from the list, they verify
the value of the hash of the salt concatenated with the received claim and the
hash from the table.
```

index.html
```diff
@@ -2794,6 +2794,2223 @@ <h4>Proof Verification</h4>
...
+          </p>
+          <p>
+In one MCS (Multi  Claim  Support) signature algorithm case, BBS, the produced 
+base proof is extremely short, 80 bytes. While the derived proof grows as 32 
+bytes per non-revealed claim.
```
TallTed: Does the derived proof grow per *revealed* claim?
```suggestion
In one MCS signature algorithm, BBS, the produced base proof is extremely short,
80 bytes, while the derived proof grows by 32 bytes per non-revealed claim.
```

index.html
```diff
@@ -2794,6 +2794,2223 @@ <h4>Proof Verification</h4>
...
+                <td>list of salts, list of salted hashes, one for each claim, 
+                  plus one signature over these two lists.</td>
+                <td>N<sub>claims</sub> &times; (salt size + hash size) + signature size</td>
+              </tr>
+              <tr>
+                <td>Multi Claim Signature, BBS</td>
+                <td>BBS signature</td>
+                <td>80 bytes</td>
+                <td>BBS <em>proof</em></td>
+                <td>272 + 32 &times;  (N<sub>claims</sub> - N<sub>revealed claims</sub>) bytes</td>
```
```suggestion
                <td>272 + 32 &times; (N<sub>claims</sub> - N<sub>revealed claims</sub>) bytes</td>
```

index.html
```diff
@@ -2794,6 +2794,2223 @@ <h4>Proof Verification</h4>
...
+              </tr>
+            </tbody>
+          </table>
+In the following sections, we define the three main selective disclosure algorithms in terms of sub-algorithms. These sub-algorithms are of two types: those that are independent of the selective disclosure implementation approaches discussed above and those that have approach-specific implementations.
```
TallTed: Duplicated just below...
```suggestion
```

index.html
```diff
@@ -2794,6 +2794,2223 @@ <h4>Proof Verification</h4>
...
+          <p>
+In the following sections, we define the three main selective disclosure 
+algorithms in terms of sub-algorithms. These sub-algorithms are of two types: 
+those that are independent of the selective disclosure implementation approaches
+discussed above and those that have  approach-specific implementations.
```
```suggestion
discussed above, and those that have approach-specific implementations.
```

index.html
```diff
@@ -2794,6 +2794,2223 @@ <h4>Proof Verification</h4>
...
+|proofConfig|, and |options| passed as parameters.
+            </li>
+            <li>
+Let |hashData| be the result of running the algorithm in
+[[[#HashSD]]] with |transformedData| and |proofConfig|
+passed as a parameters.
+            </li>
+            <li>
+In the case of a SHoC approach, add to |hashData| the |saltedHash| data that 
+results from running the algorithm in [[[#SaltedHashSD]]].
```
```suggestion
In a SHoC approach, add to |hashData| the |saltedHash| data that results
from running the algorithm in [[[#SaltedHashSD]]].
```

index.html
```diff
@@ -2794,6 +2794,2223 @@ <h4>Proof Verification</h4>
...
+            </li>
+            <li>
+Let |proofBytes| be the result of running the approach specific 
+<em>SerializeBase</em>algorithm. For example, for the SIC approach use 
+[[[#SerializeBase-SIC]]] and for the SHoC approach use 
+[[[#SerializeBase-SHoC]]]. In all cases pass |hashData| and
+|options| as parameters.
```
```suggestion
Let |proofBytes| be the result of running the approach-specific 
<em>SerializeBase</em> algorithm. For example, use [[[#SerializeBase-SIC]]] for
the SIC approach, and use [[[#SerializeBase-SHoC]]] for the SHoC approach. In
all cases, pass |hashData| and |options| as parameters.
```

index.html
```diff
@@ -2794,6 +2794,2223 @@ <h4>Proof Verification</h4>
...
+            </li>
+            <li>
+Let |proof|.|proofValue| be a <a data-cite="CID#multibase-0">
+base64-url-encoded Multibase value</a> of the |proofBytes|.
+            </li>
```
TallTed: Pushing the `>` closure of the `<a>` tag to the next line of the source makes the rendered HTML include <a href="site.example">`base64-url-encoded`</a> instead of <a href="site.example"> `base64-url-encoded`</a> (note the weirdly linked space char).
```suggestion
Let |proof|.|proofValue| be a <a data-cite="CID#multibase-0"
>`base64-url-encoded` Multibase value</a> of the |proofBytes|.
            </li>
```

index.html
```diff
@@ -2794,6 +2794,2223 @@ <h4>Proof Verification</h4>
...
+          <p>
+The following algorithm creates a selective disclosure derived proof; called by
+a holder of a [=verifiable credential=] protected with a selective disclosure 
+'base proof'.
+The derived proof is to be given to the [=verifier=]. The inputs include a
+JSON-LD document (|document|), an selecitve disclosure base proof
+(|proof|), an array of JSON pointers to use to selectively disclose
+statements (|selectivePointers|), and any custom JSON-LD API options,
+such as a document loader. A single <em>selectively revealed document</em>
+value, represented as an object, is produced as output.
```
```suggestion
The following algorithm creates a selective disclosure derived proof. It is
called by a holder of a [=verifiable credential=] that is protected by a
selective disclosure 'base proof'. The derived proof is to be given to a
[=verifier=]. The algorithm inputs include a JSON-LD document (|document|); a
selective disclosure base proof (|proof|); an array of JSON pointers to use to
selectively disclose statements (|selectivePointers|); and any custom JSON-LD
API options, such as a document loader. A single <em>selectively revealed
document</em> value, represented as an object, is produced as output.
```

index.html
```diff
@@ -2794,6 +2794,2223 @@ <h4>Proof Verification</h4>
...
+          </p>
+
+          <ol>
+            <li>
+Initialize the |proofData| object to the result returned from the approach  
+specific <em>ParseBase</em> sub-algorithm. For example, for the SIC approach 
+use [[[#ParseBase-SIC]]] and for the SHoC approach use [[[#ParseBase-SHoC]]].
+At a minimum the |proofData| object will contain the following fields: 
+|hmacKey| and |mandatoryPointers|.
```
TallTed: Does `Initialize ... to` have significant difference from `Let ... be`? I think using consistent language will help reader comprehension, and I think `let ... be` is probably the better choice (to be applied throughout; I've only literally suggested changing one of several instances, here).
```suggestion
Let the |proofData| object be the result returned from the approach-specific
<em>ParseBase</em> sub-algorithm. For example, for the SIC approach use
[[[#ParseBase-SIC]]], and for the SHoC approach use [[[#ParseBase-SHoC]]].
At a minimum, the |proofData| object will contain the fields |hmacKey| and
|mandatoryPointers|.
```

index.html
```diff
@@ -2794,6 +2794,2223 @@ <h4>Proof Verification</h4>
...
+            </li>
+            <li>
+Initialize |newProof| to a shallow copy of |proof|.
+            </li>
+            <li>
+Replace |proofValue| in |newProof| with the result of calling the approach 
+specific  <em>SerializeDerived</em> algorithm. For example, for the SIC approach
+use [[[#SerializeDerived-SIC]]] and for the SHoC approach use 
+[[[#SerializeDerived-SHoC]]]. In all cases passing
+|proofData|,and |disclosureData| as parameters.
```
```suggestion
Replace |proofValue| in |newProof| with the result of calling the approach-specific
<em>SerializeDerived</em> algorithm. For example, for the SIC approach, use
[[[#SerializeDerived-SIC]]], and for the SHoC approach, use
[[[#SerializeDerived-SHoC]]]. In all cases, pass |proofData| and |disclosureData|
as parameters.
```

index.html
```diff
@@ -2794,6 +2794,2223 @@ <h4>Proof Verification</h4>
...
+
+        <section>
+          <h3>Verify Derived Proof</h3>
+          <p>
+The following algorithm attempts verification of a selecitive disclosure derived
+proof. This algorithm is called by a verifier of an selective disclosure 
+protected [=verifiable credential=]. The inputs include a JSON-LD document
+(|document|), a selective disclosure derived proof (|proof|), and any
+custom JSON-LD API options, such as a document loader. This algorithm returns
+a [=verification result=]:
```
```suggestion
The following algorithm attempts verification of a selective disclosure
derived proof. This algorithm is called by a verifier of a selective
disclosure protected [=verifiable credential=]. Algorithm inputs include
a JSON-LD document (|document|); a selective disclosure derived proof
(|proof|); and any custom JSON-LD API options, such as a document loader.
This algorithm returns a [=verification result=]:
```

index.html
```diff
@@ -2794,6 +2794,2223 @@ <h4>Proof Verification</h4>
...
+          <ol class="algorithm">
+            <li>
+Let |unsecuredDocument| be a copy of |document| with the `proof` value removed.
+            </li>
+            <li>
+Initialize the |proofData| object to the result returned from the approach  
+specific <em>ParseDerived</em> sub-algorithm. For example, for the SIC approach 
+use [[[#ParseDerived-SIC]]] and for the SHoC approach use [[[#ParseDerived-SHoC]]].
+At a minimum the |proofData| object will contain the following fields: 
+|labelMap| and |mandatoryIndexes|.
```
```suggestion
Initialize the |proofData| object to the result returned by the approach-specific
<em>ParseDerived</em> sub-algorithm. For example, for the SIC approach, use
[[[#ParseDerived-SIC]]], and for the SHoC approach, use [[[#ParseDerived-SHoC]]].
At a minimum, the |proofData| object will contain the fields |labelMap| and
|mandatoryIndexes|.
```

index.html
```diff
@@ -2794,6 +2794,2223 @@ <h4>Proof Verification</h4>
...
+            </li>            
+            <li>
+Initialize the |verifyData| object to the value returned when calling the 
+algorithm in Section [[[#CreateVerifyData]]], passing the |document|, 
+|proofData|, and any custom JSON-LD API options, such as a document loader.
+The |verifyData| object will contain the following fields: |proofHash|, 
+|mandatoryHash|, and |nonMandatory|.
```
```suggestion
Initialize the |verifyData| object to the value returned by calling the 
algorithm in Section [[[#CreateVerifyData]]], passing the |document|, 
|proofData|, and any custom JSON-LD API options, such as a document loader.
The |verifyData| object will contain the following fields: |proofHash|, 
|mandatoryHash|, and |nonMandatory|.
```

index.html
```diff
@@ -2794,6 +2794,2223 @@ <h4>Proof Verification</h4>
...
+            </li>
+            <li>
+Initialize |verified| to the result of calling the approach specific 
+<em>VerifyDerived</em> algorithm. For example, for the SIC approach use  
+[[[#VerifyDerived-SIC]]] and  for the SHoC approach use 
+[[[#VerifyDerived-SHoC]]]. In all cases passing |proofData| and |verifyData| as
+parameters.
```
```suggestion
Initialize |verified| to the result of calling the approach-specific 
<em>VerifyDerived</em> algorithm. For example, for the SIC approach, use  
[[[#VerifyDerived-SIC]]], and for the SHoC approach, use
[[[#VerifyDerived-SHoC]]]. In all cases, pass |proofData| and
|verifyData| as parameters.
```

index.html
```diff
@@ -2794,6 +2794,2223 @@ <h4>Proof Verification</h4>
...
+            </li>
+            <li>
+Return a [=verification result=] with [=struct/items=]:
+              <dl data-link-for="verification result">
+                <dt>[=verified=]</dt>
+                <dd>The value of |verified|</dd>
+                <dt>[=verifiedDocument=]</dt>
+                <dd>
+|unsecuredDocument| if |verified| is `true`, otherwise <a
+data-cite="INFRA#nulls">Null</a>
```
```suggestion
if |verified| is `true`, |unsecuredDocument|; otherwise <a
data-cite="INFRA#nulls">Null</a>
```

index.html
```diff
@@ -2794,6 +2794,2223 @@ <h4>Proof Verification</h4>
...
+      <section>
+        <h3>Selective Disclosure Common Sub-Algorithms</h3>
+        <section  id="TransformSD">
+          <h4>TransformSD</h4>
+
+          <p>
+The following algorithm specifies how to transform an unsecured input document
+into a transformed document that is ready to be provided as input to the
+hashing algorithm in Section [[[#HashSD]]] and depending on approach may be
+input to the salted hashing algorithm [[[#SaltedHashSD]]].
```
```suggestion
The following algorithm specifies how to transform an unsecured input document
into a transformed document that is ready to be provided as input to the
hashing algorithm in Section [[[#HashSD]]] and, depending on approach, may be
provided as input to the salted hashing algorithm, [[[#SaltedHashSD]]].
```

index.html
```diff
@@ -2794,6 +2794,2223 @@ <h4>Proof Verification</h4>
...
+transformation options (|options|). The
+transformation options MUST contain a type identifier for the
+<a data-cite="vc-data-integrity#dfn-cryptosuite">
+cryptographic suite</a> (|type|), a cryptosuite
+identifier (|cryptosuite|), and a verification method
+(|verificationMethod|). The transformation options MUST contain an
+array of mandatory JSON pointers (|mandatoryPointers|) and MAY contain
+additional options, such as a JSON-LD document loader. A <em>transformed data
+document</em> is produced as output. Whenever this algorithm encodes strings, it
+MUST use UTF-8 encoding.
```
TallTed: Should `such as a JSON-LD document loader` be followed by `(|options|)`?
```suggestion
Required inputs to this algorithm are an <a data-cite=
"vc-data-integrity#dfn-unsecured-data-document">unsecured data document</a>
(|unsecuredDocument|) and transformation options (|options|). The
transformation options MUST contain a type identifier for the <a data-cite=
"vc-data-integrity#dfn-cryptosuite">cryptographic suite</a> (|type|), a
cryptosuite identifier (|cryptosuite|), and a verification method
(|verificationMethod|). The transformation options MUST contain an array of
mandatory JSON pointers (|mandatoryPointers|) and MAY contain additional
options, such as a JSON-LD document loader. A <em>transformed data
document</em> is produced as output. Whenever this algorithm encodes
strings, it MUST use UTF-8 encoding.
```

index.html
```diff
@@ -2794,6 +2794,2217 @@ <h4>Proof Verification</h4>
...
+disclosure. The holder then uses the <em>Add Derived Proof</em> algorithm to
+create a selectively disclosed version of the issuer-signed credential along 
+with a derived proof. Finally, the verifier uses the <em>Verify Derived 
+Proof</em> algorithm to verify the received selectively disclosed credential
+with derived proof against the <strong>issuer's</strong> verification 
+information.
+          </p>
+          <p>
+This specification allows for three different implementation approaches to 
+selective disclosure that work primarily at the level of the <em>claims</em>
```
```suggestion
selective disclosure that work primarily at the level of the [=claims=] 
```

Better to link to the definition of an official term (whenever possible)

index.html
```diff
@@ -2794,6 +2794,2217 @@ <h4>Proof Verification</h4>
...
+algorithm used. Only one (to be) standardized cryptographic signature 
+algorithm, BBS, natively supports selective disclosure, i.e., it produces a base 
+signature over an ordered list of claims and facilitates the creation 
+of a "derived signature" over a subset of that list of claims. In other 
+cryptographic signature algorithms, two approaches are typically used: 
+(1) provide individual signatures for each claim, or (2) sign a list of 
+"salted hashes" for all claims. We will refer to the first as `SIC` (Sign
+Individual Claims), and the second as `SHoC` (Salted Hashes of Claims).
+Cases where the signature algorithm naturally supports multiple claims,
+will be referred to as `MCS` (Multiple Claim Support).
```
```suggestion
"salted hashes" for all claims. We will refer to the first as <dfn class="export">SIC</dfn> (Sign
Individual Claims), and the second as <dfn class="export">SHoC</dfn< (Salted Hashes of Claims).
Cases where the signature algorithm naturally supports multiple claims,
will be referred to as <dfn class="export">MCS</dfn> (Multiple Claim Support).
```

We define these terms here (and use elsewhere). I added the class "export" to be able to refer to these terms by specref

/cc @msporny

index.html
```diff
@@ -2794,6 +2794,2217 @@ <h4>Proof Verification</h4>
...
+All approaches use similar mechanisms to protect the proof-related metadata,
+and to minimize data leakage of non-disclosed information from the derived
+proof. They all incur roughly similar overhead in the base and derived proofs
+for these purposes. Similarly, all approaches use the same encoding of the
+individual data items that go into a proof, CBOR, and the final encoding of
+raw octets into a string, `base64-url`. Hence, this information is not shown
+in the summary [[[#SDApproachTable]]].
+          </p>
+          <p>
+The SIC approaches sign individual claims after the transformation step. Hence,
```
```suggestion
The [=SIC=] approaches sign individual claims after the transformation step. Hence,
```

index.html
```diff
@@ -2794,6 +2794,2217 @@ <h4>Proof Verification</h4>
...
+their base proof sizes scale as the number of claims times the size of the 
+signature produced by the chosen signature algorithm. Thus this approach is
+good for signature algorithms that produce small signatures. Along similar
+lines, the size of the derived proof scales as the number of revealed claims
+times the signature size, i.e., the fewer claims revealed, the smaller the
+proof. This is a nice property, not shared by the other two approaches.
+          </p>
+          <p>
+When following the SHoC approach, a salt value (random nonce) is created for
```
```suggestion
When following the [=SHoC=] approach, a salt value (random nonce) is created for
```

index.html
```diff
@@ -2794,6 +2794,2217 @@ <h4>Proof Verification</h4>
...
+and to minimize data leakage of non-disclosed information from the derived
+proof. They all incur roughly similar overhead in the base and derived proofs
+for these purposes. Similarly, all approaches use the same encoding of the
+individual data items that go into a proof, CBOR, and the final encoding of
+raw octets into a string, `base64-url`. Hence, this information is not shown
+in the summary [[[#SDApproachTable]]].
+          </p>
+          <p>
+The SIC approaches sign individual claims after the transformation step. Hence,
+their base proof sizes scale as the number of claims times the size of the
```
iherman, line 2841: I also wonder: is it worth emphasizing that the signature of the claims happens using an ephemeral key-pair, with the private key thrown away after usage?

iherman (approved):
I would have to implement this to be able to comment on the algorithmic details. I presume it will be the job of the CR phase to do that. I approve the PR for merging to move to the text step.

iherman:
Not influencing my approval, and I know we had a discussion on all this. I _personally_ do not like the flat TOC structure we have here. If we want to avoid deep TOC levels what about:

- Renaming §4 to something like "full disclosure algorithms"
- Put everything with §4.3 and onward into a (new) §5 with "Selective disclosure algorithms"

It is not my intention to lie down the road on this, though.

index.html
iherman: It should be [[VC-DI-BBS]] or [[[VC-DI-BBS]]].

index.html
```diff
@@ -2794,6 +2794,2217 @@ <h4>Proof Verification</h4>
...
+the value of the hash of the salt concatenated with the received claim and the
+hash from the table.
+          </p>
+          <p>
+In one MCS signature algorithm, BBS, the produced base proof is extremely short,
+80 bytes, while the derived proof grows by 32 bytes per non-revealed claim.
+          </p>
+
+          <p>
+We sumarize the properties of these approaches in
```
```suggestion
We summarize the properties of these approaches in 
```

index.html
```diff
@@ -2794,6 +2794,2217 @@ <h4>Proof Verification</h4>
...
+function, i.e., SHA-256 for a P-256 curve. Per the recommendations of
+[[RFC2104]], the HMAC key MUST be the same length as the digest size; for SHA-256,
+this is 256 bits or 32 bytes.
+            </li>
+            <li>
+If |cryptosuite| is set to `ecdsa-sd-2023` then
+initialize |labelMapFactoryFunction| to the result of calling the algorithm of
+Section [[[#createhmacidlabelmapfunction]]], passing |hmac|; otherwise 
+initialize |labelMapFactoryFunction| to the result of calling the algorithm of
+Section [[[#createshuffledidlabelmapfunction]]], passing |hmac|;
```
dlongley: This is fine for getting this PR in, but we do want to abstract this a bit. One simple/quick way to do that is to let the function be optionally passed in as a parameter -- and set it to default to the one that `ecdsa-sd-2023` *does not* use, and then only have ecdsa-sd-2023 pass in the alternative (in its own spec). That allows other cryptosuites to pass other functions in as necessary when extending in the future -- without having to change this part.

index.html
```diff
@@ -2794,6 +2794,2217 @@ <h4>Proof Verification</h4>
...
+            <li>
+Initialize |hmac| to an HMAC API using |proofData|.|hmacKey|. The HMAC uses the same hash
+algorithm used in the signature algorithm, i.e., SHA-256 for a P-256 curve.
+            </li>
+            <li>
+If |cryptosuite| is set to `ecdsa-sd-2023` then
+initialize |labelMapFactoryFunction| to the result of calling the algorithm of
+Section [[[#createhmacidlabelmapfunction]]], passing |hmac|; otherwise 
+initialize |labelMapFactoryFunction| to the result of calling the algorithm of
+Section [[[#createshuffledidlabelmapfunction]]], passing |hmac|.
```
dlongley: Similar note as above on parameterizing this.

index.html
```diff
@@ -2794,6 +2794,2217 @@ <h4>Proof Verification</h4>
...
+          </ol>
+
+        </section>
+
+        <section id="ParseBase-SIC">
+          <h4>ParseBase-SIC</h4>
+          <p>
+The following algorithm parses the components of a legacy `ecdsa-sd-2023` selective
+disclosure base proof value or a general signed individual claim (SIC) approach
+selective disclosure base proof. The required inputs are a proof value
```
dlongley: Hmm, similar thing here with the legacy `ecdsa-sd-2023` approach; better that we make an option here that will default to the general approach and let the specialness be something we say in `ecdsa-sd-2023` (in its spec) that we pass in from there.

index.html
```diff
@@ -2794,6 +2794,2217 @@ <h4>Proof Verification</h4>
...
+            <li>
+Initialize |compressedLabelMap| to the result of calling the algorithm in
+Section [[[#compresslabelmap]]], passing |labelMap| as the parameter.
+            </li>
+            <li>
+If |cryptosuite| is set to `ecdsa-sd-2023` then
+initialize a byte array, |proofValue|, that starts with the ECDSA-SD disclosure
+proof header bytes `0xd9`, `0x5d`, and `0x01`. Otherwise initialize a byte 
+array, |proofValue|, that starts with the general SIC derived proof header bytes
+TBD.
```
dlongley: Similar situation -- let's let the header be an option, default to `TBD`, and let `ecdsa-sd-2023` pass in its own special value (via the di-ecdsa spec).

dlongley (approved):
Approving with some notes about how we should parameterize a few things where we special case `ecdsa-sd-2023` and let the di-ecdsa spec pass in its special options, where we default to the general value if nothing is specified. Parameterization of these options, I think, is a cleaner approach, keeping the specialness where it belongs (over in the di-ecdsa spec) -- and allowing future specs to be special as well if we discover a reason they need to be.

This can be done in a much smaller/focused PR after this is merged.

index.html
Wind4Greg: Yes, at least from a detailed cryptographic  point of view since it  reduces the number  of signatures  produced by the the  issuers long term public key.

dlongley:
Oh no, the latest commit (https://github.com/w3c/vc-data-integrity/pull/358/commits/b419ecef87bf1dc29cc0a1e82940be735e470331) introduced the whitespace EOL nonsense bug here, i.e., it caused like every line of the spec to get changed.

We might want to remove that commit (perhaps saving it on another branch) and merge this -- and then attempt another PR after we figure out how to have this stop happening.

I'm not sure if anyone has a good fix for this yet, but it's real pain.

iherman:
> Oh no, the latest commit ([b419ece](https://github.com/w3c/vc-data-integrity/commit/b419ecef87bf1dc29cc0a1e82940be735e470331)) introduced the whitespace EOL nonsense bug here, i.e., it caused like every line of the spec to get changed.
> 
> We might want to remove that commit (perhaps saving it on another branch) and merge this -- and then attempt another PR after we figure out how to have this stop happening.
> 
> I'm not sure if anyone has a good fix for this yet, but it's real pain.

This claude advice might help:

> - In a PR's Files Changed tab, append `?w=1` to the URL, or use the "Hide whitespace changes" toggle in the diff view settings (the ⚙️ icon on the Files tab). This won't fix the underlying repo state, but it'll immediately make the diff readable for that PR.

It seems to work in another repo...

PS: sorry for the extra noise on closing/opening the PR, I pushed the wrong button...

Wind4Greg:
Hi @iherman and @dlongley is there anything that I need to do with respect to the whitespace EOL thing?

TallTed:
@iherman — You may have missed this in the text you auoted --
> This won't fix the underlying repo state, but it'll immediately make the diff readable for that PR.

We need one or more actions that *will* fix the underlying repo state.

index.html
TallTed: To be clearer, `algorithm, BBS, natively` should be changed to —

- `algorithm, [[VC-DI-BBS]], natively` if we want `[VC-DI-BBS]` to show up in the text, with `href` to the citation entry in the References section.

- `algorithm, [[[VC-DI-BBS]]], natively` if we want the title of the referenced document to show up in the text, with `href` to the URI of that external document.

- `algorithm, [[[VC-DI-BBS]]] [[VC-DI-BBS]], natively` if we want the title of the referenced document to show up in the text, with `href` to the URI of that external document, followed by `[VC-DI-BBS]` to show up in the text, with `href` to the citation entry in the References section.

iherman:
> @iherman — You may have missed this in the text you auoted --
> 
> > This won't fix the underlying repo state, but it'll immediately make the diff readable for that PR.
> 
> We need one or more actions that _will_ fix the underlying repo state.

I agree. We are having some discussion in another WG/repo: https://github.com/w3c/epub-specs/pull/3058, to see where it goes. The huge difference between the two WG's is that, in the PM WG, we have one huge repo with loads of specs, whereas this WG favored a one repo per document approach. Ie, if we find a solution via .gitattributes, this has to be propagated to 20+ repositories. Could/can be done but only if we are 100% sure that it is the right solution. And, honestly, I am out of my depth to know what exactly the solution should be.

iherman:
> > @iherman — You may have missed this in the text you auoted --
> > > This won't fix the underlying repo state, but it'll immediately make the diff readable for that PR.
> > 
> > 
> > We need one or more actions that _will_ fix the underlying repo state.
> 
> I agree.

I opened a separate issue on this: https://github.com/w3c/vc-wg/issues/16. It is a general WG problem, not bound to this PR.

Issue #370, opened 2026-09-21
Bug: SHoC: derived proof reveals salts of non-disclosed claims, enabling offline guessing attacks (+ 2 algorithm inconsistencies)


studyzy:
## Summary

The current SHoC algorithms place the **complete** `salts` array into the derived proof
([SerializeDerived-SHoC](https://w3c.github.io/vc-data-integrity/#SerializeDerived-SHoC))
and require it for verification
([VerifyDerived-SHoC](https://w3c.github.io/vc-data-integrity/#VerifyDerived-SHoC),
`toVerify = proofHash ‖ mandatoryHash ‖ salts ‖ saltedHashes`).
The explainer text acknowledges this ("the complete list of salts and hashes must be
conveyed to both holder and verifier") but only lists size and total-count leakage
as downsides.

For a **non-disclosed** claim, the verifier knows `salt_i` and
`saltedHashes[i] = H(salt_i ‖ claim_i)`, which enables a per-target offline
dictionary attack: hash each candidate value against the known salt and compare.
Cost = |candidate domain| hashes — trivial for low-entropy claims (birth year,
country code, boolean attributes, postal codes). This is not a rainbow-table
attack (salts are per-claim random), but it is practically equivalent.
SIC, BBS, and SD-JWT do not have this issue: they reveal no commitment (or no
salt) for non-disclosed claims. As drafted, SHoC is strictly worse than SD-JWT
in this respect.

## Root cause

A single signature covers the concatenation `salts ‖ saltedHashes`, so the verifier
must reconstruct that exact byte string to check the signature — hence all salts
must be shipped. Signing the salts adds no security: the holder already receives
all of them, and each disclosed salt is bound to its signed hash by the check
`H(salt ‖ claim) == saltedHashes[pos]` (preimage-resistant).

## Proposed fix

Invariant: **the salt of a non-disclosed claim never leaves the holder.**

1. **SerializeBase-SHoC**: sign `proofHash ‖ mandatoryHash ‖ saltedHashes` only
   (drop `salts`; also drop the extra "hash of" — see bug 3 below).
2. **SerializeDerived-SHoC**: carry only the salts of selectively disclosed
   claims, i.e. `salts[i] = fullSalts[selectiveIndexes[i]]`.
3. **VerifyDerived-SHoC**: `toVerify` drops `salts`; length check becomes
   `len(salts) == len(selectiveIndexes) == len(nonMandatory)`; per-claim check
[...]

PR #371, opened 2026-09-21
Fix Salted Hash of Claims Algorithms

Wind4Greg:
This PR is in response to  issue https://github.com/w3c/vc-data-integrity/issues/370. The key issue resolved is only the *salts* of revealed claims are forwarded from the *holder* to the *verifier*. As such the list of *salts* cannot be hashed and signed at the  *issuer*. This PR (a) Updates "Salted Hash of Claims (SHoC) Specific Sub-Algorithms" to incorporate this update.  (b) Fixes an error  in the general "Create Disclosure Data" algorithm that  leads to an incorrect computation of *selectiveIndexes*.


index.html
```diff
@@ -3435,7 +3435,7 @@ <h4>CreateDisclosureData</h4>
 Initialize |selectiveIndexes| to an empty array.
             </li>
             <li>
-For each |absoluteIndex| in the keys in |groups|.|combined|.|matching|, convert
+For each |absoluteIndex| in the keys in |groups|.|mandatory|.|nonMatching|, convert
```
dlongley, line 3438: Seems like it would be right, but I haven't looked at an implementation to confirm this is right.

index.html
```diff
@@ -4013,8 +4012,13 @@ <h4>SerializeDerived-SHoC</h4>
 derived proof header bytes `0xd9`, `0x5d`, and `0x11`.
             </li>
             <li>
+Let |filteredSalts| be the array obtained by selecting the salt values from the
+|salts| array corresponding to the values in the |selectiveIndexes| array in
+increasing index order.
```
dlongley, line 4017: This is a no-op for this PR, but just noting that if we end up using an HMAC approach as in #372, these would just be computed below (`HMAC(saltKey, selectiveIndex) => salt`).

index.html
```diff
@@ -4125,21 +4129,21 @@ <h4>VerifyDerived-SHoC</h4>
 their property names in the |verifyData| parameter.
             </li>
             <li>
-If the length of |salts| does not match the length of |saltedHashes|,
+If the length of |salts| does not match the length of |slectiveIndexes|,
```
```suggestion
If the length of |salts| does not match the length of |selectiveIndexes|,
```

index.html
```diff
@@ -4159,11 +4163,11 @@ <h4>VerifyDerived-SHoC</h4>
               <ol class="algorithm">
                 <li>
 Let |computedHash| be the  hash of the concatenation of
```
```suggestion
Let |computedHash| be the hash of the concatenation of
```

dlongley (approved):

index.html
Wind4Greg: Hi @dlongley, my test vectors were generated using the `|groups|.|mandatory|.|nonMatching|` but I made a copy and paste error when writing up the text.

